PR #18, opened 2026-05-26
Passkeys + Small fixes

0xdewy:
## Passkey wallet support

Wallets created with `--passkey` encrypt the private key using the device
authenticator — Touch ID, Face ID, or a YubiKey — no passphrase required.

The encryption key is never stored. Instead, bloom uses the WebAuthn **PRF
extension** to request a deterministic 32-byte secret from the authenticator on
each unlock. That secret is fed into BLAKE3 to derive a wrap key, which decrypts
the private key via ChaCha20-Poly1305. The authenticator never exports its
internal secret — without the physical device, the wrap key cannot be reproduced
and the private key cannot be decrypted.

```
prf.salt      — 32 random bytes, stored on disk (not secret)
prf_output    — PRF(authenticator_secret, prf_salt)  derived live, never stored
wrap_key      — blake3::derive_key("bloom passkey wrap key", prf_output)
encrypted.key — ChaCha20Poly1305(private_key, wrap_key)
```

The recovery key (the wallet's raw private key as a hex string) is shown once in
a browser page at wallet creation, behind a confirmation step. Bloom never shows
it again. To recover: `bloom wallet import <name> 0x<key>`.

---

### Commands

```bash
bloom wallet new my-wallet --passkey    # Create  — opens browser for Touch ID / YubiKey ceremony
bloom wallet unlock my-wallet           # Unlock  — auto-detects passkey kind, opens browser ceremony
bloom wallet sign-policy my-wallet      # Re-sign on-disk policy.toml
bloom wallet rebind-passkey my-wallet   # Rotate  — bind to a new authenticator without moving funds
```

---

### Policy signing

Passkey wallets sign their `policy.toml` with the wallet's secp256k1 key,
creating a verifiable link between the wallet address and its spend limits. A
tampered policy file is rejected at read time. The hash is name-scoped —
`blake3("{name}:{policy}")` — so a valid signature from one wallet cannot be
transplanted to another.

---

### Security model

| Threat | Protected? |
|---|---|
| AI agent or script signs without user | ✅ ceremony required at first unlock each session |
| Attacker reads `~/.bloom/` from disk | ✅ cannot decrypt without authenticator |
| Stolen laptop or disk image | ✅ cannot decrypt without authenticator |
| Compromised user account | ✅ cannot decrypt without physical authenticator |
| Policy tampered on disk | ✅ signature verified on every read |
| Signature transplanted between wallets | ✅ policy hash is name-scoped |
| Partial wallet directory on crash | ✅ written to temp dir, atomically renamed into place |
| Local process forges PRF output during ceremony | ✅ `/prf-output` requires `clientDataJSON` bound to server-issued challenge |
| Malware active during unlock | ⚠️ could intercept PRF output at ceremony time |

> **Note:** once a wallet is unlocked, the signer is cached in the daemon for the
> lifetime of the process — no per-transaction re-authentication. The WebAuthn
> ceremony gates the first unlock per session, not every signature.

---

### Platform support

| Authenticator | PRF support |
|---|---|
| macOS Touch ID | ✅ |
| YubiKey 5 series | ✅ |
| Windows Hello | ✅ Chrome 147+ |
| Old FIDO U2F keys | ❌ rejected with clear error |

**Requirements:** Chrome ≥ 116 or Safari ≥ 18.

**Linux:** install **Google Chrome** (not Chromium, which has no built-in passkey
manager). Firefox on Linux does not support WebAuthn PRF and will show the
unsupported-authenticator error page.

---

### Incidental improvements

- **Per-sender nonce serialization** — concurrent `stage()` calls for the same wallet no longer race on nonce assignment
- **Enso quote staleness** — broadcasts of Enso routes older than 5 minutes are rejected
- **Gas limit fallback** — uses the route estimator's hint (with 25% buffer, capped at 30M) when `eth_estimateGas` fails
- **Stale IPC socket** — `ConnectionRefused` on an existing socket file is detected, the file removed, and a clear restart prompt shown
- **Socket cleanup on panic** — socket file is removed on drop, covering panic paths during `serve`
- **Price symbols** — `spot/` directory now lists known symbols; `SYMBOL_MAP` is public

---

### Breaking changes

- Empty passphrases for local wallets now error (previously silently accepted)
- `WalletKind` gains a `PasskeyGated` variant — exhaustive matches must handle it
- `verify_policy_sig` takes an additional `wallet_name: &str` parameter
- `--passphrase` is now `Option<String>`; passkey wallets skip it entirely
- Direct ERC-20 transfer path removed from the DeFi handler (dedicated VFS surface pending)

---

### Tests

```bash
# Keystore: PRF encryption round-trip, policy-signature verification,
# tamper detection, wrong-address rejection, passkey wallet info parsing
cargo test -p bloom-keystore -- passkey policy prf

# Tx engine: policy hard-deny (blocks confirm, marks failed),
# soft-warn (requires override sentinel to proceed)
cargo test -p bloom-tx -- policy_soft_warn policy_hard_deny
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1641,6 +1641,45 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "asn1-rs"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 1.0.69",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "965c2d33e53cb6b267e148a4cb0760bc01f4904c1cd4bb4002a085bb016d1490"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+ "synstructure",
+]
+
+[[package]]
+name = "asn1-rs-impl"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "assert-json-diff"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


josh-richardson:
@codex reivew

josh-richardson:
@codex review

PR #26, opened 2026-05-29
Add petal VFS namespace endpoints

josh-richardson:
## Summary

This adds the first stage of the petal VFS namespace so deployed petal view functions are projected into the mounted filesystem under `/bloom/petals/...`.

Key changes:

- Adds a `PetalsEndpointHandler` that lists deployed petal paths and exposes executable files for manifest functions marked `view = true`.
- Wires the daemon to mount the new `petals` handler alongside the existing transaction handler.
- Adds `bloom chain view-call` with `view` as an alias and stdin JSON parameter merging so mounted endpoint shims can accept arguments.
- Enforces petal deploy paths under `/bloom/petals/` and migrates examples/tests to the namespace.
- Extends Docker coverage with a live mounted petal VFS test that verifies direct vs mounted view output, then creates and mutates a `Counter` object and observes the state change through the mounted endpoint (`42 -> 77`).

## Validation

- `cargo fmt`
- `git diff --check`
- `cargo test -p bloom-vfs petals_handler -- --nocapture`
- `cargo test -p bloom-chain-node --test petal_admission -- --nocapture`
- `cargo test -p bloom-script read_only -- --nocapture`
- `cargo test -p bloom --features mount -- --nocapture`
- `cargo test -p bloom-chain-node rpc::tests::view_call_at_block_uses_retained_snapshot -- --nocapture`
- `cargo test -p bloom-petal-dex-it --test docker_petal_dex --no-run`
- `scripts/test-docker-petal-vfs.sh`


josh-richardson:
@codex review

violet-agent:
# Review comments for `feat/petals-in-fs`

Reviewed against `origin/master...HEAD` in `/root/dev/bloom-directory/bloom`.

Scope: static review of the branch diff, focused on code smells, correctness bugs, and security/abuse risks. Follow-up verification: `cargo`/`rustc` are installed via rustup under `/root/.cargo/bin`, but that directory was missing from this shell's `PATH`; after exporting it, `cargo check -p bloom-chain-node -p bloom-petals -p bloom-petal-manifest -p bloom-script --locked` passed.

## Findings

### 1. Medium — Default `chain_view_call` can report the new height while reading old live state

**Files:**
- `crates/bloom-chain-node/src/rpc.rs:841-879`
- `crates/bloom-chain-node/src/consensus_driver.rs:1154-1169`

`resolve_view_snapshot(None)` derives the default height from `block_store.latest_height()` and, when that height equals `block_head`, clones the in-memory `self.state`. During commit, the driver writes the block and state-index first, then updates `self.state`:

```rust
self.block_store.put(height, block)?;
self.state_index.put(height, &state_root, &blob_hash)?;
*self.state.lock() = execution.state.clone();
```

There is a race window where `latest_height()` has advanced but the live state is still the previous state. A concurrent view call can return old VFS/object data while reporting `"at_block": new_height`.

**Why it matters:** clients may cache or compose view results under the wrong block height. This is a consistency bug around block boundaries.

**Suggested fix:** either update `self.state` before publishing the block height/index visible to RPC, or make default view calls use the indexed snapshot for `block_head` unless the live state is known to correspond to that exact height.

---

### 2. Medium — Historical view snapshots do weaker integrity checks than startup restore

**Files:**
- `crates/bloom-chain-node/src/rpc.rs:890-908`
- `crates/bloom-chain-node/src/node.rs:100-128`

`load_indexed_state()` verifies the content-addressed blob hash and then calls `State::from_blob(&blob, state_root)`, but it does not verify:

- the blob header height equals the requested height;
- the blob header root equals the indexed root;
- the block header's `state_root` equals the indexed root.

Startup restore performs these extra checks and skips mismatched checkpoints.

**Why it matters:** a corrupt/stale `state_index` row could serve a mismatched state snapshot under the requested block context. View calls should fail closed the same way startup restore does.

**Suggested fix:** reuse the startup checkpoint validation path or add the same `State::blob_header` + block root checks in `load_indexed_state()`.

---

### 3. Medium — `chain_ls_objects { "all": true }` is an unpaginated full-state scan exposed via RPC

**File:** `crates/bloom-chain-node/src/rpc.rs:584-632`

The new `all: true` path scans every object while holding the state lock and builds a single full JSON array. Response-size enforcement happens after the response has already been built/serialized (`rpc.rs:307-314`). TCP RPC is explicitly unauthenticated when enabled (`rpc.rs:245-248`).

**Why it matters:** a caller can force O(number_of_objects) CPU/memory work and hold the state lock. On large state this becomes a straightforward DoS footgun, even if the final response is rejected for exceeding `RPC_MAX_RESPONSE_BYTES`.

**Suggested fix:** add pagination/limit parameters, cap maximum returned objects, release the lock before expensive serialization if possible, and consider disabling `all` over public TCP.

---

### 4. Medium — View endpoint stdin merge lets piped JSON override a baked endpoint into `commands` mode

**Files:**
- `crates/bloom/src/commands/chain.rs:479-516`
- `crates/bloom-chain-node/src/rpc.rs:963-981`

The CLI merges stdin JSON into `view-call` params while only preserving baked `path` and `function`. Any other key, including `commands`, is accepted from stdin. The RPC helper then prioritizes non-empty `commands` over `path/function`.

A generated VFS endpoint shim is intended to mean “call this endpoint”, but piped JSON can redirect it into an arbitrary composed view plan.

**Why it matters:** this weakens the filesystem endpoint boundary and can surprise scripts that pass JSON through stdin. It is not a chain mutation bug because RPC still enforces view-only execution, but it is a UX/security-boundary smell.

**Suggested fix:** when `path/function` are baked, reject or ignore stdin `commands`; or split endpoint-shim mode from raw `chain view-call --commands` mode.

---

### 5. Medium — Sync multi-chunk NFS writes can intentionally return weaker stability than requested

**File:** `crates/bloom-mount/src/adapter.rs:937-995`

The adapter comments note that embednfs rejects `actual < requested`, but for incomplete `DATA_SYNC` / `FILE_SYNC` buffers it still returns `WriteStability::Unstable`:

```rust
let actual_stability = if complete_payload.is_some() {
    requested
} else {
    WriteStability::Unstable
};
```

**Why it matters:** real clients can issue sync writes in multiple chunks for larger files or O_SYNC/fsync-like flows. Returning weaker stability than requested can surface as `SERVERFAULT`/`EREMOTEIO` and make valid mount writes fail nondeterministically.

**Suggested fix:** for sync writes, either flush/commit all contiguous chunks required to satisfy the request before replying, or reject unsupported multi-chunk sync writes with a clearer error before returning an invalid stability level.

---

### 6. Medium — Manifest signer/capability requirements are defined but dropped before PTB validation

**Files:**
- `crates/bloom-petal-manifest/src/types.rs:156-159`
- `crates/bloom-petal-manifest/src/stub.rs:52-76`
- `crates/bloom-script/src/chain_iface.rs:53-68`
- `crates/bloom-script/src/validator.rs:480-489`

The full manifest type contains `required_signers` and `required_capabilities`, but the stub projected into the PTB validator omits both fields. The validator only checks signer argument indices against the signed set.

**Why it matters:** manifest-declared authorization policy is not enforced by the PTB path. If those fields are intended to be security relevant, petals can appear to declare requirements that the chain silently ignores.

**Suggested fix:** either remove/rename these fields as documentation-only, or carry them into `FunctionDeclStub` and enforce them in `validate_ptb`.

---

### 7. Medium — Manifest decoder preallocates attacker-controlled list lengths

**Files:**
- `crates/bloom-petal-manifest/src/extract.rs:25-27`
- `crates/bloom-petal-manifest/src/codec.rs:121-128`

`extract_petal_manifest_v0()` decodes untrusted wasm custom-section bytes during admission. `read_list()` reads a `u32` length and immediately does `Vec::with_capacity(len)` before proving enough bytes exist.

**Why it matters:** a malformed custom section can advertise huge list lengths and cause large allocations before decode fails. The wasm-size cap limits input bytes, but not the capacity requested from the allocator.

**Suggested fix:** bound manifest list lengths globally and/or compare `len` to remaining input before allocating. A simple `MAX_MANIFEST_LIST_ITEMS` checked in `read_list()` would close the obvious DoS path.

---

### 8. Low/Medium — `getattr` render failures become silent zero-length files

**File:** `crates/bloom-mount/src/adapter.rs:648-671`

For renderable read-only files, `getattr` eagerly renders content to compute `st_size`. On render timeout/backend error it logs and returns `size = 0`; the comment notes that `cat` may then short-circuit and never surface the original error.

**Why it matters:** transient backend failures on dynamic petal endpoint/state files can look like legitimate empty files to normal tools, making correctness/debugging worse.

**Suggested fix:** cache a negative render result and make the next `read` surface the backend error, or return an unknown/optimistic size where the NFS layer supports it instead of silently reporting zero.

---

### 9. Low/Medium — View functions may emit logs despite the “view” purity check

**Files:**
- `crates/bloom-petals/src/chain_vm.rs:207-208`
- `crates/bloom-petals/src/chain_vm.rs:455-478`
- `crates/bloom-petals/src/chain_vm.rs:1450-1484`

The static view check rejects reachable mutating `object.*` imports, but `log.emit` remains allowed and appends to the execution report. This is not persisted state, but it is observable output from a supposedly read-only call.

**Why it matters:** if “view” means pure/read-only with no observable side effects beyond return values, logs violate that contract and can be used for misleading/noisy view responses.

**Suggested fix:** decide the contract explicitly. If views must be pure, include `log.emit` in the rejected imports for reachable view functions. If logs are allowed in views, document that and add tests covering the intended behavior.

---

### 10. Low/Medium — Coin mint authority is hard-coded to mutable VFS paths

**File:** `crates/bloom-petals/src/chain_vm.rs:674-689`, `crates/bloom-petals/src/chain_vm.rs:957-962`

`object.create` allows `Coin` creation if the caller's code hash is currently bound at one of a hard-coded list of paths:

```rust
const COIN_MINTER_PATHS: &[&str] = &[
    CORE_FUNGIBLE_PATH,
    "/bloom/petals/dex/faucet",
    "/bloom/petals/dex/pool",
    "/bloom/petals/dex/router",
];
```

**Why it matters:** authorization is tied to namespace binding rather than an explicit capability/manifest policy. That is brittle if those paths can be upgraded, squatted, or repurposed. It also bakes example DEX paths into core VM authorization.

**Suggested fix:** move coin-mint authority to an explicit capability or genesis/chain config policy, and keep example app paths out of core VM rules.

## Notes

- I did not include style-only observations unless they carried correctness or security risk.
- The branch has good defensive work around path/function segment validation and VFS collision checks; the main risks I found are consistency windows, unbounded scans/allocations, and authorization/purity boundaries.


---
Verification note: after correcting PATH with `export PATH=/root/.cargo/bin:$PATH`, `cargo check -p bloom-chain-node -p bloom-petals -p bloom-petal-manifest -p bloom-script --locked` passed locally.

josh-richardson:
Superseding my two earlier review-summary comments: those incorrectly summarized the older inline Codex review threads. Violet's actual review was the top-level 10-item comment from 2026-06-01, and I addressed those in commit `1ac3ab1`.

Violet review fixes:

1. Default `chain_view_call` height/state race: default view calls now prefer the indexed snapshot head instead of pairing a newly published block height with stale live state.
2. Historical snapshot integrity: `load_indexed_state` now checks blob hash, blob header height/root, and block header state root before serving a snapshot.
3. `chain_ls_objects { all: true }` unbounded scan: added capped `limit`/`offset` handling and stop scanning once the capped page is full.
4. View endpoint stdin override: endpoint-mode `chain view-call --path/--function` now rejects stdin `commands` instead of letting piped JSON switch it into composed command mode.
5. Sync multi-chunk NFS writes: incomplete `DATA_SYNC`/`FILE_SYNC` writes now fail explicitly instead of returning weaker stability than requested.
6. Manifest signer/capability requirements: the PTB stub now carries `required_signers`/`required_capabilities`; validation enforces signer count and fails closed on unsupported capability requirements.
7. Manifest decoder allocation: manifest list decoding now rejects impossible/oversized list lengths before `Vec::with_capacity`.
8. `getattr` render failures: GETATTR can still return size 0, but it now caches a negative render result so a follow-up READ surfaces EIO instead of looking like a valid empty file.
9. View logs: view-reachable `log.emit` is now rejected during chain wasm validation.
10. Coin mint authority: removed example DEX paths from core VM mint authorization; only the core fungible binding remains authorized, with a regression test that the DEX faucet path is denied.

Verification run locally:
- `cargo clippy -p bloom-chain-node -p bloom-petals -p bloom-petal-manifest -p bloom-script -p bloom-mount -p bloom --tests --features bloom-mount/mount -- -D warnings`
- Focused tests for the RPC snapshot/object listing changes, manifest allocation bound, PTB requirement validation, view-call stdin guard, view log rejection, coin mint authority, and mount sync/render-error behavior.

PR #35, opened 2026-06-05
Enforce invariants with host predicates

josh-richardson:
## Summary
- project manifest predicate ASTs into the runtime manifest stub
- evaluate invariant predicates host-side instead of trusting `__inv_*` wasm return bytes
- add a regression test where a tampered invariant export returns satisfied while the manifest predicate is violated

## Why
A hand-written or tampered petal can expose a valid `__inv_*` ABI and return success for every invariant. Runtime enforcement now derives the verdict from the manifest predicate over the canonical scope buffer, so the wasm export remains ABI/tooling metadata rather than the source of truth.

## Validation
- `cargo test --workspace --all-targets`
- `cargo test -p bloom-script -p bloom-petal-manifest`

### DIFF
--- a/crates/bloom-chain-node/tests/ptb_submit_e2e.rs
+++ b/crates/bloom-chain-node/tests/ptb_submit_e2e.rs
@@ -43,7 +43,7 @@ use bloom_petal_manifest::{
     },
 };
 use bloom_script::{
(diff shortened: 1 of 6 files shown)


josh-richardson:
Superseded by #36, which reapplies the reverted invariants branch on top of reverted master and includes the host-side enforcement plus follow-up soundness fixes.

PR #47, opened 2026-06-17
feat: add paid HTTP requests with x402 and Tempo MPP

violet-agent:
## Summary

- Adds paid HTTP request support to Bloom VFS and CLI, including unpaid probe/stage flows, confirmation controls, durable request artifacts, and response/receipt/audit persistence.
- Implements protocol backends for x402 and Tempo MPP in dedicated crates, with shared parsing, challenge normalization, wallet policy checks, and artifact-safe output types.
- Extends Bloom config/policy types for paid request wallet selection and spend gating, and documents the shipped behavior in `docs/specs/2026-06-15-paid-http-requests.md`.
- Refreshes the README/docs wordmark asset included on the branch.

## What changed

- New crates:
  - `crates/bloom-paid-http`: shared request parsing, payment requirement selection, policy evaluation, challenge normalization, and redaction-safe common types.
  - `crates/bloom-paid-x402`: x402 signing and EIP-3009 `X-PAYMENT` credential construction.
  - `crates/bloom-paid-mpp`: Tempo MPP authorization/retry integration through the `mpp` SDK.
- VFS integration:
  - Adds `/bloom/requests` request staging and confirmation lifecycle.
  - Keeps free reads/list/stat-style operations non-spending and side-effect safe.
  - Writes plans, policy checks, payment metadata, redacted challenge/credential artifacts, receipts, response bodies/headers, and audit records.
- CLI/test coverage:
  - Adds CLI parity for paid request flows.
  - Adds request handler and CLI tests covering staging, policy gates, x402 nonce binding, Tempo MPP behavior, failure states, and redaction guarantees.

## Safety and behavior notes

- Wallet selection is explicit request wallet, configured default wallet, then single-wallet fallback; ambiguous multi-wallet cases fail closed before probing.
- Confirmation is the spend/signing boundary. Hard policy denials block confirmation; soft warnings require the wallet policy override sentinel.
- x402 nonces are bound to the staged request id. Raw `X-PAYMENT` material is only sent on the paid retry and is not persisted.
- Tempo MPP signing goes through `Keystore::signer(wallet)`, so passkey-gated wallets require foreground unlock and locked passkeys fail with an explicit unlock error.
- Readable VFS artifacts should not contain raw `Authorization`, raw `X-PAYMENT`, signed transactions, voucher signatures, or other replayable credential material.

## Deferred / intentionally out of scope

- Durable Tempo MPP channel reuse, top-up, and close controls are not implemented here.
- Settled Tempo MPP sessions are recorded as `settled_no_durable_channel`, not reusable open channels.
- `/requests/sessions/<id>/topup` and `close` are not advertised as usable persisted controls yet.
- Server-side paid endpoint support remains out of scope.

## Verification

Recorded local verification on this branch:

- `cargo fmt --check`
- `cargo check -p bloom-vfs`
- `cargo clippy -p bloom-vfs --all-targets -- -D warnings`
- `cargo test -p bloom-vfs handlers::requests -- --nocapture` — 16 passed
- `cargo test -p bloom-vfs --lib` — 334 passed, 1 ignored
- `git diff --check`

Current GitHub checks for PR head `9059dae` are green:

- `fmt`
- `clippy`
- `test`
- `cargo test -- --ignored`
- `chain subprocess test`
- `locked release build`
- `docker petal VFS`
- `docker petal DEX acceptance`
- `real-wasm invariant gates`


crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+                            .join("sessions")
+                            .join(sid)
+                            .join("spent"),
+                    )
+                    .ok()
+                    .and_then(|s| parse_money(&s))
+                })
+                .unwrap_or(0.0);
+            checks.extend(evaluate_session_policy(&policy, &challenge, already_spent));
+            let dir = self.req_dir("pending", &id);
```
josh-richardson, line 217: `new.dry-run` reaches this same `pending` state and only records dry-run in the rendered plan. `confirm` does not later reject these requests, so a dry-run artifact remains spend/sign capable. Please persist a machine-readable dry-run flag and block confirmation, or stage dry-runs in a non-confirmable location.

crates/bloom-paid-x402/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+        if let Some(pay_to) = ctx.requirement.pay_to.as_deref()
+            && !candidate.pay_to.eq_ignore_ascii_case(pay_to)
+        {
+            return Err(format!(
+                "x402 selected pay_to {}, expected staged pay_to {pay_to}",
+                candidate.pay_to
+            ));
+        }
+        let header_value = candidate
```
josh-richardson, line 92: The staged `request_id` is not used in the signed x402 payload here; it is only copied into public metadata later. That makes the request-id-bound nonce claim false and means repeated confirmation after an ambiguous retry failure can mint another valid authorization for the same staged request.

josh-richardson (commented):
Comments from Codex

crates/bloom-paid-http/src/lib.rs
```diff
@@ -0,0 +1,861 @@
...
+                    .map(str::to_string)
+            })
+    };
+    let amount_usd = if intent == "session" {
+        json_f64(session, &["voucherAmountUsd", "amountUsd", "usd"])
+            .or_else(|| json_f64(&body_json, &["voucherAmountUsd", "amountUsd", "usd"]))
+            .or_else(|| parse_payment_amount_usd(asset.as_deref(), amount.as_deref()))
+    } else {
+        json_f64(charge, &["amountUsd", "usd"])
+            .or_else(|| json_f64(&body_json, &["amountUsd", "usd"]))
```
josh-richardson: Policy caps trust merchant-supplied `amountUsd` before deriving cost from the selected x402 requirement. A malicious 402 can advertise a low USD amount here while `accepts` contains a larger token amount that is actually signed. Please derive policy cost from the selected requirement, or reject inconsistent challenge amounts.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+                .map_err(|e| HandlerError::invalid(format!("invalid header {k}: {e}")))?;
+            let val = HeaderValue::from_str(v)
+                .map_err(|e| HandlerError::invalid(format!("invalid header {k}: {e}")))?;
+            retry = retry.header(name, val);
+        }
+        retry = retry.header(credential.header_name, credential.header_value.clone());
+        if let Some(body) = &request.body {
+            retry = retry.body(body.clone());
+        }
+        let response = retry
```
josh-richardson: At this point an x402 credential has already been signed. If the retry transport/read fails, the error returns before `credential.json`, `receipt.json`, `audit.json`, status, or the pending-to-failed move are written. That leaves the request pending and confirmable again, creating duplicate-payment risk. Please make post-sign failures durable and idempotent.

crates/bloom-paid-mpp/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+                return Err(format!("unsupported MPP intent '{other}'"));
+            }
+        }
+        .map_err(|e| format!("Tempo MPP credential: {e}"))?;
+        let authorization = mpp::format_authorization(&credential)
+            .map_err(|e| format!("format MPP Authorization: {e}"))?;
+        let authorization_sha256 = bloom_tools::sha256_hex(authorization.as_bytes());
+        let credential_value = serde_json::to_value(&credential)
+            .map_err(|e| format!("serialize MPP credential metadata: {e}"))?;
+        let retry = retry_paid_request(&self.client, request, &authorization).await?;
```
josh-richardson: Same durability issue for Tempo MPP: authorization has been formatted before this retry, but a retry error propagates out before the VFS writes failure/audit artifacts or moves the request out of `pending`. A second confirm can generate another credential/voucher.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+                return Ok(id);
+            }
+        };
+        let host = request.url.host_str().unwrap_or("unknown").to_string();
+
+        let mut req = self.client.request(
+            request.method.parse().unwrap_or(reqwest::Method::GET),
+            request.url.clone(),
+        );
+        for (k, v) in &request.headers {
```
josh-richardson, line 143: The unpaid probe forwards all user-provided headers, including replayable payment/auth material such as `Authorization`, `X-Payment`, and `Payment-Signature`, before any plan or confirmation. This weakens the confirm boundary and makes dry-run/staging unsafe for inputs that already contain payment credentials. Strip or reject sensitive headers on the probe path.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+    async fn read(&self, path: &VfsPath) -> Result<Vec<u8>, HandlerError> {
+        let segs = path.segments();
+        match segs {
+            [one] if one == "latest" => {
+                let (state, id) = self.read_latest()?;
+                Ok(format!("{state}/{id}\n").into_bytes())
+            }
+            [reference, name] if reference == "latest" => {
+                let (state, id) = self.read_latest()?;
+                fs::read(self.req_dir(&state, &id).join(name)).map_err(Into::into)
```
josh-richardson, line 835: This blindly joins VFS path segments into a real filesystem path. Because the mount path layer can percent-decode components, a segment containing decoded `/` or `..` can escape the request artifact directory. Please validate every segment as a single safe filesystem component before joining.

crates/bloom/src/main.rs
```diff
@@ -938,6 +986,66 @@ async fn run(cli: Cli) -> Result<()> {
...
+            let d = Daemon::from_home(home).context("build daemon")?;
+            let path = VfsPath::parse(&format!("/requests/{id}/plan.md"))?;
+            let bytes = d.vfs.read(&path).await.context("request plan")?;
+            std::io::Write::write_all(&mut std::io::stdout(), &bytes)?;
+            Ok(())
+        }
+        Cmd::Request(RequestCmd::Confirm { id }) => {
+            let (_home_permit, d) = build_write_daemon(home)?;
+            let path = VfsPath::parse(&format!("/requests/{id}/confirm"))?;
+            d.vfs
```
josh-richardson, line 1089: `bloom request confirm` writes directly to an in-process daemon with no wallet/passphrase/passkey unlock option and bypasses the existing IPC/write-unlocked path used for signer-consuming writes. The advertised CLI path likely cannot confirm locked wallets and does not mirror VFS unlock semantics.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+    }
+    out.sort_by(|a, b| a.name.cmp(&b.name));
+    Ok(out)
+}
+fn new_request_id() -> String {
+    let ms = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_millis();
+    format!("req_{ms}")
```
josh-richardson: Request ids are millisecond timestamps only. Concurrent writes in the same millisecond can collide, overwrite artifacts, and later `confirm` can remove an existing destination with the same id. Please add randomness/monotonic allocation and create directories with collision checks.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+) -> Result<(), HandlerError> {
+    write_json(
+        dir.join("request.toml"),
+        &json!({
+            "method": req.method,
+            "url": req.url.as_str(),
+            "wallet": wallet,
+            "max_amount_usd": req.max_amount_usd,
+            "headers": redacted_headers(&req.headers),
+            "body": req.body,
```
josh-richardson: The request body is persisted verbatim in readable artifacts. For paid HTTP, bodies can contain bearer-equivalent tokens, signed payloads, or API secrets. Either redact/omit bodies by default, classify sensitive content explicitly, or make the replay source private and non-readable.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2160 @@
...
+    if let Some(body) = &req.body {
+        http.push('\n');
+        http.push_str(body);
+        http.push('\n');
+    }
+    fs::write(dir.join("request.http"), http)?;
+    Ok(())
+}
+
+fn headers_to_json(headers: &HeaderMap) -> serde_json::Value {
```
josh-richardson, line 1323: Response headers are persisted without redaction. Paid APIs can return cookies, authorization-like headers, payment receipts with embedded signatures, or other bearer material. Please reuse a response-header redaction policy instead of dumping the full `HeaderMap`.

crates/bloom-paid-mpp/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+        let credential = match challenge.intent.as_str() {
+            "charge" => {
+                let provider = TempoProvider::new((*signer).clone(), &rpc_url)
+                    .map_err(|e| format!("TempoProvider: {e}"))?;
+                provider.pay(&payment_challenge).await
+            }
+            "session" => {
+                let mut provider = TempoSessionProvider::new((*signer).clone(), &rpc_url)
+                    .map_err(|e| format!("TempoSessionProvider: {e}"))?;
+                if let Some(max) = policy
```
josh-richardson: `max_deposit_usd` is a USD-denominated policy field, but `TempoSessionProvider::with_max_deposit` expects token atomic units. Passing `5.0` as `5` atomic units will clamp sessions to dust or fail unexpectedly. Convert using the challenge asset decimals/amount model or enforce this cap only in policy evaluation.

crates/bloom-paid-x402/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+            .map_err(|e| x402_keystore_signer_error(ctx.wallet, &self.keystore, e))?;
+        let info = self
+            .keystore
+            .info(ctx.wallet)
+            .map_err(|e| format!("x402 signer wallet metadata unavailable: {e}"))?;
+        let payment_required = parse_payment_required(ctx.challenge)?;
+        let candidate = select_candidate(&payment_required, signer).ok_or_else(|| {
+            "x402 upstream signer found no matching EIP-155 exact payment option".to_string()
+        })?;
+        if let Some(network) = ctx.requirement.network.as_deref() {
```
josh-richardson: This compares the raw staged requirement network to the candidate's normalized chain id string. Valid x402 v1 challenges often use names such as `base`, while the candidate may normalize to `eip155:8453`, causing confirmation to fail after successful candidate selection. Normalize both sides before comparing.

crates/bloom/src/main.rs
```diff
@@ -2048,6 +2170,24 @@ fn persist_outbox_review_approved(
     Ok(())
 }
 
+fn request_body_with_wallet(mut request: String, wallet: Option<&str>) -> String {
+    let Some(wallet) = wallet else {
+        return request;
+    };
+    if request.trim_start().starts_with("url") || request.trim_start().starts_with("method") {
```
josh-richardson: This TOML detection is too narrow for the advertised request formats. TOML with comments or reordered keys can be corrupted by inserting ` wallet=...` into the first line instead of adding a TOML `wallet` field. Prefer parsing the request and serializing/overriding the wallet field, or append valid TOML whenever the input parses as TOML.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 3452b22: dry-run paid requests now persist `dry_run.json` and both x402 and backend confirmation paths reject dry-run artifacts as non-confirmable. Added regression coverage that `/requests/new.dry-run` cannot be confirmed.

crates/bloom-paid-x402/src/lib.rs
violet-agent: Resolved in 3452b22: confirmation now writes a redacted single-use `signed_credential.json` marker immediately after minting an x402 credential and refuses to mint another credential for the same pending request after an ambiguous retry failure. The credential metadata remains request-id-bound and redacted in VFS.

crates/bloom-paid-http/src/lib.rs
violet-agent: Resolved in 3452b22: x402 normalization now derives the policy USD cost from the selected requirement amount/amountUsd before trusting top-level merchant `amountUsd`, with a regression test covering a low top-level amount and larger USDC requirement.

crates/bloom-paid-mpp/src/lib.rs
violet-agent: Resolved in 3452b22: removed the `max_deposit_usd` -> `with_max_deposit` conversion entirely. The USD cap remains enforced by Bloom policy evaluation, avoiding accidental token-atomic-unit clamping in the Tempo provider.

crates/bloom-paid-x402/src/lib.rs
violet-agent: Resolved in 3452b22: x402 network validation now normalizes staged network names (e.g. `base`) to x402 `ChainId` before comparing against the candidate chain id, while still accepting exact CAIP-2 matches.

crates/bloom/src/main.rs
violet-agent: Resolved in 3452b22: CLI wallet injection now parses valid TOML request bodies regardless of comments or key order and serializes an overridden TOML `wallet` field, falling back to request-line injection only for non-TOML HTTP-style input. Added a TOML regression test.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 9620f3c: the unpaid probe path now skips sensitive/replayable request headers (`Authorization`, `Proxy-Authorization`, `X-Payment`, `Payment-Signature`, API-key variants) before sending the probe. The persisted request artifacts continue to redact those headers as well.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 9620f3c: every VFS handler entrypoint (`lookup`, `list`, `read`, and `write`) now validates path segments with `validate_vfs_segments` before joining them into artifact paths. Empty, `.`, `..`, slash/backslash-containing, and absolute components are rejected.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 9620f3c: request IDs are no longer millisecond-only. They now include milliseconds, subsecond nanoseconds, and a process-local atomic counter (`req_<ms>_<nanos>_<seq>`), which avoids same-millisecond collisions under concurrent staging.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 9620f3c: readable request artifacts no longer persist the request body verbatim. `request.toml` now records `body_redacted` plus `body_sha256` when a body was present, and `request.http` omits the body content, so bearer-equivalent payloads are not exposed through normal VFS reads.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 9620f3c: response header persistence now applies a response-header redaction policy. Cookies, auth headers, payment credentials/receipts, and API-key variants are written as `redacted` instead of dumping the raw `HeaderMap`.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Partially addressed in 3452b22/9620f3c: x402 confirmation now writes a redacted `signed_credential.json` marker immediately after credential minting and refuses a second confirm for the same pending request, which closes the duplicate-credential/idempotency risk. I’m leaving this thread unresolved because the stronger durability ask—writing full failure/audit artifacts and moving the request out of `pending` when the post-sign retry transport/read fails—still needs a follow-up patch.

crates/bloom-paid-mpp/src/lib.rs
violet-agent: Not fully resolved yet. The dry-run/idempotency hardening landed, but the Tempo MPP path still needs the same durable post-authorization failure handling: once authorization has been formatted, retry transport/read errors should persist failure/audit artifacts and move the request out of `pending` rather than propagating before VFS state is finalized. Leaving this unresolved for the follow-up patch.

crates/bloom/src/main.rs
violet-agent: Not resolved yet. To close this, `bloom request confirm` needs to mirror signer-consuming VFS writes: add unlock options (wallet + passphrase/passkey flow, likely matching `vfs write --unlock-wallet --passphrase`) or route through the existing unlocked-write path so the signing key is unlocked in the process/daemon that actually signs. Leaving unresolved until that CLI path is implemented and tested.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 6db7ec7: after x402 signing, retry send/read errors are now finalized durably instead of bubbling out. The handler writes redacted credential metadata, response/status `0`, response body/error text, body hash, receipt/audit entries (`event: signed_retry_failed`), marks status `failed`, moves the artifact to `failed/<id>`, and updates `latest`, while the existing signed-credential marker keeps re-confirm idempotent.

crates/bloom-paid-mpp/src/lib.rs
violet-agent: Resolved in 6db7ec7: the Tempo MPP backend now catches retry transport/read errors after authorization formatting and returns a redacted failed `PaymentExecution` with status `0` instead of an error. `confirm_with_backend` then writes credential/receipt/audit/response artifacts and moves the request out of `pending` into `failed`, preventing another credential/voucher mint on re-confirm.

crates/bloom/src/main.rs
violet-agent: Resolved in 6db7ec7: `bloom request confirm` now accepts `--unlock-wallet` plus `--passphrase`/`BLOOM_PASSPHRASE`, tries the existing IPC `write_unlocked` path first, and falls back to in-process unlock before writing `/requests/<id>/confirm`. Passkey wallets use the foreground passkey unlock flow, so the signing key is unlocked in the process/daemon performing the signer-consuming write.

crates/bloom-paid-http/src/lib.rs
```diff
@@ -0,0 +1,870 @@
...
+            asset: challenge.asset.clone(),
+            amount: challenge.amount.clone(),
+            pay_to: None,
+            resource: None,
+            raw: json!({}),
+        }]
+    } else {
+        challenge.accepts.clone()
+    };
+    candidates.into_iter().find(|req| {
```
josh-richardson: This still lets policy approve a different x402 requirement than the one the signer ultimately uses. The check evaluates every candidate with the challenge-level `amount_usd` and does not filter to signer-supported schemes; then the x402 adapter independently picks the first supported EIP-155 exact candidate and only checks network/asset/payee. A cheap unsupported `accepts` entry can pass policy while a later expensive exact entry is signed. Please select the concrete signer-supported requirement first and evaluate policy against that exact requirement's scheme/network/asset/payee/amount.

crates/bloom-paid-http/src/lib.rs
```diff
@@ -0,0 +1,870 @@
...
+            out.push(deny(
+                "payments.assets.allow",
+                format!("{asset} not allowed"),
+            ));
+        } else if !payments.assets.allow.is_empty() {
+            out.push(pass("payments.assets.allow", format!("{asset} allowed")));
+        }
+    }
+    if let Some(network) = network {
+        if contains_ci(&payments.networks.deny, network) {
```
josh-richardson: Network allow/deny is still raw string matching. x402 signing now treats names like `base` and `eip155:8453` as equivalent, but policy here does not. A policy denying `base` will not deny `eip155:8453`. Please normalize network names before policy comparison, or compare against all canonical aliases.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2366 @@
...
+) -> Result<(), HandlerError> {
+    write_json(
+        dir.join("request.toml"),
+        &json!({
+            "method": req.method,
+            "url": req.url.as_str(),
+            "wallet": wallet,
+            "max_amount_usd": req.max_amount_usd,
+            "headers": redacted_headers(&req.headers),
+            "body_redacted": req.body.is_some(),
```
josh-richardson, line 1288: This redacts the body from `request.toml`, but confirm reconstructs the paid retry from `request.toml`, so staged paid POST/PUT requests lose their body before signing/retry. At the same time, `request.http` below still writes the body in cleartext. Please keep a private non-readable replay copy for confirm and keep readable artifacts redacted.

crates/bloom-paid-mpp/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+        "mpp_tempo"
+    }
+
+    async fn confirm(
+        &self,
+        challenge: &NormalizedChallenge,
+        request: &ParsedRequest,
+        wallet: &str,
+        _policy: &Policy,
+        _request_id: &str,
```
josh-richardson, line 94: MPP still has no durable single-use marker before credential creation, and the request id is ignored. If the process crashes or two confirms race after the Tempo credential is minted but before VFS artifacts are written/moved, the same pending request can mint another voucher/payment. The x402 path now has a pre-retry marker; MPP needs equivalent idempotency before `provider.pay`.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2366 @@
...
+                })
+                .unwrap_or(0.0);
+            checks.extend(evaluate_session_policy(&policy, &challenge, already_spent));
+            let dir = self.req_dir("pending", &id);
+            fs::create_dir_all(dir.join("response"))?;
+            write_request_artifacts(&dir, &request, &wallet, "pending")?;
+            fs::write(dir.join("status"), b"pending\n")?;
+            write_json(dir.join("dry_run.json"), &json!({"dry_run": dry_run}))?;
+            fs::write(dir.join("challenge.raw"), &body)?;
+            write_json(dir.join("challenge.json"), &challenge)?;
```
josh-richardson: `challenge.json` persists `NormalizedChallenge.headers`, which currently includes all 402 response headers unredacted. The response header redaction helper is only applied to `response/headers.json`, so challenge-time `set-cookie`, auth-like headers, `www-authenticate`, or receipt material can leak through this pending artifact. Please redact/sanitize challenge headers before writing them.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -0,0 +1,2366 @@
...
+                    } else {
+                        v.clone()
+                    },
+                )
+            })
+            .collect::<std::collections::BTreeMap<_, _>>()
+    )
+}
+
+fn is_sensitive_request_header(name: &str) -> bool {
```
josh-richardson, line 1358: `cookie` is missing from the request-side sensitive header list. User-supplied cookies will still be sent on the unpaid probe and persisted in readable request artifacts. Since response-side redaction already treats cookies as sensitive, request-side handling should do the same.

crates/bloom/src/main.rs
```diff
@@ -491,6 +512,37 @@ enum BuilderKeysCmd {
...
+        /// Paying wallet. If omitted, config.default_wallet or the only wallet is used.
+        #[arg(long)]
+        wallet: Option<String>,
+        /// Stage/probe only; never spends or signs.
+        #[arg(long)]
+        dry_run: bool,
+    },
+    /// Show the staged payment plan for an id or `latest`.
+    Plan { id: String },
+    /// Confirm a pending paid request.
```
josh-richardson, line 530: The first-class `bloom request confirm` command still has no way to write the wallet policy override sentinel. Both IPC and local paths always write literal `confirm`, but warning policies require the configured sentinel. Please add an explicit confirmation text/override option so soft-warning paid requests do not require falling back to low-level `vfs write`.

josh-richardson (commented):
Comments from Codex

crates/bloom-paid-http/src/lib.rs
violet-agent: Resolved in 88a1037: selection now filters to signer-supported x402 requirements first, then evaluates policy against that exact concrete requirement (scheme/network/asset/payee/amount). The VFS staging path also stamps the selected requirement back onto the challenge and uses that requirement amount for the confirm-time policy recheck, so the signer cannot switch from a cheap unsupported candidate to a later expensive exact candidate after policy approval.

crates/bloom-paid-http/src/lib.rs
violet-agent: Resolved in 88a1037: payment policy network comparisons now use canonical alias matching. Network allow/deny entries are compared across equivalent forms such as `base`, `base-mainnet`, `8453`, and `eip155:8453`, matching the x402 signer-side network equivalence behavior.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 88a1037: readable artifacts stay redacted, and confirm now uses a private replay copy. `request.toml`/`request.http` no longer expose the body, while `private/request_replay.json` keeps the original headers/body for paid POST/PUT replay during confirm and is not listed as a readable VFS artifact.

crates/bloom-paid-mpp/src/lib.rs
violet-agent: Resolved in 88a1037: MPP confirm now writes a single-use marker before calling the Tempo backend/provider pay path, and re-confirm refuses pending requests with that marker. This gives MPP the same pre-credential idempotency guard as x402 for crash/race cases after voucher/payment creation begins.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 88a1037: staged `challenge.json` is now written through a redaction helper that sanitizes challenge-time headers before persistence, so sensitive 402 headers such as cookies/auth material/receipt-like fields are not stored in the pending artifact.

crates/bloom-vfs/src/handlers/requests.rs
violet-agent: Resolved in 88a1037: `cookie` is now included in the request-side sensitive header list, so user-supplied cookies are skipped for the unpaid probe/retry forwarding path and redacted from readable request artifacts consistently with response-side handling.

crates/bloom/src/main.rs
violet-agent: Resolved in 88a1037: `bloom request confirm` now accepts explicit confirmation text via `--confirm-text`, and both IPC and local/unlock paths write those bytes to `/requests/{id}/confirm`. Soft-warning paid requests can now pass the configured override sentinel without dropping down to raw `vfs write`.

GavinPacini:
Superseded by #51.

PR #49, opened 2026-06-19
Hyperliquid agent integration + supporting wallet/tx fixes

0xdewy:
## What & why

This PR makes Bloom easier for agents to use safely by combining the Hyperliquid VFS integration with the first capability-oriented wallet surfaces agents need to understand authority, balances, and next actions.

The key goals are:
- make Hyperliquid trading usable through bounded agent sessions instead of repeated owner-wallet signing;
- make wallet authority legible through read-only capability views;
- make agent onboarding clearer with `/next.md` and updated docs;
- harden Hyperliquid session creation/recovery so daemon restarts do not silently strand usable API-wallet slots.

### Hyperliquid agent sessions

- Adds `/hyperliquid/<network>/...` VFS surfaces for reads, owner-signed exchange actions, and bounded `agent_sessions/` trading.
- Uses a separate user-signed `approveAgent` path for session creation and API-wallet signing for session actions.
- Enforces wallet `[hyperliquid]` policy before signing: allowed assets/order types, max notional, max position, max loss, max leverage, and default-deny for unsupported transfer/withdrawal actions.
- Adds session status, audit trail, per-session `last_response.json`, forced cleanup hooks, stale snapshot indicators, and owner-signed orphan recovery.
- Hardens creation against concurrent session races with pending reservations and live persisted-session checks.
- Fixes review display so micro-USD caps are rendered as human USD amounts.
- Blocks plain IPC for Hyperliquid owner-signer paths while keeping bounded API-wallet session actions compatible.

### Hyperliquid restart resilience

- Seals newly generated ephemeral agent keys before submitting `approveAgent`.
- Stores sealed keys locally with a daemon-local KEK and `0600` file permissions on Unix.
- Persists original session bounds into `session.json`, so recovered sessions use the approved policy limits rather than current policy guesses.
- On status/action/cleanup paths after daemon restart, Bloom attempts to reopen the sealed key and verifies the agent still appears in Hyperliquid `extraAgents` before rehydrating the active session.
- If recovery is not possible, the session remains non-tradable/orphaned with clearer status metadata.

### Agent-obvious capability model

- Adds `CapabilityView`/`CapabilityViewEntry` types in `bloom-proto`.
- Adds `/wallets/<wallet>/capabilities/` read surfaces with `active.json` and `active.md`.
- Includes EVM policy-session and Hyperliquid agent-session capability views.
- Adds `unlocked` wallet metadata and clearer locked-wallet remediation paths.
- Vendors agent-facing `/polymarket/README.md` and `/defi/README.md` VFS docs.

### `/next.md`

- Adds a dynamic root `/next.md` aggregator so agents have one obvious place to inspect recommended next actions.
- Adds dynamic root-file VFS support needed for that aggregator.

### Docs and UX clarity

- Updates agent guidance, quickstart/examples, passkey docs, and Hyperliquid docs to explain the capability-first model and session-first trading path.
- Keeps Tier 2 key-residency changes out of this PR.

## Deferred follow-ups

- Tier 2.1: `KeyResidencyLease` + auto-lock on expiry. This changes key lifetime/security behavior and should be reviewed separately.
- Tier 2.2: Polymarket capability primitive.
- Unified writable `/capabilities/` namespace remains deferred; this PR keeps capability surfaces read-only.
- Full API-wallet revoke/replace-agent story is still a follow-up.
- WebSocket VFS surface remains future work.

## Test evidence

Latest local verification after the final commits:

- `cargo clippy -p bloom-vfs --all-targets -- -D warnings`
- `cargo test -p bloom-vfs` — 383 passed, 1 ignored
- Focused Hyperliquid tests are included in the full `bloom-vfs` run, including sealed key round-trip and persisted-bounds recovery coverage.

Earlier broad PR verification before the capability/read-surface layer:

- `cargo test --workspace`
- `cargo nextest run --workspace --no-fail-fast -E 'not (binary(chain_testnet_provision) | binary(chain_smoke))'`
- ignored/acceptance suites and locked release build as previously recorded in CI.

After push, GitHub checks should be treated as the source of truth for the final branch state.


crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -933,37 +1367,49 @@ impl WalletsHandler {
```
josh-richardson, line 932: This direct VFS path mints a bounded EVM policy session after only `keystore.info()`. For a mounted VFS caller, there is no fresh passkey/local unlock ceremony and no reviewed marker, but later `confirm()` treats a matching session as authorization and skips `ensure_action_authorized()`. IPC blocks `policy-session/new` behind `write_unlocked`, but the mounted VFS path still bypasses the ceremony. Please require a fresh reviewed unlock marker/capability here, or make session minting available only through the secure IPC lane. A regression test should assert a plain mounted VFS write to `/wallets/<wallet>/policy-session/new` is refused for passkey wallets unless the review artifact is present.

crates/bloom-vfs/src/handlers/hyperliquid.rs
```diff
@@ -0,0 +1,4376 @@
...
+                    self.append_session_audit(
+                        target.network,
+                        target.wallet,
+                        target.id,
+                        &json!({"event": "blocked_by_session_state", "breach_action": format!("{action:?}")}),
+                    )?;
+                    return Err(HandlerError::invalid(format!(
+                        "Hyperliquid agent session is not tradable: {action:?}"
+                    )));
+                }
```
josh-richardson, line 891: Session actions are evaluated against the wallet's current `[hyperliquid]` policy, not the bounds that were approved and persisted when the API wallet was created. That means a user can approve a narrow session, later broaden `policy.toml`, and the already-approved API wallet will trade under broader authority than the original `approveAgent` ceremony granted. Please split the policy evaluator so owner-signed `/exchange/...` writes reload current verified policy, but agent-session writes evaluate against `active.session.bounds`. Add a test that broadening the wallet policy after session creation does not broaden the live session.

crates/bloom-vfs/src/handlers/hyperliquid.rs
```diff
@@ -0,0 +1,4376 @@
...
+            // failures keep the original timestamp so callers can see staleness age.
+            guard.stale_since_ms.get_or_insert(now);
+        }
+    }
+
+    async fn refresh_session_snapshot(
+        &self,
+        client: &HyperliquidClient,
+        active: &Arc<Mutex<ActiveHlSession>>,
+        orphaned: bool,
```
josh-richardson, line 795: This refresh always snapshots the owner wallet, even though session writes accept `vaultAddress`. Cleanup has the same problem: it collects owner orders/positions and signs forced close/cancel payloads with `vault_address: None`. A session can therefore trade in a vault/subaccount while max position/loss monitoring and cleanup look at or flatten the wrong account. Please either reject `vaultAddress` for agent sessions until supported, or bind one target account/vault at session creation, persist it in `session.json`, use it for snapshots, and include the same vault address in cancel/close/orphan recovery.

crates/bloom-vfs/src/handlers/hyperliquid.rs
```diff
@@ -0,0 +1,4376 @@
...
+            builder: None,
+        };
+        let response = self
+            .submit_owner_orphan(
+                client,
+                network,
+                network_name,
+                wallet,
+                id,
+                &owner_signer,
```
josh-richardson, line 1360: Forced cleanup intentionally skips the trading policy, but this still calls verified `keystore.info()` just to read the owner address. For passkey wallets, `info()` rejects unsigned or stale `policy.toml.sig`, so a policy edit after session creation can block expiry/breach cleanup. The sealed-key recovery path has the same address-only dependency. Please use `info_unverified()` for address-only safety cleanup/recovery/orphan checks, while keeping verified `info()` for paths that authorize new trading.

crates/bloom-proto/src/hyperliquid_policy.rs
```diff
@@ -0,0 +1,727 @@
...
+                None => vec![check(
+                    "action",
+                    PolicyOutcome::Deny,
+                    "usdSend requires transfer_cap_usd in the wallet [hyperliquid] policy",
+                )],
+                Some(cap) => match ctx.notional_microusd {
+                    Some(amount) if amount > cap => vec![check(
+                        "transfer_cap_usd",
+                        PolicyOutcome::Deny,
+                        format!("transfer {} exceeds cap {}", fmt_usd(amount), fmt_usd(cap)),
```
josh-richardson, line 273: `updateLeverage` returns immediately after the leverage cap, before asset allowlist and vault/subaccount checks run. The handler also does not populate the leverage action's asset into the context. As a result, a wallet with `allow_vault_or_subaccount = false` or `allowed_assets = ["BTC"]` can still update leverage for a vault or disallowed asset. Please resolve the leverage asset into the context and apply asset/vault checks before returning from this branch, with unit tests for both denied cases.

crates/bloom-tx/src/session.rs
```diff
@@ -0,0 +1,261 @@
...
+                Some(v) => {
+                    if s.spent_micro_usd.saturating_add(v) > s.max_micro_usd {
+                        // This session can't afford it; keep looking — another
+                        // session might also cover this id.
+                        continue;
+                    }
+                    s.spent_micro_usd = s.spent_micro_usd.saturating_add(v);
+                    v
+                }
+                // Unpriced tx: the explicit id allowlist is the bound; debit 0.
```
josh-richardson: Treating unknown USD value as a zero debit undermines the advertised `max total spend` cap. A value-moving allowlisted tx with missing valuation can be session-authorized and consume no budget, whereas the normal under-policy path fails closed on missing USD valuation. Please pass whether the tx is value-moving into session authorization and reject `None` valuations for value-moving txs, or snapshot each pending tx's micro-USD value when minting the session. Add a regression test for an allowlisted value-moving tx with `usd_value = None`.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -933,37 +1367,49 @@ impl WalletsHandler {
         }
 
         let spec = parse_new_wallet_spec(body)?;
-        let env_pass = std::env::var("BLOOM_PASSPHRASE").ok();
-        let pass = spec
-            .passphrase
```
josh-richardson, line 938: This revokes by session id only, ignoring the wallet segment in the path. If Alice knows Bob's session id, `/wallets/alice/policy-session/<bob-session>/revoke` can revoke Bob's session. Please make revoke wallet-scoped, or look up the session and verify `session.wallet == wallet` before removing it. A two-wallet test would catch this.

crates/bloom-proto/src/hyperliquid_policy.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    // Risk-adding caps below apply only to risk-adding orders.
+    if is_risk_reducing(ctx) {
+        out.push(check(
+            "caps",
+            PolicyOutcome::Pass,
+            "reduce-only order is risk-reducing; notional/position caps not applied",
+        ));
+        return out;
+    }
```
josh-richardson, line 387: `max_position_usd` projects only current position plus this order's notional. Resting open orders are reduced to a count in the snapshot and are not included in projected exposure, so multiple under-cap resting orders can together exceed the intended fillable position cap. Please include open-order notional and/or per-session pending reservations keyed by asset/side, and test two sequential orders that each pass alone but exceed `max_position_usd` together.

crates/bloom-vfs/src/handlers/hyperliquid.rs
```diff
@@ -0,0 +1,4376 @@
...
+            tracing::warn!("hyperliquid.agent_sessions.monitor_skipped: no tokio runtime");
+            return;
+        };
+        handle.spawn(async move {
+            let mut tick = tokio::time::interval(Duration::from_secs(10));
+            loop {
+                tick.tick().await;
+                self.monitor_sessions_once().await;
+            }
+        });
```
josh-richardson, line 232: The wallet capability roll-up only iterates in-memory sessions, so after daemon restart it misses persisted orphaned sessions even though the session status path can discover and render them. It also maps `stale_since_ms` to `CapabilityStatus::Orphaned`, but stale snapshots mean the risk read is stale, not that the daemon lost the agent key. Please enumerate persisted live/orphaned sessions in `capability_views_for` and represent stale risk data separately from key-loss orphaning. Add tests for persisted-orphan inclusion and stale-not-orphan.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -81,6 +109,209 @@ impl WalletsHandler {
...
+
+    fn all_capability_views_for(&self, wallet: &str) -> Vec<CapabilityViewEntry> {
+        let mut all = self.evm_capability_views_for(wallet);
+        if let Some(ref hl) = self.hyperliquid_handler {
+            let hl_views = hl.capability_views_for(wallet);
+            all.extend(hl_views);
+        }
+        all.sort_by(|a, b| {
+            a.created_ms
+                .cmp(&b.created_ms)
```
josh-richardson, line 263: This strips the chain id from the chain-qualified pending key and hardcodes `ethereum`, so a policy session minted for Base/Arbitrum/etc. will point agents at the wrong confirm path. Please derive the chain from the allowlist key and map it to the configured VFS chain name, or store the chain name at mint time. A capability-view test for a non-Ethereum session should assert the rendered `next_write_path` resolves.

crates/bloom/src/main.rs
```diff
@@ -2048,6 +2933,666 @@ fn persist_outbox_review_approved(
...
+    Ok((price, size, notional))
+}
+
+fn format_decimal(value: f64, decimals: usize) -> String {
+    let mut s = if decimals == 0 {
+        format!("{value:.0}")
+    } else {
+        format!("{value:.decimals$}")
+    };
+    if s.contains('.') {
```
josh-richardson, line 3731: `run()` resolves `--connect`, `--ipc-socket`, and endpoint env vars into `client_endpoint`, but the Hyperliquid command path drops it. The session helpers later use `ResolvedEndpoint::default_for_home(...)`, so `bloom --connect unix:/tmp/bloom.sock hyperliquid session status ...` can talk to the default daemon or fail even though a custom daemon was requested. Please pass the resolved endpoint through `handle_hyperliquid` / `handle_hl_session` and use it in the session IPC helpers.

crates/bloom/src/main.rs
```diff
@@ -1253,6 +1953,168 @@ async fn run(cli: Cli) -> Result<()> {
...
+                    .with_context(|| format!("confirm {chain}:{id}"))?;
+                println!(
+                    "broadcast {}:{} hash={}",
+                    chain,
+                    staged.id,
+                    staged.tx_hash.as_deref().unwrap_or("?")
+                );
+            }
+            Ok(())
+        }
```
josh-richardson, line 2204: `request confirm` writes a hardcoded `confirm` through a fresh in-process daemon. That bypasses the running-daemon IPC path, has no unlock/passphrase lane for paid confirms that sign payment credentials, and cannot send the policy override sentinel for warnings. Plain IPC confirms are correctly blocked as signer-consuming, so this CLI should route through `write_unlocked` when confirming paid requests and expose data/override plus unlock options. Please add tests for locked local/passkey paid confirms and warning-sentinel confirms.

josh-richardson (changes requested):
Comments from Codex

PR #59, opened 2026-07-04
feat: wire mounted sealed approval demo

violet-agent:
## Summary
- Adds mounted Sealed Approval challenge projection with `approval_challenge.json` carrying a deterministic `ceremony_url` while keeping the URL out of the signed challenge preimage.
- Reuses unexpired challenge nonces/expiry, makes mounted confirms fail closed after writing challenge artifacts, and mirrors runtime artifacts into the central `/outbox/{pending,sent,failed}/<action_id>` projection.
- Adds a Petal-neutral sealed-action dispatcher with EVM/ERC20 as the first executor, plus mounted approval docs and an ERC20 demo recipe.
- Final gate fixes: central EVM outbox status records Petal identity (`evm-wallet`, placeholder digest, `v0`), and `scripts/acceptance.sh` now exercises the Sealed Approval gate for native + ERC20 local Anvil demos.

## Checklist
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — `docs/specs/2026-07-02-sealed-approval.md`
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — VFS guidance updated; affected Petal READMEs N/A for this central outbox/EVM demo slice

## Verification
- `cargo fmt --check`
- `cargo check -p bloom-tx -p bloom-vfs -p bloom-auth-api -p bloom-auth -p bloom-daemon`
- `cargo test -p bloom-tx --lib` (131 passed)
- `cargo test -p bloom-vfs --lib` (446 passed, 1 ignored live Polymarket API test)
- `cargo test -p bloom-auth-api --lib` (78 passed)
- `cargo test -p bloom-auth --lib` (68 passed)
- `cargo test -p bloom-daemon --lib` (39 passed)
- `cargo test -p bloom-it --test erc20_e2e` (3 passed, local Anvil/Foundry)
- `BLOOM_BIN=$PWD/target/debug/bloom scripts/acceptance.sh` (native + ERC20 Anvil scenarios passed through sealed-approval gate; Uniswap/Enso fork scenarios skipped because `BLOOM_MAINNET_RPC`/`BLOOM_ENSO_KEY` unset)
- `git diff --check`
- Added-line static scan for hardcoded secrets, shell injection, eval/exec, pickle, and SQL-formatting patterns: 0 findings

## Manual / smoke verification
- The acceptance script now boots Anvil, imports the funded dev key with the current local-passphrase CLI, stages a native ETH send, deploys `MockToken`, stages an ERC20 transfer, and verifies for both staged actions:
  - initial confirm fails with `broadcast approval required`;
  - wallet projection writes `approval_challenge.json`;
  - central `central_outbox/pending/<action_id>/approval_challenge.json` exists;
  - central challenge `action_id` matches the directory id;
  - central challenge includes a local `ceremony_url` (`http://localhost...` or `http://127.0.0.1...`).
- Unit/integration coverage verifies grant-only retry and grant+execute execution paths, Petal-neutral dispatcher rejection for unregistered Petals, sent/failed projection consistency with result/status artifacts, and ERC20 owner-session execution.

## Manual-only gaps
- Full browser/passkey ceremony UX and kernel-mounted NFS `EACCES` behavior were not exercised in this headless worker. The script verifies the file-backed central/wallet challenge projection and CLI denial boundary; browser-mediated `grant` / `grant + execute` are covered by focused tests rather than a live passkey ceremony.


crates/bloom-tx/src/tx_engine.rs
```diff
@@ -2315,17 +2903,31 @@ impl TxEngine {
                 TxEngineError::BroadcastApprovalRequired(format!(
                     "issue outbox approval challenge failed: {e}"
                 ))
-            })?;
+            })?
+            .with_local_ceremony_url();
```
josh-richardson, line 3053: This method now issues the mounted challenge, but the earlier live-grant branch in the same method accepts any active grant for `(wallet, action_id, petal_id, petal_digest)` before rebuilding the current sealed action. It then returns the intent hash for the *current* staged bytes, without checking that the live grant's `intent_hash` matches that recomputed value. If the staged tx/signing hash changes under the same action id, the old approval can authorize the new bytes. Please compute the current sealed action hash before accepting the grant and require `grant.intent_hash == current_intent_hash`; a regression test with a tuple-matching but intent-mismatching grant would catch this.

crates/bloom-daemon/src/ceremony_server.rs
```diff
@@ -0,0 +1,636 @@
...
+        .verify_and_mint_grant(signed.clone(), grant_store.as_ref(), now)
+        .await
+    {
+        Ok(g) => g,
+        Err(e) => return err_json(StatusCode::UNAUTHORIZED, format!("verify approval: {e}")),
+    };
+
+    // Cache the decrypted signer under the grant so the sign_hash path can use
+    // it without a fresh ceremony.
+    daemon.signer_cache.insert(
```
josh-richardson, line 425: This caches signer material under the grant for the generic `PetalHost::sign_hash` path, but that primitive signs the caller-provided `hash_hex` after tuple/intent/schema checks; it does not compare the requested hash to the attested `signing_hash` or to the sealed subject. The EVM sealed dispatcher recomputes before calling the host, but the host/IPC boundary is still a raw-hash signing primitive for any caller with the live grant tuple. Please enforce `request.hash_hex == attested signing_hash` for EVM, and an equivalent per-schema hash binding for other Petals, before exposing cached signer material to this path.

crates/bloom-tx/src/tx_engine.rs
```diff
@@ -1953,6 +2161,386 @@ impl TxEngine {
...
+        {
+            return Err(TxEngineError::BroadcastHashMismatch {
+                expected: format!("{:#x}", signed.hash),
+                returned: submitted
+                    .returned_hash
+                    .map(|h| format!("{:#x}", h))
+                    .unwrap_or_else(|| "<none>".into()),
+            });
+        }
+        self.outbox.transition_central_action(
```
josh-richardson: `grant_execute` reaches this sealed execution path with only the `SealedAction`, so this transition moves the central action but never moves the original per-wallet pending outbox directory. Normal confirms use `Outbox::transition`, which moves the wallet entry and mirrors central state; this path constructs a transient `StagedTx` and leaves `/wallets/<wallet>/chains/<chain>/outbox/pending/...` stale with the challenge/approval artifacts. Please transition or terminally update the matching wallet projection as part of sealed execution.

crates/bloom-daemon/src/ceremony_server.rs
```diff
@@ -0,0 +1,636 @@
...
+            .tx_engine
+            .execute_sealed_action(&action, &chain_name, &client, &policy)
+            .await
+        {
+            Ok(exec) => {
+                executed = true;
+                tx_hash = Some(format!("{:#x}", exec.tx_hash));
+            }
+            Err(e) => {
+                return err_json(
```
josh-richardson: At this point the approval nonce has been consumed, the grant has been minted, and signer material has been cached. Returning an execute error here leaves a live grant/cache entry and no failed projection/result, so user-selected `grant_execute` can silently degrade into an errored grant-only state with stale pending outbox artifacts. Please either revoke/drop the grant and cache entry on execute failure, or intentionally return a grant-only success and project the failure state/result so the action is not left ambiguous.

crates/bloom-tx/src/tx_engine.rs
```diff
@@ -1570,9 +1630,57 @@ impl TxEngine {
...
+            "action_id": central.action_id,
+            "state": central.state.dirname(),
+            "outcome": central.outcome,
+            "action_kind": central.action_kind,
+            "tx_hash": format!("{:#x}", central.tx_hash),
+            "nonce": central.nonce,
+            "signing_hash": format!("{:#x}", central.signing_hash),
+            "created_ms": now_ms(),
+        });
+        let status = serde_json::json!({
```
josh-richardson: The central pending `status.json` is initially staged with Petal identity (`petal_id`, `petal_digest`, `petal_digest_kind`, `petal_version`), but this writer replaces the status with a smaller object that drops those fields. That undercuts the PR's central-outbox identity/audit story exactly when the action reaches its final state. Please merge with the existing status or include the Petal identity fields in the final status.

crates/bloom-tx/src/tx_engine.rs
```diff
@@ -1953,6 +2161,386 @@ impl TxEngine {
...
+            std::fs::write(dir.join(OUTBOX_APPROVAL_FILE), &body).map_err(|e| {
+                TxEngineError::BroadcastApprovalRequired(format!("write approval.json: {e}"))
+            })?;
+            persisted = true;
+        }
+        // Central outbox projection, keyed by the sealed action id (same place
+        // the challenge is mirrored). No-op when no projection is attached, but
+        // a real write error must not be swallowed — a silent allowlist reject
+        // is exactly what previously kept approval.json out of the central dir.
+        if let Err(e) = self.outbox.write_central_action_artifact(
```
josh-richardson: The comment says a real central write error must not be swallowed, but this branch catches the error, logs a warning, and still returns `Ok(())`. That means a completed ceremony can miss the central `approval.json` artifact even though the central outbox is meant to be the canonical cross-surface projection. Please propagate the error, or explicitly downgrade both the comment and contract to best-effort mirroring.

crates/bloom-daemon/src/ceremony_server.rs
```diff
@@ -0,0 +1,636 @@
...
+    let addr = format!("127.0.0.1:{LOCAL_CEREMONY_PORT}");
+    let listener = match tokio::net::TcpListener::bind(&addr).await {
+        Ok(l) => l,
+        Err(e) => {
+            tracing::warn!(
+                addr = %addr,
+                err = %e,
+                "ceremony.server.bind_failed; mounted Sealed Approval URLs will be unreachable"
+            );
+            return CeremonyServer {
```
josh-richardson: Continuing with an inert handle means Bloom will still emit `http://localhost:18734/ceremony/...` in later `approval_challenge.json` files even though this daemon did not bind that address. At best the URL is known-broken; at worst another local process owns the port and the user opens a non-Bloom service at a Bloom approval URL. Please fail `serve`, disable URL projection, or make confirm fail with a clear ceremony-server-unavailable error when this bind fails.

docs/architecture/Wallet.md
```diff
@@ -0,0 +1,282 @@
...
+
+Petals and VFS handlers must not use the wallet key directly.
+
+Allowed path:
+
+```text
+active Sealed Approval Grant
+  -> host validates wallet/action/petal/digest/intent/terms/policy/expiry/count
+  -> Petal provides structured signing attestation
+  -> host signs the exact hash or performs the exact authority use
```
josh-richardson: This says the host signs the exact hash, but the current generic host signing API signs `SignHashRequest.hash_hex`; exact-hash binding is enforced by selected callers like the EVM sealed dispatcher, not by the grant/signing primitive itself. Either make the primitive enforce the binding, or soften this doc to say the caller/Petal must prove the requested hash matches the approved sealed subject.

docs/architecture/Open-Internet Sealed Approval Ceremony.md
```diff
@@ -1,20 +1,24 @@
...
-[`Interaction Modes.md`](./Interaction%20Modes.md) and applies in both
-exposure modes.
+The open-internet relay described here is not implemented today. The current
+mounted-VFS implementation supports daemon-owned loopback ceremony URLs on
+`http://localhost:18734`: `approval_challenge.json` carries a local
+`ceremony_url`, the token is derived from `server_nonce`, and the daemon owns
+grant minting for the mounted EVM slice. This document is the target design for
+making that same ceremony reachable from another device over the open internet.
+The `ceremony_url` contract itself — the field in `approval_challenge.json`,
+single-use token, `expiry_ms` bound — is defined in
```
josh-richardson, line 18: The current loopback server allows repeated page/plan/challenge GETs until completion or expiry; the token becomes gone only after nonce consumption/expiry. If `single-use` means one successful approval completion, please say that explicitly. If it means first-open claims the ceremony, the current code does not implement that behavior.

josh-richardson (changes requested):
Focused review of the mounted Sealed Approval runtime/docs. The core direction looks useful, but the grant/signing and projection invariants need tightening before merge.

crates/bloom-tx/src/tx_engine.rs
GavinPacini: Thanks. I rechecked this path. I am not changing this in this PR: for the mounted EVM flow, the sealed execution path reconstructs the EVM subject and verifies the signing hash before host signing/execution, so the EVM Petal path already enforces the current bytes. We are intentionally leaving broader tuple/live-grant tightening out of this PR unless we decide to harden the generic grant-store contract separately.

crates/bloom-daemon/src/ceremony_server.rs
GavinPacini: Leaving this as-is by design. Petals are allowed to propose arbitrary bytes at this trust boundary; the host enforces grant envelope, Petal identity, expiry/count, and records the structured attestation/audit, but it does not independently prove every schema hash binding. I updated the docs to describe that boundary rather than claim universal exact-hash enforcement.

crates/bloom-tx/src/tx_engine.rs
GavinPacini: Fixed. Sealed EVM execution now locates the matching per-wallet pending outbox entry and transitions it to `sent`, rewriting `intent.json` and `tx_hash` there. It falls back to central-only transition only if no matching wallet entry exists.

crates/bloom-daemon/src/ceremony_server.rs
GavinPacini: Fixed. Post-mint execute failures now revoke the minted grant, drop cached signer material, and project a failed terminal result/state where the chain projection can be resolved. Approval artifact persistence failure also now revokes/drops instead of continuing with a live grant.

crates/bloom-tx/src/tx_engine.rs
GavinPacini: Fixed. Final EVM result/status writing now reads and merges the existing central `status.json` instead of replacing it with a smaller object, so Petal identity fields from the pending projection survive terminal status updates.

crates/bloom-tx/src/tx_engine.rs
GavinPacini: Fixed. `persist_outbox_ceremony_approval` now propagates central `approval.json` write errors. The ceremony server treats that as a post-mint failure and revokes/drops the grant/cache instead of silently continuing.

crates/bloom-daemon/src/ceremony_server.rs
GavinPacini: Fixed. The ceremony server `spawn` now returns an error if the loopback bind fails, and `bloom serve` fails startup with context instead of later projecting known-broken `ceremony_url`s.

docs/architecture/Wallet.md
GavinPacini: Fixed in docs. The wallet architecture now describes the intended boundary: Bloom Machine enforces the grant envelope and signs the hash the Petal presents while binding/auditing the Petal structured attestation, rather than claiming the generic primitive independently proves exact-hash binding.

docs/architecture/Open-Internet Sealed Approval Ceremony.md
GavinPacini: Fixed in docs. The open-internet ceremony doc now matches the implemented loopback semantics: repeated page/plan/challenge reads remain valid while the challenge is pending, and the token is considered used only after successful approval completion consumes the underlying challenge nonce.

PR #66, opened 2026-07-06
Wire Hyperliquid sealed approval signing

josh-richardson:
## Summary

- route Hyperliquid approveAgent and usdSend through grant-backed PetalHost signing when auth services are wired
- stage Hyperliquid SealedActions with explicit signing intents and persistent pending agent keys
- make the ceremony server treat Hyperliquid approvals as grant-only instead of EVM outbox artifacts

## Checklist

- [x] Architecture docs (docs/architecture/) updated if contracts or behavior changed — N/A, implementation wires existing Sealed Approval/PetalHost contracts without changing architecture docs contracts
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — see docs/architecture/Sealed Approvals.md
- [x] Agent Documentation updated (crates/bloom-vfs/src/docs/agent-guidance.md and affected Petal READMEs) — N/A, no user-facing agent guidance text changed

## Tests

- cargo check -p bloom-auth-api -p bloom-hyperliquid -p bloom-vfs -p bloom-daemon
- cargo test -p bloom-hyperliquid -p bloom-auth-api -p bloom-vfs hyperliquid --lib
- cargo test -p bloom-daemon ceremony --lib
- cargo clippy --workspace --all-targets -- -D warnings


GavinPacini:
This is broadly moving in the right direction: owner-authorized Hyperliquid actions are routed through Sealed Approval / `PetalHost::sign_hash`, and the agent-session key is Hyperliquid-specific rather than owner key
  material.

  One refinement I think we need before merging: the `approveAgent` sealed subject should commit to the exact venue-signing inputs that will be signed after approval. Right now `create_agent_session` signs an `approveAgent`
  payload using a fresh `nonce` and the resolved `agent_name`, but the sealed agent-session envelope appears to store the raw request fields and does not include the venue nonce. That leaves a gap where the ceremony
  approves “this session request” rather than the exact Hyperliquid EIP-712 `approveAgent` message/hash.

  Can we make the agent-session flow mirror the stronger shape used by `usdSend` / `/requests`?

  - Persist or derive stable `approveAgent` signing inputs at staging time.
  - Include the exact `nonce`, resolved `agent_name`, agent address, network, session id, and any other EIP-712 message fields in the sealed subject.
  - On execution, reconstruct/validate the `approveAgent` action from the sealed subject before calling `PetalHost::sign_hash`.
  - Reuse the live ceremony-token challenge check used by `usdSend`, rather than only trusting a local reusable challenge file.

  That would make the approval bind to the actual owner-signed Hyperliquid action, not just to a nearby session-intent projection.

PR #72, opened 2026-07-07
feat: /requests (x402 + MPP) and VFS policy editing via Sealed Approval

GavinPacini:
## Summary

Builds on the Sealed Approval ceremony infrastructure (base branch) to bring the paid-HTTP `/requests` surface and in-VFS wallet policy editing onto the ceremony model, and closes the mount/IPC integration gaps that previously blocked those writes from reaching their handlers.

**Paid HTTP requests (x402 + MPP/Tempo)**
- Confirming a paid request is now a first-party Sealed Approval action: the first `confirm` write stages an approval challenge (with a local `ceremony_url`); the x402/Tempo credential is signed only under a grant-gated `PetalHost` signature and retried from sealed bytes. Credential material stays redacted in the VFS.
- MPP/Tempo (chainId 4217) charge flow wired end-to-end alongside x402.

**Wallet policy editing in the mounted VFS**
- Editing `wallets/<w>/policy.toml` is a Sealed Approval action that installs the updated `policy.toml` plus its matching `policy.toml.sig`; a `policy-updates/` surface exposes the challenge/status (no secrets). Ordinary read/sign/broadcast paths keep verifying `policy.toml.sig`.

**Integration-gap fixes (what makes the flows actually work over the NFS mount)**
- Removed the coarse signer-path gates that denied `policy.toml` and `requests/…/confirm` before they reached the VFS handler, in both the IPC (`do_write`) and mount adapter lanes. The handler's own Sealed Approval gate is the authority; `write_unlocked` remains disabled for passkey wallets.
- Ceremony server no longer routes non-broadcast (non-tx) actions through the tx-engine outbox — wallet-policy and paid-HTTP confirms are grant-only.
- Mount: pending request listings advertise `confirm`/`cancel`; `response` resolves as a directory via `lookup` (previously reported as a 0-byte file, which surfaced as ENOTDIR on the mount).

Manually verified end-to-end against a live daemon: staged and confirmed both an x402 and an MPP/Tempo request to `api.nansen.ai` — ceremony → grant → signed credential → `200` response, with credential material kept redacted.

## Checklist

      behavior changed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/bloom-keystore/src/petal_host.rs
+++ b/crates/bloom-keystore/src/petal_host.rs
@@ -391,6 +391,68 @@ impl PetalHost for KeystorePetalHost {
                 )
                 .await;
         }
+        if attestation.petal_id == bloom_auth_api::petal_identity::PETAL_ID_PAID_HTTP {
+            if attestation.intent != request.intent {
+                return self
+                    .deny(
+                        "petal.sign.deny",
+                        Some(request.wallet.clone()),
+                        Some(request.action_id.clone()),
+                        "attestation intent does not match sign-hash request intent",
+                        Some(serde_json::json!({
+                            "request_intent": request.intent,
+                            "attestation_intent": attestation.intent,
+                        })),
+                    )
+                    .await;
+            }
+            if attestation_fact_str(attestation, "wallet") != Some(request.wallet.as_str()) {
+                return self
+                    .deny(
+                        "petal.sign.deny",
+                        Some(request.wallet.clone()),
+                        Some(request.action_id.clone()),
+                        "attestation wallet does not match sign-hash request wallet",
+                        None,
+                    )
+                    .await;
+            }
+            if attestation_fact_str(attestation, "action_id") != Some(request.action_id.as_str()) {
+                return self
+                    .deny(
+                        "petal.sign.deny",
+                        Some(request.wallet.clone()),
+                        Some(request.action_id.clone()),
+                        "attestation action_id does not match sign-hash request action_id",
+                        None,
+                    )
+                    .await;
+            }
+            if !attestation_signing_hash_matches(attestation, &request.hash_hex) {
+                return self
+                    .deny(
+                        "petal.sign.deny",
+                        Some(request.wallet.clone()),
+                        Some(request.action_id.clone()),
+                        "attestation signing_hash does not match sign-hash request hash",
+                        None,
+                    )
+                    .await;
+            }
+            if attestation_fact_str(attestation, "policy_snapshot_digest")
+                != Some(grant.petal_policy_digest.as_str())
+            {
+                return self
+                    .deny(
+                        "petal.sign.deny",
+                        Some(request.wallet.clone()),
+                        Some(request.action_id.clone()),
+                        "attestation policy_snapshot_digest does not match the sealed grant",
+                        None,
+                    )
+                    .await;
+            }
+        }
 
         // Step 7: ensure the signer is in memory. A per-grant signer cache
         // (set by the daemon after `sealed_approval_ceremony`) lets us reuse
@@ -587,6 +649,23 @@ fn parse_hash_hex(s: &str) -> Result<B256, String> {
     let mut out = [0u8; 32];
     out.copy_from_slice(&bytes);
     Ok(B256::from(out))
+}
+
+fn attestation_fact_str<'a>(attestation: &'a SigningAttestation, key: &str) -> Option<&'a str> {
+    attestation.facts.get(key).and_then(|v| v.as_str())
+}
+
+fn attestation_signing_hash_matches(attestation: &SigningAttestation, request_hash: &str) -> bool {
+    let Some(attested) = attestation_fact_str(attestation, "signing_hash") else {
+        return false;
+    };
+    strip_0x(attested).eq_ignore_ascii_case(strip_0x(request_hash))
+}
+
+fn strip_0x(s: &str) -> &str {
+    s.strip_prefix("0x")
+        .or_else(|| s.strip_prefix("0X"))
+        .unwrap_or(s)
 }
 
 // `SealedApprovalGrant` must remain in scope for downstream use (the host
--- a/crates/bloom-mount/src/adapter.rs
+++ b/crates/bloom-mount/src/adapter.rs
@@ -189,12 +189,11 @@ fn mount_write_path_uses_wallet_signer(path: &VfsPath) -> bool {
         {
             true
         }
-        [root, _wallet, leaf] if root == "wallets" && leaf == "policy.toml" => true,
-        [root, _wallet, chains, _chain, leaf]
-            if root == "wallets" && chains == "chains" && leaf == "policy.toml" =>
-        {
-            true
-        }
+        // policy.toml writes flow through to the VFS wallets handler, which
+        // stages a first-party Sealed Approval for passkey wallets (challenge +
+        // grant-gated install) and writes local policy immediately. They no
+        // longer route through the disabled write_unlocked re-sign lane, so the
+        // mount must forward them to `vfs.write` rather than deny on flush.
         _ => false,
     }
 }
--- a/crates/bloom-mount/src/adapter.rs
+++ b/crates/bloom-mount/src/adapter.rs
@@ -152,15 +152,6 @@ fn mount_write_path_uses_wallet_signer(path: &VfsPath) -> bool {
         {
             true
         }
-        // Confirming a paid HTTP request is a first-party Sealed Approval action:
-        // the requests handler stages an approval challenge on the first write and
-        // only signs under a grant-gated PetalHost signature, so it must reach the
-        // VFS handler rather than be denied at the mount lane (mirrors policy.toml).
-        [root, _wallet, ps, leaf]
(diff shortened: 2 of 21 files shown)


crates/bloom-keystore/src/petal_host.rs
```diff
@@ -391,6 +391,68 @@ impl PetalHost for KeystorePetalHost {
                 )
                 .await;
         }
+        if attestation.petal_id == bloom_auth_api::petal_identity::PETAL_ID_PAID_HTTP {
```
josh-richardson, line 394: This adds paid-HTTP-specific self-consistency checks, but it still does not bind the requested hash/facts back to the sealed `/requests` subject. A local IPC caller can call `sign_hash` while a one-shot paid-HTTP grant is live and provide any 32-byte hash plus matching `wallet`/`action_id`/`signing_hash`/`policy_snapshot_digest` facts; the host will sign it without checking URL, amount, payee, selected requirement, or protocol against the approved sealed action. This should load/validate against the sealed action or otherwise restrict this path so only the protocol adapter can consume the grant for the approved payment digest.

josh-richardson (commented):
Requesting changes for concrete correctness/security issues in the Sealed Approval request/policy flows. See inline comments.

crates/bloom-vfs/src/handlers/requests.rs
```diff
@@ -634,49 +727,66 @@ impl RequestsHandler {
...
+            );
             let backend = RealMppBackend {
-                keystore: self.keystore.clone(),
                 client: self.client.clone(),
                 rpc_resolver: Arc::clone(&self.paid_http_rpc_resolver),
+                wallet_address,
+                host_signer,
+                facts,
             };
             let result = confirm_with_backend(
```
josh-richardson, line 780: The MPP branch validates `sealed_execution_inputs` above, but then `confirm_with_backend` re-reads `challenge.json` and `request.toml` from the mutable pending directory before preparing/signing/retrying. That reopens the projection-drift/TOCTOU problem this branch is trying to close: an interleaving write after sealed validation can affect the MPP payment digest/retry. Please pass the sealed request/challenge/requirement into this helper/backend path instead of re-reading projection artifacts.

crates/bloom/src/main.rs
```diff
@@ -3151,12 +3147,16 @@ async fn sign_request_sealed_approval_if_challenged(
...
+        &d.keystore,
+        &d.auth_services,
+        unsigned,
+        intent,
+        cli_now_ms(),
+        d.signer_cache.as_ref(),
+    )
+    .await
+    .context("run request sealed approval browser ceremony")?;
     let approval_path = dir.join("approval.json");
```
josh-richardson, line 3211: This writes the signed request approval under `requests/pending/<id>/approval.json`, and `RequestsHandler` currently serves arbitrary files under `pending`/`sent`/`failed` while finalization renames the whole directory. That exposes the signed approval/WebAuthn assertion through the mounted `/requests` tree. Please either keep this artifact outside the mounted request directory, remove it after minting the grant, or explicitly hide `approval.json` from `lookup`/`list`/`read` like `private`.

crates/bloom/src/main.rs
```diff
@@ -1574,9 +1572,7 @@ async fn run(cli: Cli) -> Result<()> {
             let info = d.keystore.info(&wallet)?;
             let passkey_wallet = matches!(info.kind, bloom_keystore::WalletKind::PasskeyGated);
             match info.kind {
-                bloom_keystore::WalletKind::PasskeyGated => {
-                    bail!(PASSKEY_WRITE_UNLOCKED_DISABLED);
-                }
+                bloom_keystore::WalletKind::PasskeyGated => {}
```
josh-richardson, line 1626: This fixes only the in-process fallback. When a daemon socket is present, the CLI still sends request confirmation via `write_unlocked` above, but the daemon now rejects passkey `write_unlocked` before `RequestsHandler` and the new request flow expects plain `write` to stage/read the Sealed Approval challenge. `bloom request confirm` should use plain IPC `write` for `/requests/.../confirm` and then drive the challenge/retry path on `PermissionDenied`.

crates/bloom-mount/src/adapter.rs
```diff
@@ -152,12 +152,10 @@ fn mount_write_path_uses_wallet_signer(path: &VfsPath) -> bool {
...
-        [root, state, _id, action]
-            if root == "requests" && state == "pending" && action == "confirm" =>
-        {
-            true
-        }
+        // Confirming a paid HTTP request is a first-party Sealed Approval action:
+        // the requests handler stages an approval challenge on the first write and
+        // only signs under a grant-gated PetalHost signature, so it must reach the
+        // VFS handler rather than be denied at the mount lane (mirrors policy.toml).
         [root, _wallet, ps, leaf]
```
josh-richardson: This still treats `/wallets/<wallet>/policy-session/new` as signer-consuming at the mount layer, so the write is denied before `WalletsHandler` can stage its Sealed Approval challenge. Since passkey `write_unlocked` is hard-disabled, policy-session minting for passkey wallets remains unreachable through the mounted/plain lane. Either let this reach the VFS handler like `policy.toml`/request confirm, or document and test that this flow is intentionally unavailable from the mount.

josh-richardson (changes requested):
Requesting changes. I left inline comments on the concrete blockers: raw `sign_hash` grant consumption, daemon-backed `bloom request confirm`, request approval artifact exposure, MPP sealed/projection re-read, and mounted policy-session reachability.

crates/bloom/src/main.rs
GavinPacini: `approval.json` carries a WebAuthn assertion but no key material/PRF/grant, and EVM outbox intentionally exposes the same artifact. `/requests` is aligned with EVM Petal here. If we want wallet-policy-style hiding, that should be a broader artifact policy change rather than a request-specific security fix.

crates/bloom-keystore/src/petal_host.rs
GavinPacini: This is the intended PetalHost::sign_hash trust model: the host signs a 32-byte digest under a live (wallet, action_id, petal_id, petal_digest) grant and an attestation. The /requests handler constructs the paid-http hash/facts from the sealed request, matching the EVM tx and wallet-policy patterns where the engine/handler owns digest construction. We should add paid-http fact-shape validation for parity, but host-side reconstruction of each protocol digest is outside the current PetalHost contract.

crates/bloom-vfs/src/handlers/requests.rs
GavinPacini: Fixed. `confirm_with_backend` now takes `sealed_execution: Option<ConfirmExecutionInputs>` (references to the already-validated `ParsedRequest`, `NormalizedChallenge`, `PaymentRequirement`, wallet, and dry-run state). When present it uses those bytes instead of re-reading `challenge.json`/`request.toml`, so the MPP path no longer re-projects the mutable pending directory before prepare/sign/retry. The MPP branch threads its sealed values in; x402 was already using `sealed_inputs.*` directly. The direct-`confirm_with_backend` fallback (no sealed inputs) is preserved for the existing tests. Added a regression test — `mpp_confirm_uses_sealed_execution_not_tampered_projection` — that tampers the pending `challenge.json`/`request.toml` after sealed validation and asserts the backend/retry hit the sealed target (sealed mock server hit once, tampered server never contacted).

crates/bloom/src/main.rs
GavinPacini: Fixed. Daemon-backed `bloom request confirm` now sends a plain IPC `write` to `/requests/<id>/confirm` so it reaches `RequestsHandler` and stages the Sealed Approval challenge. On a handler `PermissionDenied` it runs the existing request ceremony helper (`sign_request_sealed_approval_if_challenged`) to write `approval.json` to the shared home, then retries the same plain `write` so the daemon verifies and consumes it. The in-process fallback (no daemon socket) is unchanged. Note the ceremony helper is run against a read-only `Daemon::from_home` in this path because the serving daemon holds the home write lock — that's enough to drive the WebAuthn ceremony and drop `approval.json`; the serving daemon re-verifies it on retry. Daemon-side classification already covers `/requests/.../confirm` reaching plain `write` (`plain_ipc_write_rejects_signer_consuming_paths`).

crates/bloom-mount/src/adapter.rs
GavinPacini: Fixed — let it reach the handler like `policy.toml`. Removed the `policy-session/new` arm from `mount_write_path_uses_wallet_signer` (mount) and the matching arm from `write_path_uses_wallet_signer` (daemon IPC), so both mounted and plain-IPC writes forward to `WalletsHandler`, which stages the Sealed Approval challenge and mints the bounded session only under a grant-gated signature. Handler-side enforcement is unchanged (the `wired_auth_policy_session_mint_*` tests still pass). Classification tests updated in both crates: `policy-session/new` is now asserted non-signer-consuming, while the raw signer lanes (`sign/{message,hash,typed_data}`, outbox `cancel`/`replace`, HL exchange order/cancel/etc.) remain denied — see the new `mount_classifier_forwards_handler_owned_sealed_approval_writes` and the updated `plain_ipc_write_rejects_signer_consuming_paths`.

PR #97, opened 2026-07-09
[vfs] Surface natural metadata timestamps

### LINKED ISSUE #70: [vfs]: Timestamps for files exposed over NFS / VFS should be based on the file / artifact
When `ls`ing inside dynamic dirs e.g. `/bloom/requests/pending`, the timestamp returned over NFS should match the request timestamp. Many dynamic files and folders have some form of a timestamp (evm transaction creation time, evm transaction confirmed time, http request staged time, http request confirmed time), we need to surface this timestamp over the NFS protocol so navigating the files and folders feels more natural relative to the underlying data.

### LINKED ISSUE #95: VFS chain tx reads fail on Base system tx type 0x7e
# Base system transaction decode in VFS tx views

Manual VFS testing of timestamp metadata on Base found that some system/deposit
transactions use transaction/receipt type `0x7e`. The current chain tx VFS reads
route through Alloy's typed transaction/receipt decoders, which reject that type:

```text
unknown variant `0x7e`, expected one of `0x0`, `0x1`, `0x2`, `0x3`, `0x4`
```

Affected examples:

- `/chains/base/tx/<hash>/full.json`
- `/chains/base/tx/<hash>/receipt.json`
- receipt-derived leaves such as `status`, `block_number`, `gas_used`, and
  `logs.json`

This also means timestamp metadata for those tx entries falls back to synthetic
metadata time, because the containing block cannot be derived through the typed
receipt path.

Likely fix: add a raw JSON-RPC fallback for unknown transaction/receipt types so
the VFS can still expose raw JSON and derive `blockNumber` / `blockTimestamp`
without requiring Alloy to decode the transaction into a known typed receipt.

### LINKED ISSUE #96: Add non-mounted VFS stat or clearer mount permission UX
# VFS stat without privileged kernel mount

## Problem

Manual verification of VFS metadata currently requires a real mounted filesystem
and host `stat`. On Linux, `bloom serve --mount <path>` can start Bloom's NFS
server successfully, but the kernel mount still fails for an ordinary user:

```text
mount: <path>: must be superuser to use mount
sudo: a password is required
```

This blocks verification of metadata-only behavior, such as issue #70 timestamp
surfacing, in environments without root/passwordless sudo.

This is separate from PR #92 / issue #77, which improves mounted write buffering
and permission/challenge surfacing after a mount exists. The failure here
happens at the OS mount permission boundary before normal mounted VFS operations
can be tested.

## Impact

- Agents and CI-like environments may be able to use `bloom vfs cat` and
  `bloom vfs ls`, but not verify `mtime`, `ctime`, mode, or size through `stat`.
- Manual QA for VFS metadata changes depends on host sudo configuration rather
  than Bloom behavior.
- Mount failures are technically correct but not very actionable for users
  trying to verify VFS metadata.

## Possible fixes

- Add `bloom vfs stat <path>` to expose VFS `Entry` metadata without a kernel
  mount.
- Improve `bloom serve --mount` failure output with concrete Linux/macOS setup
  guidance for NFS mount permissions.
- Optionally add a small metadata smoke command that compares VFS lookup/list
  metadata with mounted `stat` when a mount is available, but degrades
  gracefully when it is not.

## Acceptance criteria

- A user without root can inspect VFS metadata for a path, including kind, mode,
  size, and modified time.
- `bloom serve --mount` permission failures clearly explain that the NFS server
  started but the OS refused the kernel mount.
- This remains separate from mounted write/permission-denial behavior covered by
  #77 / PR #92.

0xdewy:
## Summary

- add optional `Entry.modified` metadata and teach the mount adapter to expose it through file attrs
- surface natural timestamps for request artifacts, outbox artifacts, wallet EVM outbox actions, DeFi sessions, Polymarket drafts/receipts, Hyperliquid sessions, simulate sessions, watch specs/artifacts, and chain RPC metadata
- make chain head/block/confirmed-tx timestamps best-effort from RPC block timestamps, falling back to the previous synthetic timestamp behavior when unavailable
- expose Base-style unknown typed chain tx/receipt JSON through raw RPC fallback so `0x7e` system/deposit tx reads work
- ignore local `cache/` and `out/` directories

## Notes

- Direct VFS reads were manually exercised for chain head/block/tx metadata and existing artifact-backed paths.
- Real Base `0x7e` transaction reads were manually exercised through `bloom vfs cat` using tx `0xc2080cd9bfe974c488e3d78ee60f04d8fbeb93042d668b5dafd16e27179fecc8`.
- Mounted `stat` could not be verified in this environment because kernel mounting requires superuser/passwordless sudo. The non-mounted `vfs stat` command from #99 was merged into this branch and ships with this PR.
- Closes #70. Fixes #95. Closes #96.

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed - N/A, metadata timestamps and raw tx fallback use existing VFS entry/read contracts
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) - no signing paths changed
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) - N/A, no agent command surface or Petal guidance changed

## Tests

- `cargo check -p bloom-vfs`
- `cargo check -p bloom-mount --features mount`
- `cargo test -p bloom-vfs handlers::chains::tests --lib`
- `cargo test -p bloom-vfs handlers::outbox::tests --lib`
- `cargo test -p bloom-vfs handlers::simulate::tests --lib`
- `cargo test -p bloom-vfs handlers::watch::tests --lib`
- `cargo test -p bloom-vfs handlers::wallets::tests --lib`
- `cargo test -p bloom-vfs handlers::requests::tests --lib`
- `cargo test -p bloom-vfs handlers::defi::tests --lib`
- `cargo test -p bloom-vfs handlers::polymarket::tests --lib`
- `cargo test -p bloom-vfs handlers::hyperliquid::tests --lib`
- `cargo test -p bloom-mount --features mount adapter::tests --lib`
- `cargo test -p bloom-vfs unknown_typed_tx_receipt_falls_back_to_raw_json -- --nocapture`
- `cargo test -p bloom-vfs handlers::chains::tests:: -- --nocapture`
- `cargo test -p bloom-evm`
- `git diff --check`
- Manual: `bloom --home /tmp/bloom-base-vfs-test vfs cat /chains/base/tx/0xc2080cd9bfe974c488e3d78ee60f04d8fbeb93042d668b5dafd16e27179fecc8/{full.json,receipt.json,status,block_number,gas_used,logs.json,error.json}`

### DIFF
--- a/crates/bloom-vfs/src/handlers/chains.rs
+++ b/crates/bloom-vfs/src/handlers/chains.rs
@@ -100,6 +100,11 @@ pub struct ChainsHandler {
     /// configured. Keys are chain names (e.g., "ethereum").
     mempool_handlers:
         Arc<std::collections::BTreeMap<String, Arc<super::chains_mempool::MempoolHandler>>>,
+    /// Block timestamps backing directory-listing metadata, so kernel
+    /// readdir/getattr bursts don't fan out into repeated RPC fetches.
+    /// Head timestamps expire on a short TTL (heads advance);
+    /// mined-block timestamps are immutable, so only map size is bounded.
+    block_ts_cache: Arc<PlMutex<BlockTsCache>>,
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
@@ -113,6 +118,22 @@ struct TokenMetadata {
     decimals: u8,
     symbol: String,
     expires_at: Instant,
+}
+
+/// TTL for cached head-block timestamps in directory listings. Short
+/// enough that `ls` metadata tracks the chain head, long enough that a
+/// burst of readdir/getattr traffic collapses into one RPC fetch.
+const HEAD_TS_TTL: Duration = Duration::from_secs(5);
+
+/// Cap on cached per-number block timestamps. Entries are immutable so
+/// eviction is purely a memory bound; the map is cleared wholesale when
+/// full (a rebuild costs one RPC per listed block directory).
+const BLOCK_TS_CACHE_CAP: usize = 4096;
+
+#[derive(Default)]
+struct BlockTsCache {
+    head: std::collections::HashMap<String, (Instant, u64)>,
+    by_number: std::collections::HashMap<(String, u64), u64>,
 }
 
 impl ChainsHandler {
@@ -131,6 +152,7 @@ impl ChainsHandler {
             revert_decoder: Arc::new(DecoderChain::new()),
             revert_cache: Arc::new(PlMutex::new(std::collections::HashMap::new())),
             mempool_handlers: Arc::new(std::collections::BTreeMap::new()),
+            block_ts_cache: Arc::new(PlMutex::new(BlockTsCache::default())),
         }
     }
 
@@ -679,6 +701,50 @@ impl Handler for ChainsHandler {
 impl ChainsHandler {
     fn entry_with_unix_timestamp(entry: Entry, timestamp_secs: u64) -> Entry {
         entry.with_modified(SystemTime::UNIX_EPOCH + Duration::from_secs(timestamp_secs))
+    }
+
+    /// Head-block timestamp for `chain`, cached for [`HEAD_TS_TTL`].
+    /// Returns `None` on RPC failure — listings must not fail (or slow
+    /// down under readdir bursts) because a metadata fetch did.
+    async fn head_timestamp_cached(&self, chain: &str, client: &ChainClient) -> Option<u64> {
+        if let Some((fetched_at, ts)) = self.block_ts_cache.lock().head.get(chain)
+            && fetched_at.elapsed() < HEAD_TS_TTL
+        {
+            return Some(*ts);
+        }
+        let ts = match client.block_latest().await {
+            Ok(Some(block)) => block.header.timestamp,
+            _ => return None,
+        };
+        self.block_ts_cache
+            .lock()
+            .head
+            .insert(chain.to_string(), (Instant::now(), ts));
+        Some(ts)
+    }
+
+    /// Timestamp of mined block `number` on `chain`. Cached without a
+    /// TTL: block timestamps are immutable once mined.
+    async fn block_timestamp_cached(
+        &self,
+        chain: &str,
+        number: u64,
+        client: &ChainClient,
+    ) -> Option<u64> {
+        let key = (chain.to_string(), number);
+        if let Some(ts) = self.block_ts_cache.lock().by_number.get(&key) {
+            return Some(*ts);
+        }
+        let ts = match client.block_by_number(number).await {
+            Ok(Some(block)) => block.header.timestamp,
+            _ => return None,
+        };
+        let mut cache = self.block_ts_cache.lock();
+        if cache.by_number.len() >= BLOCK_TS_CACHE_CAP {
+            cache.by_number.clear();
+        }
+        cache.by_number.insert(key, ts);
+        Some(ts)
     }
 
     fn json_hex_u64(value: &serde_json::Value, key: &str) -> Option<u64> {
@@ -1329,26 +1395,24 @@ impl ChainsHandler {
                     Entry::file("timestamp"),
                     Entry::file("full.json"),
                 ];
-                match client.block_latest().await {
-                    Ok(Some(block)) => Ok(entries
+                match self.head_timestamp_cached(chain, &client).await {
+                    Some(ts) => Ok(entries
                         .into_iter()
-                        .map(|entry| Self::entry_with_unix_timestamp(entry, block.header.timestamp))
+                        .map(|entry| Self::entry_with_unix_timestamp(entry, ts))
                         .collect()),
-                    _ => Ok(entries),
+                    None => Ok(entries),
                 }
             }
             3 if segs[1] == "blocks" => {
                 // /chains/<chain>/blocks/<number>
                 let entries: Vec<Entry> = BLOCK_FILES.iter().map(|n| Entry::file(n)).collect();
                 Ok(match segs[2].parse::<u64>() {
-                    Ok(n) => match client.block_by_number(n).await {
-                        Ok(Some(block)) => entries
+                    Ok(n) => match self.block_timestamp_cached(chain, n, &client).await {
+                        Some(ts) => entries
                             .into_iter()
-                            .map(|entry| {
-                                Self::entry_with_unix_timestamp(entry, block.header.timestamp)
-                            })
+                            .map(|entry| Self::entry_with_unix_timestamp(entry, ts))
                             .collect(),
-                        _ => entries,
+                        None => entries,
                     },
                     Err(_) => entries,
                 })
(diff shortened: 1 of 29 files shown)


crates/bloom-vfs/src/handlers/chains.rs
```diff
@@ -661,29 +661,142 @@ impl Handler for ChainsHandler {
...
+
+    async fn receipt_json(
+        client: &ChainClient,
+        hash: alloy::primitives::B256,
+    ) -> Result<serde_json::Value, HandlerError> {
+        match client.receipt(hash).await {
+            Ok(Some(receipt)) => serde_json::to_value(receipt).map_err(err_be),
+            Ok(None) => Err(HandlerError::not_found(format!("receipt {hash:#x}"))),
+            Err(typed_err) => client
+                .raw_receipt(hash)
```
GavinPacini: I think rather than a `raw_receipt` fallback, we should try to understand all types of transactions that might come in, and this would kind of hide that from us in future.

In the case of Base's (and any op-stack chain's) 0x7E type, we should probably add `op-alloy` to our deps and Op-stack providers so that they are parsed correctly: https://crates.io/crates/op-alloy

GavinPacini (changes requested):

violet-agent:
## Consolidated review: PR #97

Reviewed exact range `969b0c0a76b7524a1f071bfd661fc7c84bd0697c...6855154a138712ce6a70c36468ca1879c8d32519`, including full context around the 27 changed files. I also ran an independent read-only `claude -p` review and manually reconciled its findings rather than forwarding them unverified.

### High

1. **Existing installations do not opt into the new OP-stack decoder, so the advertised Base fix remains broken after upgrade.**  
   `ChainSpec::op_stack` defaults to `false` when absent (`crates/bloom-proto/src/chain.rs:114-119`), while Base/Optimism are marked only in `Config::local_default()` (`crates/bloom-proto/src/config.rs:376-381`, `404-409`). `Config::load()` simply deserializes an existing file (`config.rs:502-506`) and performs no migration. Therefore any pre-PR `config.toml` keeps `op_stack = false` for Base/Optimism. I reproduced this against the PR's documented Base `0x7e` transaction: with an existing-style config lacking `op_stack`, `vfs cat .../full.json` fails with `unknown variant 0x7e`; with `op_stack = true`, the same read succeeds and preserves `sourceHash`.  
   **Fix:** preserve whether the field was absent during deserialization (for example `Option<bool>` in a migration/input type), default known chain IDs 10/8453 to OP-stack when absent, and persist the upgraded config. Add an upgrade test using a pre-PR config fixture plus the OP receipt mock.

2. **Chain `lookup()`/`stat` now performs uncached RPC fan-out.**  
   The timestamp helpers call `block_latest`, `receipt_block_number`, and `block_by_number` from `lookup_inner()` (`crates/bloom-vfs/src/handlers/chains.rs:708-741`, with call sites around `755-766` and `848-857`). `Vfs::lookup()` does not use the read `PathCache`, and the mount invokes lookup repeatedly during component resolution/getattr. A normal read of `tx/<hash>/receipt.json` can therefore fetch the receipt/block for the hash directory, fetch them again for the leaf, then fetch the receipt again for the read. `ls -la`, `find`, and kernel getattr traffic can generate RPC load without reading content. This is especially risky against the public/rate-limited endpoints in the default config.  
   **Fix:** keep lookup local, or add a dedicated cached metadata path/snapshot. Do not make each path component independently refetch immutable receipt/block data.

### Medium

3. **Unknown mtimes became “now on every stat,” breaking change detection for untouched handlers.**  
   `entry_to_attrs()` uses `e.modified.unwrap_or_else(SystemTime::now)` (`crates/bloom-mount/src/adapter.rs:290`) instead of retaining the previous stable epoch fallback. Many handlers still leave `modified = None`, so static paths under docs/status/tools/prices/etc. appear freshly modified on every stat. That makes mtimes unusable for `rsync`, `find -newer`, indexers, and incremental backups.  
   **Fix:** when `modified` is unknown, retain a stable sentinel (the prior epoch behavior) rather than fabricating a changing timestamp. Keep `change` as the dynamic cache-invalidation mechanism.

4. **Pre-Byzantium Ethereum receipt status regresses from typed handling to a hard error.**  
   `json_status()` accepts only a JSON `status` field (`crates/bloom-vfs/src/handlers/chains.rs:673-679`). Alloy serializes `Eip658Value::PostState` as `root`, while the old typed `receipt.status()` path coerced `PostState` consistently. As a result, historical `/status` and `/error.json` reads can now fail with “missing decodable status.”  
   **Fix:** expose status through the typed receipt API (including the OP receipt type), or explicitly handle `root` with the same documented legacy semantics. Add a receipt fixture containing `root` and no `status`.

5. **Fresh-home request paths now fail as EIO before the first list/create operation.**  
   `requests::lookup()` calls `entry_for_fs_path()` for the synthetic state directories (`crates/bloom-vfs/src/handlers/requests.rs:1138-1162`), but `ensure_layout()` runs only from create/list. The metadata helper propagates `NotFound` as `HandlerError::Io`. I reproduced this on a fresh home: `bloom ... vfs stat /requests/pending` exits with `vfs lookup: io: No such file or directory`.  
   **Fix:** ensure the request layout before lookup, or translate missing metadata to the correct synthetic entry/`NotFound` behavior. Add a direct-lookup test on a fresh home (without calling `list()` first).

6. **Metadata enrichment can hide or block recovery of corrupted financial records.**  
   New per-record metadata loading uses `.filter_map(... .ok())` for Polymarket fund sessions/drafts/receipts (`crates/bloom-vfs/src/handlers/polymarket.rs:3575-3579`, `3607-3621`) and wallet outbox entries (`crates/bloom-vfs/src/handlers/wallets.rs:2332-2348`), silently dropping records whose JSON cannot be parsed. Hyperliquid instead propagates one bad `session.json` and fails the whole listing (`crates/bloom-vfs/src/handlers/hyperliquid.rs:2962-2992`, `3142-3153`). This also turns cheap directory listings into O(n) file reads/parses. A torn write should remain visible for diagnosis, not disappear because optional timestamp extraction failed.  
   **Fix:** always return the name-only `Entry`; enrich it with a timestamp on successful parse/metadata lookup, and log a warning on fallback.

7. **OP-stack revert replay drops execution-relevant transaction fields.**  
   The OP branch of `trace_revert()` manually rebuilds a request with only `from`, input, gas limit, value, destination, and access list (`crates/bloom-evm/src/lib.rs:451-464`). It omits gas price / EIP-1559 fee fields (and other typed fields available on the transaction). Contracts can inspect `tx.gasprice`, so the replay can follow a different branch and return the wrong/no revert data. The L1 branch still uses `into_request()` and preserves these fields.  
   **Fix:** centralize a typed OP transaction-to-call conversion that preserves all execution-relevant standard fields; add a replay test whose revert depends on `GASPRICE`.

### Security / supply-chain result

I did not identify a new critical exploit, secret leak, unsafe-code issue, or suspicious dependency/typosquat in this diff. The main risks are upgrade correctness, RPC amplification/availability, and financial-record observability.

### Verification performed

- Current PR head/base rechecked immediately before posting; head remained `6855154a...`, and all GitHub checks are green.
- `git diff --check 969b0c0...HEAD` — pass.
- `cargo test -p bloom-evm --lib` — **62 passed**.
- `cargo test -p bloom-vfs handlers::chains::tests --lib` — **58 passed**.
- Focused timestamp tests across requests/outbox/DeFi/watch, mount attrs, and CLI `vfs stat` — all passed.
- Fresh-home direct request stat — reproduced EIO regression.
- Live Base `0x7e` read with migrated vs. legacy-style config — success vs. deterministic decode failure reproduced.
- Independent `claude -p` review (Claude Code 2.1.209) completed read-only; its actionable findings were manually validated and folded into this single comment.

GavinPacini:
Screenshot 2026-07-16 at 18 46 32

This is so sick, very nice sir! LGTM.

PR #101, opened 2026-07-10
Surface insufficient gas funds before approval

### LINKED ISSUE #63: [evm]: If the account has insufficent funds for gas we should surface that in the plan
Insufficient funds for gas (gas required exceeds allowance, insufficient funds): fail staging (or stage but mark the entry blocked) with an actionable message like "account 0xABC… has 0 ETH on Base; fund it to cover gas before approving," and surface that in plan.md / the challenge so the user sees it before running the passkey ceremony rather than after a silent no‑broadcast.

0xdewy:
## Summary
- compare the pinned native balance during EVM staging with transaction value plus buffered gas at the staged fee ceiling
- add a hard balance.native_funds policy denial with actionable funding and restaging guidance in plan.md and policy artifacts
- stop confirmation before Sealed Approval when the staged transaction cannot be funded
- preserve existing staging behavior, with a warning log, when the balance RPC is unavailable

## Stacking
- stacked on #50 (feat/local-petal-plugins) because that PR changes the same staging and fee-selection path
- base branch: feat/local-petal-plugins
- implements #63; retarget to master after #50 merges so GitHub can close the issue

## Verification
- cargo fmt --all -- --check
- cargo test -p bloom-tx
- cargo clippy -p bloom-tx --all-targets -- -D warnings
- cargo check --workspace
- git diff --check

## Checklist
- [x] Tests added for the hard-denial plan message and exact required-balance boundary
- [x] Architecture docs unchanged: this uses the existing staged policy-check contract
- [x] Sealed Approval invariants preserved: insufficiently funded transactions are denied before ceremony/signing
- [x] Agent documentation unchanged: no command or VFS path semantics changed

Closes #63

### DIFF
--- a/crates/bloom-tx/src/tx_engine.rs
+++ b/crates/bloom-tx/src/tx_engine.rs
@@ -1217,6 +1217,28 @@ impl TxEngine {
         } else {
             (Some(max_fee.to_string()), Some(prio.to_string()), None)
         };
+        let funding_check = match session.balance(from).await {
+            Ok(available) => insufficient_native_funds_check(
+                &bloom_proto::checksum_address(&from),
+                spec.display_name.as_deref().unwrap_or(&spec.name),
+                &spec.native_symbol,
+                spec.native_decimals,
+                available,
+                value_wei,
+                gas_limit,
+                if spec.legacy_tx { gas_price } else { max_fee },
+            ),
+            Err(e) => {
+                tracing::warn!(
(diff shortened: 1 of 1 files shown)


GavinPacini:
LGTM, just tested it, nice one! @josh-richardson you happy if we merge this? It targets `feat/local-petal-plugins` FYI.

PR #107, opened 2026-07-13
Remove native Polymarket integration

josh-richardson:
## Summary

- Add a typed, pinned list of pre-installed Petals, with Polymarket as the default.
- Provision configured pre-installed Petals during `bloom init` and before `bloom serve`.
- Delete the duplicate native `bloom-polymarket` crate, `bloom polymarket` CLI, `/polymarket` VFS handler, daemon mount/configuration, daemon-owned venue state, and first-party Polymarket attestation path.
- Keep the wallet `[polymarket]` policy schema consumed by the external Petal.

## Why

The external Polymarket Petal now has parity through Bloom's generic HTTP, chain-read, storage, outbox, signing, and Sealed Approval host contracts. Retaining the native implementation would create two venue owners and two authority paths. The replacement must also be present by default, rather than requiring an undocumented manual install.

This branch was reconstructed directly on current `master`. Work already squash-merged through the Petal-platform PR is no longer shown as new, so the changeset now reflects the real delta.

## Provisioning and migration

- Fresh defaults use `[petals] preinstalled = ["polymarket"]`.
- `bloom init` and `bloom serve` install the trusted repository at immutable commit `1ffb267a1e1d4acd137c184806c20cc98d20a3f4`.
- Existing matching provenance is retained; conflicting ownership, source, commit, or package provenance fails closed without overwrite.
- `[petals] preinstalled = []` is a persistent opt-out.
- Existing `[polymarket] enabled = false` configs retain that opt-out until an explicit new `petals.preinstalled` list is provided.
- Polymarket is exposed at `/petals/polymarket/`; there is no native `/polymarket/` subtree or `bloom polymarket` CLI.

## Checks

- `git diff --check`
- `cargo test --workspace --all-targets --no-fail-fast`
- `cargo test --workspace --doc --no-fail-fast`
- `cargo build --release -p bloom --all-features --locked`
- ignored live source-contract test against the exact pinned Polymarket commit, including clone, build, install, provenance, and route dispatch

## Checklist

- [x] Default, explicit opt-out, legacy opt-out migration, validation, provenance mismatch, and idempotency coverage
- [x] Native CLI and VFS absence covered
- [x] Wallet Polymarket policy retained
- [x] Active user, agent, architecture, testing, and CI documentation updated to `/petals/polymarket/`
- [x] Generic Petal Sealed Approval and host-signing invariants retained; venue-specific authority removed


GavinPacini:
@josh-richardson Could you resolve conflicts so we can see the real changeset delta? Also, does this automatically install the Polymarket petal? If not, I think we should handle that as part of / before merging this PR, i.e. the ability to have a defined list of "pre-installed" Petals.

PR #109, opened 2026-07-15
feat: tag-driven releases + daemon update checker

### LINKED ISSUE #108: [ci/daemon]: release workflow with versioned tags & daemon update checks
AC: 
- A release workflow on this repo that publishes on tag (reference: https://github.com/vibefi/client/blob/master/.github/workflows/release.yml). 
- The daemon checks its own version against the latest released version on GitHub, prompting the user / agent to update when required.
- Versions for the `bloom` binary need to be derived from / matching the version in the GitHub release tag (e.g. done via a script here: https://github.com/vibefi/client/blob/master/.github/workflows/release.yml#L125 - maybe we should do it manually and ensure it matches in CI? Open to suggestions here). 
- Surface the currently installed bloom version and latest GitHub version in the VFS somewhere.

0xdewy:
Resolves #108.

## Summary

This PR adds versioned Bloom releases and the daemon update checker. Stable releases now require a reviewed version-bump PR followed by a human-pushed immutable vX.Y.Z tag.

CI can prepare the version bump, but it does not write to the default branch, create stable tags, or publish stable releases.

## Reviewed release flow

- Run Propose Release with the next X.Y.Z version.
- CI opens a release-proposal/vX.Y.Z pull request containing only Cargo.toml and the required Cargo.lock workspace metadata update.
- CI rejects non-increasing versions, duplicate proposals, and unrelated dependency resolution changes.
- Human review and normal pull-request checks are required before merge.
- A maintainer pushes the matching vX.Y.Z tag from the reviewed merge commit.
- The release workflow verifies the tag and workspace version, runs cargo test --workspace --locked against the immutable SHA, builds all platform artifacts from that SHA, and publishes only after those checks pass.

Before using it, configure the RELEASE_PR_TOKEN repository secret; a non-GITHUB_TOKEN credential is needed so the generated PR receives normal CI events.
[GitHub token documentation](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token)

## Review fixes

- Missing update timestamps render empty instead of the Unix epoch; summary.json now mirrors the normalized leaf fields.
- SemVer handling uses the standard semver crate, preserves prerelease ordering, ignores build metadata for precedence, and rejects malformed versions. Multiple prerelease identifiers such as 1.2.3-rc.1.2 are valid SemVer and are intentionally accepted.
- Failed refreshes no longer expose a contradictory behind_by value.
- CLI version hints normalize a leading v and tests no longer depend on a hardcoded package version.
- Automatic polling has a documented BLOOM_DISABLE_UPDATE_CHECK opt-out, and requests no longer disclose the installed version in User-Agent.
- Release actions are pinned to reviewed SHAs; permissions are job-scoped; non-publishing checkouts do not persist credentials.
- Release retry latest selection is monotonic and handles the legacy latest release alias without regressing stable downloads.

## Verification

- cargo test --workspace --locked
- cargo test -p bloom-update — 33 passed
- cargo check --workspace --locked
- cargo fmt --all -- --check
- Workflow YAML and embedded shell syntax checks

Conditional ETag/backoff polling remains an optional future optimization; the requested privacy control is provided by the opt-out and generic User-Agent.

## Checklist

- [x] Architecture docs updated if contracts or behavior changed — N/A; release/update-status behavior is documented in docs/operations/release-process.md.
- [x] Sealed Approval invariants respected — N/A; no signing or authorization path changed.
- [x] Agent Documentation updated — N/A; no guidance or Petal contract changed.


.github/workflows/release.yml
```diff
@@ -1,73 +1,238 @@
...
+        with:
+          ref: ${{ needs.prepare_release.outputs.release_branch_sha }}
+          fetch-depth: 0
 
       - name: Install Rust toolchain
         run: rustup show
 
-      - name: Build full-feature binary
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
```
josh-richardson: **High — release supply-chain risk.** This mutable third-party action runs before the release build while the workflow inherits `contents: write`, and checkout has persisted a write-capable credential. If the `v2` tag or action account is compromised, it can alter the binaries later published and push repository refs. Please pin actions—especially this one—to reviewed full commit SHAs, default the workflow to read-only permissions, grant write only to the jobs that push or publish, and set `persist-credentials: false` in non-pushing jobs.

.github/workflows/release.yml
```diff
@@ -1,73 +1,238 @@
...
+            # already point at a commit whose workspace version matches it.
+            if [[ "$CURRENT_VERSION" != "$VERSION" ]]; then
+              echo "::error::tag v$VERSION points at workspace version $CURRENT_VERSION; bump and commit before tagging"
+              exit 1
+            fi
+          else
+            # A manual CI-only release starts from the default branch and
+            # creates the version commit before creating the tag.
+            sed -i -E 's/^version = "[0-9]+\.[0-9]+\.[0-9]+"/version = "'"$VERSION"'"/' Cargo.toml
+            cargo update --workspace
```
josh-richardson: **High — manual releases can ship unreviewed dependency upgrades.** `cargo update --workspace` re-resolves all compatible registry/git dependencies, while the following `git diff --stat` only prints filenames and cannot enforce the stated invariant. The generated lockfile is then committed and the release path runs only `cargo build`. This PR already demonstrates broad lockfile churn, including Alloy 2.0.5 → 2.2.0. Please preserve existing transitive selections, fail if non-workspace package tuples change, and run the normal tests against the immutable release SHA; ideally make the version bump a reviewed PR and keep release publication tag-only.

.github/workflows/release.yml
```diff
@@ -81,42 +246,66 @@ jobs:
...
-              --title "Latest master build" \
-              --notes-file "$notes_file" \
+          if gh release view "$TAG" >/dev/null 2>&1; then
+            # Retry path: release already exists. Update metadata and
+            # re-upload assets (clobbering any previous ones).
+            gh release edit "$TAG" \
+              --verify-tag \
+              --title "bloom v$VERSION" \
+              --notes-file "$NOTES" \
               --latest
```
josh-richardson: **Medium — retrying an older release redirects the stable latest alias backward.** This retry path unconditionally passes `--latest`, so repairing assets for `v0.1.9` after publishing `v0.2.0` makes `/releases/latest/download/...` serve the older version. New lower-version tags have the same problem because there is no monotonicity check. Please only mark this release latest when its stable SemVer is greater than or equal to the current latest release.

.github/workflows/release.yml
```diff
@@ -1,73 +1,238 @@
...
+            echo "::error::release tag must match vX.Y.Z, got: $TAG"
+            exit 1
+          fi
+          git fetch --tags --force
+          if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+            TAG_EXISTS=true
+            REF="refs/tags/$TAG"
+          elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG_EXISTS=false
+            REF="refs/heads/${{ github.event.repository.default_branch }}"
```
josh-richardson: **Medium — this is a moving release input.** `prepare` records the default branch ref, and `prepare_release` resolves it in a later job. A merge between the jobs can therefore release a different commit than the workflow-dispatch revision selected by the operator. Please resolve and output an immutable commit SHA during `prepare`, then use that SHA for checkout and provenance checks.

crates/bloom/tests/cli.rs
```diff
@@ -568,6 +568,96 @@ fn vfs_default_missing_socket_falls_back_in_process() {
...
+    bloom_cmd(home.path())
+        .args(["vfs", "cat", "/status/update/latest"])
+        .assert()
+        .success()
+        .stdout(predicate::eq("0.2.0\n"));
+    bloom_cmd(home.path())
+        .args(["vfs", "cat", "/status/update/available"])
+        .assert()
+        .success()
+        .stdout(predicate::eq(if env!("CARGO_PKG_VERSION") == "0.1.0" {
```
josh-richardson: **Medium — this assertion breaks on the documented `0.1.1` version bump.** With `latest = 0.2.0`, installed `0.1.1` is correctly out of date, but this test expects `up_to_date` for every installed version other than exactly `0.1.0`. That blocks the recommended version-bump PR, while the CI-only release path misses it because it does not run tests. Please compare versions semantically or construct the fixture relative to `CARGO_PKG_VERSION`.

crates/bloom-update/src/snapshot.rs
```diff
@@ -0,0 +1,263 @@
...
+            std::cmp::Ordering::Less => UpdateAvailable::OutOfDate,
+            _ => UpdateAvailable::UpToDate,
+        }
+    }
+
+    /// "Behind by" count, or `None` if we don't have enough
+    /// information. See [`crate::cache::behind_by`] for the
+    /// counting rule.
+    pub fn behind_by(&self) -> Option<u64> {
+        crate::cache::behind_by(self)
```
josh-richardson, line 174: **Medium — failed refreshes expose contradictory verdicts.** `available()` returns `Unknown` whenever status is not OK, but this method computes a positive distance from the cached `latest` without applying the same status gate. After a network failure, consumers can therefore see `available=unknown` and `behind_by=100` simultaneously. Please return `None` here unless `status_kind() == Ok`, or consistently retain both cached-success verdicts, and add a failed-refresh-with-cache test.

crates/bloom-update/src/semver.rs
```diff
@@ -0,0 +1,120 @@
...
+    let s = s.strip_prefix('v').unwrap_or(s);
+    let mut parts = s.split('.');
+    let major = parts.next()?.parse().ok()?;
+    let minor = parts.next()?.parse().ok()?;
+    // The third component may carry a pre-release suffix
+    // ("0.2.0-rc.1" → "0-rc.1"). We split on '-' to strip it; if
+    // a `-` is present we tolerate an additional dot-separated tail
+    // ("0.2.0-rc.1" has 4 dot-separated parts: "0", "2", "0-rc", "1").
+    let third = parts.next()?;
+    let (patch_str, had_prerelease) = match third.split_once('-') {
```
josh-richardson: **Medium — stripping prerelease data breaks SemVer ordering.** Installed `0.2.0-rc.1` and latest stable `0.2.0` both become `(0, 2, 0)`, so the checker reports up-to-date and exits 0 even though the stable release is newer. The parser also accepts malformed trailing components such as `1.2.3-rc.1.2` because it does not exhaust the iterator. Please use the standard `semver` crate, preserve prerelease ordering, and add these cases as tests.

crates/bloom-vfs/src/handlers/status.rs
```diff
@@ -748,6 +830,58 @@ impl StatusHandler {
...
+                        UpdateAvailable::UpToDate => b"up_to_date\n".to_vec(),
+                        UpdateAvailable::Unknown => b"unknown\n".to_vec(),
+                    }),
+                    "behind_by" => Ok(format!("{}\n", snap.behind_by.unwrap_or(0)).into_bytes()),
+                    "checked_at" => {
+                        let secs = snap
+                            .checked_at
+                            .and_then(|t| t.duration_since(UNIX_EPOCH).ok())
+                            .map(|d| d.as_secs())
+                            .unwrap_or(0);
```
josh-richardson: **Low — an unknown timestamp is rendered as a real check in 1970.** Mapping `checked_at: None` to zero produces `1970-01-01T00:00:00Z`, contradicting the field definition and the daemon comment that unknown values render empty. `summary.json` also serializes the raw nullable/SystemTime DTO at line 880, so its representation disagrees with the text leaves. Please render an empty line for `None` and emit one normalized JSON shape with RFC3339-or-null semantics.

crates/bloom/src/main.rs
```diff
@@ -1327,6 +1350,23 @@ async fn run(cli: Cli) -> Result<()> {
...
+                let available = match snap.available() {
+                    bloom_update::UpdateAvailable::OutOfDate => "out_of_date",
+                    bloom_update::UpdateAvailable::UpToDate => "up_to_date",
+                    bloom_update::UpdateAvailable::Unknown => "unknown",
+                };
+                println!("latest_release: {}", latest);
+                println!("update_available: {}", available);
+                if matches!(snap.available(), bloom_update::UpdateAvailable::OutOfDate) {
+                    eprintln!(
+                        "hint: bloom v{} is available (you have v{}); see /status/update",
```
josh-richardson, line 1365: **Low — the hint doubles the version prefix.** GitHub returns and stores the required release tag verbatim as `vX.Y.Z`, then this format string prepends another `v`, producing `bloom vv0.2.0`. Please normalize the tag at ingestion or strip an optional `v` for display, and add an assertion using a realistic `v0.2.0` cached tag.

crates/bloom-update/src/checker.rs
```diff
@@ -0,0 +1,447 @@
...
+            let mut ticker = tokio::time::interval(REFRESH_INTERVAL);
+            ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+            // Consume the immediate first tick (interval fires
+            // immediately on first poll, which we don't want because
+            // we just did a refresh above).
+            ticker.tick().await;
+            loop {
+                tokio::select! {
+                    _ = &mut rx => {
+                        debug!("bloom_update.background_shutdown");
```
josh-richardson, line 209: **Low / privacy — automatic polling has no operator opt-out.** Starting `bloom serve` triggers an immediate GitHub request and repeats every five minutes, exposing the host IP, usage cadence, and exact installed version via User-Agent to GitHub or intermediaries. Please add a documented config/environment disable switch and consider conditional requests with ETag/backoff or a substantially longer interval.

josh-richardson (changes requested):
Requesting changes for two high-severity release-integrity issues: the mutable third-party action executing with workflow-wide write access, and CI-only releases re-resolving and shipping unreviewed dependency upgrades without tests. The inline review also identifies release-order/provenance problems, correctness and test regressions, inconsistent update-status semantics, SemVer handling errors, and lower-severity UX/privacy concerns. Please address each inline thread before merging.

crates/bloom-update/src/semver.rs
```diff
@@ -0,0 +1,120 @@
...
+    let patch = patch_str.parse().ok()?;
+    // A 4th dot-separated component is only OK if it came from a
+    // pre-release suffix (e.g. the "1" in "rc.1"). A bare
+    // "1.2.3.4" is rejected.
+    if let Some(extra) = parts.next()
+        && (!had_prerelease || !extra.chars().all(|c| c.is_ascii_digit()))
+    {
+        return None;
+    }
+    Some((major, minor, patch))
```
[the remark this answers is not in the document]
0xdewy: Implemented in commit dc933bb. parse_semver now uses the standard semver crate and preserves prerelease ordering. One correction to the original concern: 1.2.3-rc.1.2 is valid SemVer because prerelease identifiers may contain multiple dot-separated identifiers, so the test intentionally preserves that case; malformed versions are rejected.

crates/bloom-update/src/checker.rs
0xdewy: Implemented in commit dc933bb. Added BLOOM_DISABLE_UPDATE_CHECK and removed the installed version from User-Agent. ETag/backoff remains an optional future optimization and is called out in the PR description; the requested operator opt-out and version privacy fix are addressed.

GavinPacini:
I found two remaining release-workflow issues that should be addressed before merging.

**P1 — Release tags are not proven to reference reviewed code**

The `prepare` job verifies that the tag exists and that its `Cargo.toml` version matches the tag, but it does not verify that the tagged commit is reachable from the protected default branch.

Consequently, a `vX.Y.Z` tag pointing to an unmerged feature commit with a matching version could pass validation and be published, bypassing the stated guarantee that stable releases originate from reviewed code.

A repository ruleset restricting creation, updates, and deletion of `v*` tags has now been configured. That limits who can create release inputs. The workflow should complement it by rejecting accidental off-branch tag targets:

```bash
DEFAULT_BRANCH="${{ github.event.repository.default_branch }}"
git fetch --force origin \
  "refs/heads/$DEFAULT_BRANCH:refs/remotes/origin/$DEFAULT_BRANCH"

if ! git merge-base --is-ancestor \
  "$RELEASE_SHA" \
  "refs/remotes/origin/$DEFAULT_BRANCH"
then
  echo "::error::tag $TAG does not point to a commit on $DEFAULT_BRANCH"
  exit 1
fi
```

An ancestry check is preferable to requiring equality with the current branch tip, because unrelated commits may legitimately merge after the release tag is created.

**P2 — A failed latest-release lookup can move the latest alias backward**

The publish step currently uses:

```bash
CURRENT_LATEST="$(gh api "repos/${{ github.repository }}/releases/latest" --jq '.tag_name' 2>/dev/null || true)"
```

This converts every API failure—including authentication problems, rate limiting, GitHub outages, and network errors—into an empty result. The following condition treats an empty result as “no latest release” and sets `MARK_LATEST=true`.

Therefore, retrying an older release during a transient API failure could mark that older release as latest and redirect `/releases/latest/download/...` backward.

Please fail closed when the lookup fails. Only a positively identified “no eligible release exists” response should be treated as empty; operational API failures should abort publication. A test covering “older retry + latest lookup failure” would help preserve the monotonic latest guarantee.

josh-richardson (approved):

PR #116, opened 2026-07-16
Enable per-chain broadcast by default

violet-agent:
## Summary

- default each generated chain and any omitted `allow_broadcast` field to `true`
- use `allow_broadcast = false` on a specific chain, including Ethereum mainnet, to disable broadcast there
- remove the daemon-wide `block_mainnet_broadcast` setting and its `/status/policies/` projection
- enable the Polymarket VFS by default with its public Polygon endpoints
- enable Polymarket wallet trading policy by default while retaining explicit `enabled = false` opt-out and policy limits
- rewrite `/polymarket/README.md` around the live VFS read surfaces, writable sinks, status and challenge files, and current execution constraints, with no CLI commands
- keep signing, wallet policy, review, confirmation, and Sealed Approval as independent transaction safeguards
- make embedded VFS guidance describe only the live mounted product surface, without setup or host configuration

## Change type

- [x] Feature
- [x] Refactor
- [x] Docs
- [x] Tests

## Testing

- [x] `cargo fmt --check`
- [x] `cargo test --workspace`
- [x] `cargo check --workspace`
- [x] focused Polymarket default-policy, default-config, VFS README, and CLI/VFS parity tests
- [x] `bash -n` on the modified shell and Docker test scripts from the broadcast-control change
- [x] `git diff --check`

## Safety / compatibility

- Broadcast routing has one source of truth: the selected chain's `allow_broadcast` value.
- `allow_broadcast = false` is enforced by confirm, replacement, cancellation, and public-RPC broadcast paths.
- Polymarket trading defaults on, but orders still pass market invariants, wallet policy, price/exposure, review, signing, venue-compliance, receipt, and audit gates.
- Value-moving actions still pass the applicable signing, policy, review, confirmation, and Sealed Approval gates.
- No secrets, generated artifacts, or unrelated dependencies are included.


crates/bloom-vfs/src/docs/README.md
```diff
@@ -315,8 +315,8 @@ echo '0x000000000000000000000000000000000000beef' \
 cat /bloom/addressbook/alice
 ```
 
-Mainnet broadcasts are **disabled by default**. Configure via
-`~/.bloom/config.toml` (`block_mainnet_broadcast = false` is required
-to allow live broadcasts).
+Per-chain `allow_broadcast` defaults to `true` and
```
GavinPacini: I think we need to stop surfacing anything that's controlled outside of the VFS inside the VFS.

crates/bloom-vfs/src/docs/agent-guidance.md
```diff
@@ -31,6 +31,9 @@ bloom gatekeeps every value-moving action through capabilities:
   reside in daemon RAM for a bounded window and auto-lock on expiry.
   Hyperliquid already uses an ephemeral agent key that does not need the
   owner key after session creation.
+- **Broadcast config has two gates.** Per-chain `allow_broadcast` defaults to
```
GavinPacini: Same here, this is projected inside the VFS but irrelevant to an agent without access outside the FS.

GavinPacini (changes requested):

crates/bloom-vfs/src/docs/README.md
violet-agent: Addressed in 17a3aca. Embedded VFS docs now stay within the mounted product surface: setup, host configuration, CLI fallbacks, and other outside-VFS guidance were removed.

crates/bloom-vfs/src/docs/agent-guidance.md
violet-agent: Addressed in 17a3aca. Agent guidance now describes only mounted VFS paths and live challenge/capability flows; host config, CLI-only recovery, future behavior, and outside-signed payload guidance were removed.

0xdewy:
**P1 — Preserve the legacy mainnet broadcast lock**\n\nWhen an existing config has `block_mainnet_broadcast = true` and a chain omits `allow_broadcast`, the removed global field is silently ignored and `allow_broadcast` defaults to `true`. After upgrade, confirmations can therefore broadcast on mainnet even though the prior config explicitly disabled mainnet broadcasts. Please migrate the legacy setting into per-chain `allow_broadcast = false` values, or reject configs containing the legacy field, rather than enabling broadcasts unexpectedly.

0xdewy:
**P1 — Preserve the legacy default-deny Polymarket policy**\n\n`PolymarketPolicy.enabled` changed from default `false` to default `true`. Existing wallet policies that omit `[polymarket]` therefore now parse as trading-enabled with no order cap, daily cap, price cap, or allowlist. An open BUY passes the policy checks, whereas the same policy previously denied it. For passkey wallets, the signed `policy.toml` content is unchanged, so signature verification does not detect this semantic change after upgrade. Please preserve the legacy default-deny behavior for existing policies or add an explicit versioned migration/opt-in; do not silently enable unrestricted Polymarket trading for existing wallets.

crates/bloom-proto/src/polymarket_policy.rs
```diff
@@ -40,9 +41,8 @@ fn default_true() -> bool {
 /// `[polymarket]` section of a wallet's policy file.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PolymarketPolicy {
-    /// Master gate. Defaults to **false**: a wallet cannot place Polymarket
-    /// orders until its owner opts in via the policy file.
-    #[serde(default)]
+    /// Master gate. Defaults to **true**; wallet policy can disable trading.
+    #[serde(default = "default_true")]
```
josh-richardson, line 45: **[P1] Preserve the meaning of existing signed policies.** Changing the serde default here from `false` to `true` means a previously signed `policy.toml` that omitted `[polymarket]` remains cryptographically valid but is now interpreted as authorizing trading. The signature verification covers the unchanged bytes, not the deserialization semantics, so this expands wallet authority without owner review or re-signing. Please preserve the legacy interpretation for existing policy bytes, or introduce a versioned migration that requires the owner to sign the expanded policy.

crates/bloom-proto/src/config.rs
```diff
@@ -44,9 +44,9 @@ pub struct Config {
     pub etherscan: Option<EtherscanConfig>,
     #[serde(default)]
     pub enso: Option<EnsoConfig>,
-    /// Polymarket read surface. Presence of `[polymarket]` opts the
-    /// `polymarket/` VFS subtree in (like `[enso]` opts in `defi/`).
-    #[serde(default)]
+    /// Polymarket read and trading surface. Enabled with public Polygon
+    /// endpoints by default; a `[polymarket]` table can override them.
+    #[serde(default = "default_polymarket_config")]
```
josh-richardson, line 49: **[P2] Keep a persistent Polymarket opt-out.** With an absent field defaulting to `Some(...)`, TOML cannot express `None` because it has no null value. Omitting `[polymarket]` enables it, while serializing a programmatic `None` omits the field and reloading turns it back into `Some`, so even the `else`/disabled path in the daemon cannot survive a save/load cycle. Please add an explicit `enabled = false` setting or another serialization-safe opt-out.

josh-richardson (commented):
Two compatibility issues should be addressed before merging. The existing Tempo documentation inconsistency already has an unresolved inline thread, so I have not duplicated it.

crates/bloom-proto/src/polymarket_policy.rs
GavinPacini: If you agree, I think we just move forward with this defaulting to on, even for old configs. Anyway this will be moved to a Petal soon.

GavinPacini:
Merged to get this in for this week's user testing, will follow up on open comments if required.

@0xdewy 
> **P1 — Preserve the legacy mainnet broadcast lock**\n\nWhen an existing config has `block_mainnet_broadcast = true` and a chain omits `allow_broadcast`, the removed global field is silently ignored and `allow_broadcast` defaults to `true`. After upgrade, confirmations can therefore broadcast on mainnet even though the prior config explicitly disabled mainnet broadcasts. Please migrate the legacy setting into per-chain `allow_broadcast = false` values, or reject configs containing the legacy field, rather than enabling broadcasts unexpectedly.

I think we just move forward with this, we need to default to allowing broadcast now, time to test more things. :)

> **P1 — Preserve the legacy default-deny Polymarket policy**\n\n`PolymarketPolicy.enabled` changed from default `false` to default `true`. Existing wallet policies that omit `[polymarket]` therefore now parse as trading-enabled with no order cap, daily cap, price cap, or allowlist. An open BUY passes the policy checks, whereas the same policy previously denied it. For passkey wallets, the signed `policy.toml` content is unchanged, so signature verification does not detect this semantic change after upgrade. Please preserve the legacy default-deny behavior for existing policies or add an explicit versioned migration/opt-in; do not silently enable unrestricted Polymarket trading for existing wallets.

Same here, no real users yet so let's keep moving forward.

crates/bloom-proto/src/config.rs
GavinPacini: Opening a new PR to tackle this.

violet-agent: Fixed in follow-on PR #120 with an explicit default-true runtime `enabled` field, persistent `enabled = false` TOML round-trip coverage, and daemon mount-gating tests.

PR #121, opened 2026-07-21
feat(auth): Redesign passkey wallet setup

GavinPacini:
## What changed

- redesign the passkey wallet registration and policy form in Bloom's visual system
- add a responsive layout, accessible controls, website-matching lockup, and theme-aware favicons
- self-host Instrument Serif, Inter Tight, and JetBrains Mono so the local ceremony makes no outbound requests
- remove the separate recovery-key webpage while retaining the one-time terminal acknowledgment
- serve embedded font and favicon assets from the short-lived Axum registration server

## Why

The wallet-creation ceremony used a minimal monospace form that did not match Bloom's product identity and was difficult to scan on smaller screens. The redesign improves clarity around wallet policy choices without changing WebAuthn or PRF behavior. Self-hosting the fonts preserves the local-only ceremony trust boundary.

## Review follow-ups

- remove external Google Fonts requests and pin a no-referrer policy
- improve helper-text contrast and switch labeling
- prevent stale localhost asset caching and include the fonts' OFL notice
- make Enter submit the registration form correctly

Four Claude review rounds were completed. The final round returned no actionable findings.

## Validation

- `cargo fmt --check`
- `git diff --check`
- registration-page JavaScript syntax check with `node --check`
- verified the registration page contains no external URLs
- `cargo test -p bloom-keystore --lib -- --test-threads=1` — 98 passed, 3 real-authenticator tests ignored

## Checklist

- [x] Tests added or updated for behavior changes — N/A: existing passkey ceremony tests cover the server behavior; the HTML/JavaScript redesign was validated through syntax checks and manual preview
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A: no architecture contract changed
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — N/A: no agent-facing VFS or Petal behavior changed


crates/bloom-keystore/src/passkey.rs
```diff
@@ -1625,9 +1572,9 @@ impl super::Keystore {
         let final_policy = toml::from_str::<Policy>(&final_policy_toml)
             .map_err(|e| KeystoreError::Policy(format!("final policy parse: {e}")))?;
 
-        // Show the recovery key in a browser page; fall back to returning it
-        // for terminal display if the browser cannot be launched.
-        let recovery_key = recovery_key_field(name, signer).await;
+        // Return the recovery key for the existing one-time terminal prompt.
+        // It is never written to disk.
+        let recovery_key = Some(Zeroizing::new(hex::encode(signer.to_bytes())));
```
0xdewy: When passkey creation is initiated via `WalletsHandler::write_new`, this `WalletInfo` (including its one-time `recovery_key`) is discarded and the handler returns only `Ok(())`. That path no longer reaches the terminal recovery-key acknowledgment, so VFS-created wallets provide no recovery key; losing the passkey then permanently loses access. Please preserve recovery-key delivery and acknowledgment for VFS creation before returning.

crates/bloom-daemon/src/wallet_registration.html
```diff
@@ -1,79 +1,244 @@
...
+    body::after{content:"";position:fixed;width:440px;height:440px;right:-180px;top:-190px;border:1px solid rgba(138,42,58,.13);border-radius:50%;box-shadow:0 0 0 72px rgba(138,42,58,.025),0 0 0 144px rgba(82,111,81,.02);pointer-events:none}
+    button,input{font:inherit}
+    button,input,label{touch-action:manipulation}
+    button:focus-visible,input:focus-visible{outline:3px solid rgba(138,42,58,.2);outline-offset:2px}
+    .page{position:relative;z-index:1;width:min(1180px,100%);min-height:100vh;min-height:100svh;margin:0 auto;padding:28px clamp(20px,4vw,56px) clamp(32px,5vw,64px)}
+    .brand-lockup{display:inline-flex;min-width:0;align-items:center;gap:10px;color:var(--ink);font-family:var(--mono);font-size:12px;line-height:1.15;letter-spacing:0;text-decoration:none;text-transform:uppercase}
+    .brand-mark{width:30px;height:30px;flex:0 0 auto}
+    .brand-lockup strong{color:var(--accent-deep);font-weight:600}
+    .layout{display:grid;grid-template-columns:minmax(280px,.72fr) minmax(480px,1fr);gap:clamp(48px,8vw,108px);align-items:start;margin-top:32px}
```
0xdewy, line 49: At viewport widths just above 820px, this desktop grid remains active even though its minimum tracks plus gap exceed the available content width (`280px + 480px + 8vw`). Because `body` hides horizontal overflow, roughly 821–905px-wide windows clip the right side of the policy form and submit button. Please switch to the single-column layout sooner or allow the tracks to shrink so the form remains fully accessible.

crates/bloom/src/main.rs
```diff
@@ -1162,6 +1162,88 @@ fn is_ipc_handler_permission_denied(e: &std::io::Error) -> bool {
...
+) -> Result<serde_json::Value> {
+    let status_path = format!("/wallets/registrations/{name}/status.json");
+    let mut printed_url = false;
+    loop {
+        let res = try_ipc(
+            client,
+            client_endpoint,
+            "read",
+            serde_json::json!({ "path": status_path }),
+        )
```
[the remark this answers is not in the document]
GavinPacini: Verified — this isn't actually a bug. `serde_json::json!`'s expression-interpolation arm expands to `to_value(&$other)` (see `serde_json-1.0.150/src/macros.rs`: `($other:expr) => { $crate::to_value(&$other).unwrap() }`), so `status_path` is borrowed on every iteration, never moved. The loop compiles and runs fine — confirmed with a full `cargo build --workspace --all-targets` and by exercising this exact polling path end-to-end in `crates/bloom/tests/wallet_registration_ipc.rs`. No change needed here.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -2594,6 +2708,32 @@ impl WalletsHandler {
...
+            self.write_permit()?;
+            let status = self
+                .auth_services
+                .require_registration_coordinator()?
+                .stage(&spec.name, now_ms_u64())
+                .await
+                .map_err(err_be)?;
+            tracing::info!(wallet = %spec.name, state = status.state.as_str(), "wallet.registration_staged");
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — real bug. `spec.name` went straight into `RegistrationCoordinator::stage()` with no validation, and the coordinator joins it into a filesystem path (both its existence check and, on completion, `recovery_ack`'s `finalize_passkey_wallet` rename target), so a name like `../../escape` could land that rename outside the keystore root. Fixed in 4527ff8: `Keystore::validate_name` is now `pub` and enforced both in `WalletsHandler::write_new_wallet` (clear error before staging) and inside `RegistrationCoordinator::stage` itself (defense in depth for any caller that reaches the coordinator directly). Added regression tests in both crates (`passkey_new_wallet_rejects_path_traversal_names`, `stage_rejects_path_traversal_names`). Thanks for catching this.

crates/bloom-keystore/src/passkey.rs
```diff
@@ -1506,144 +1435,165 @@ fn unique_rebind_backup_dir(root: &std::path::Path, name: &str) -> std::path::Pa
...
+    })?;
+    let out = FinalizedPasskeyWallet {
+        address: prepared.address,
+        pubkey_hex: prepared.pubkey_hex.clone(),
+        policy: prepared.policy.clone(),
+    };
+    // The temp dir was just renamed away; forget `prepared` so its `Drop`
+    // does not try to remove a path that no longer exists at `tmp_dir`.
+    std::mem::forget(prepared);
+    Ok(out)
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — real bug. Forgetting the whole `PreparedPasskeyWallet` skipped its `Drop`, so `recovery_key: Zeroizing<String>` never actually zeroized, defeating the point of wrapping it. Fixed in 4527ff8 exactly as suggested: removed the `mem::forget` and let `prepared` drop normally — `tmp_dir` is already gone after the rename so `Drop`'s `remove_dir_all` is a harmless no-op (its `Result` was already ignored), and the drop is what actually zeroizes the key. Thanks for catching this.

crates/bloom-daemon/src/ceremony_server/wallet_registration.rs
```diff
@@ -0,0 +1,273 @@
...
+            axum::http::header::CONTENT_TYPE,
+            "image/svg+xml; charset=utf-8",
+        )],
+        FAVICON_DARK,
+    )
+}
+
+fn font_response(font: &'static [u8]) -> impl IntoResponse {
+    ([(axum::http::header::CONTENT_TYPE, "font/woff2")], font)
+}
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — this was a regression from moving these routes out of the old foreground passkey server, which did set `Cache-Control: no-cache` on these same assets. Fixed in 4527ff8: `favicon_light`/`favicon_dark`/`font_response` now all set `Cache-Control: no-cache`, matching the ceremony HTML page's own header.

0xdewy (commented):
Follow-up: same-name wallet creation has a pre-existing keystore concurrency race. Tracked in #124; this should be fixed separately and does not need to block this PR.

crates/bloom-daemon/src/wallet_registration.html
```diff
@@ -0,0 +1,610 @@
...
+    PENDING_RECEIPT=outcome.receipt;
+    document.getElementById('recovery-key-text').textContent='0x'+outcome.recovery_key;
+    document.getElementById('done-address').textContent=outcome.address;
+    showPanel('panel-recovery');
+  }catch(e){
+    btn.disabled=false;
+    st.className='err';
+    var m=e.message||String(e);
+    if(m.indexOf('connection')!==-1||m.indexOf('Receiving end')!==-1||m.indexOf('extension')!==-1){
+      st.innerHTML='A browser extension (e.g. Bitwarden) is interfering with passkey creation.<br><br>Fix: in the extension settings, disable passkey management for this browser. <b>Do not use incognito on Linux</b> — passkeys are not saved to the OS keychain from incognito mode. Alternatively, install Google Chrome and disable Bitwarden passkeys there.<br><br>Error: '+m;
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — `m` is built from a server-supplied error string (or a browser exception message) and this page later holds the recovery key in memory, so this was a real risk. Fixed in 2ac83e6: the extension-warning branch now builds the message with `document.createTextNode`/`createElement` and sets the tail via `textContent` instead of `innerHTML`, so no part of the error text is ever parsed as markup.

crates/bloom/src/main.rs
```diff
@@ -1162,6 +1162,88 @@ fn is_ipc_handler_permission_denied(e: &std::io::Error) -> bool {
...
+            "completed" => return Ok(status),
+            "failed" | "expired" | "cancelled" => {
+                let err = status
+                    .get("error")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or(state);
+                anyhow::bail!("passkey registration {state}: {err}");
+            }
+            _ => tokio::time::sleep(std::time::Duration::from_millis(1500)).await,
+        }
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — the daemon's own sweep_expired should reach a stale session first (runs every 60s), but if that background task ever wedges there was no backstop and this loop would spin forever. Fixed in 2ac83e6: poll_wallet_registration_ipc now compares against status.expires_at_ms each iteration and bails with a clear timeout error once it's passed.

crates/bloom/src/main.rs
```diff
@@ -1162,6 +1162,88 @@ fn is_ipc_handler_permission_denied(e: &std::io::Error) -> bool {
...
+                    "passkey registration {}: {}",
+                    status.state.as_str(),
+                    status.error.as_deref().unwrap_or("no further detail")
+                );
+            }
+            AwaitingUser | AwaitingRecoveryAck => {
+                tokio::time::sleep(std::time::Duration::from_millis(1500)).await
+            }
+        }
+    }
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — same gap as the IPC poller (comment above), just in the in-process/command-scoped fallback path. Fixed in 2ac83e6: poll_wallet_registration_inproc now compares against status.expires_at_ms each iteration and bails with a clear timeout error rather than looping indefinitely.

crates/bloom/src/main.rs
```diff
@@ -1854,30 +1936,96 @@ async fn run(cli: Cli) -> Result<()> {
...
+                "write",
+                serde_json::json!({ "path": "/wallets/new", "bytes_b64": B64.encode(name.as_bytes()) }),
+            )
+            .await
+            .with_context(|| format!("ipc write via {}", client_endpoint.display))?;
+
+            let address = if ipc_res.is_some() {
+                debug!(endpoint = %client_endpoint.display, "cli.wallet.new.via_ipc");
+                let status = poll_wallet_registration_ipc(&client, &client_endpoint, &name).await?;
+                status
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — status/daemon.json's wallet_registration_protocol_version does exist for exactly this purpose but nothing was reading it before this. Fixed in 2ac83e6: bloom wallet new's IPC path now reads status/daemon.json first and bails with a clear "restart bloom serve" error if the field is missing or below bloom_auth_api::WALLET_REGISTRATION_PROTOCOL_VERSION, before ever attempting to stage a registration.

crates/bloom/src/main.rs
```diff
@@ -1899,26 +2106,73 @@ async fn run(cli: Cli) -> Result<()> {
...
+                return Ok(());
+            }
+
+            // Passkey-import: foreground-only, not routed through the VFS
+            // asynchronous registration coordinator (which does not accept
+            // a caller-supplied private key — see
+            // docs/plans/2026-07-21-async-vfs-passkey-registration.md). The
+            // private key never leaves this process; it is held only in
+            // memory for the duration of the ceremony below.
+            if d.keystore.info_unverified(&name).is_ok() {
```
josh-richardson, line 2141: **High — validate the wallet name before using it in filesystem paths.** `info_unverified()` performs name validation, but `.is_ok()` discards validation errors, so names such as `../x`, `a/b`, empty strings, or absolute paths continue into `temp_id`, `prepare_passkey_wallet`, and `root.join(&name)`. The temporary-directory cleanup can therefore target paths outside the keystore root, and the final directory can escape it as well. Call `Keystore::validate_name(&name)?` before the existence check or ceremony, and add rejection tests for traversal, separators, empty, and absolute names.

crates/bloom-daemon/src/lib.rs
```diff
@@ -1997,10 +1998,33 @@ impl Daemon {
...
+        // whether this process ends up being `bloom serve` or a one-shot
+        // CLI command.
+        let registration_coordinator: Arc<dyn bloom_auth_api::WalletRegistrationCoordinator> =
+            Arc::new(registration::RegistrationCoordinator::new(
+                keystore.clone(),
+                auth_verifier.clone(),
+                auth_verifier.clone(),
+                home.keystore_dir(),
+            ));
+        if let Err(e) = auth_verifier.mark_unfinished_wallet_registrations_failed_sync(
```
josh-richardson: **High — constructing any `Daemon` can fail another process's live registration.** `Daemon::from_home` is also used by one-shot/read-only commands such as `wallet list`; while `bloom serve` owns an active ceremony, a second CLI process reaches this unconditional reconciliation and persists that session as failed even though the owner still has valid in-memory state. Move restart reconciliation out of the generic constructor and run it only after acquiring exclusive ceremony/listener ownership, before arming the coordinator. A subprocess test should stage via `serve`, run a read-only CLI command, and confirm the session remains `awaiting_user`.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,991 @@
...
+            let recovery_key = prepared.recovery_key.clone();
+            session.winner = Some(WinnerState {
+                prepared,
+                receipt: receipt.clone(),
+            });
+
+            (wallet, address, recovery_key, receipt)
+        };
+
+        let mut status = self
```
josh-richardson, line 755: **High — `/complete` races cancellation and expiry after releasing the state lock.** The winner is installed under the mutex, but the persisted transition to `AwaitingRecoveryAck` happens later. In between, `cancel_by_token` or `sweep_expired` can remove/drop the session and persist a terminal state; depending on write order, this code can overwrite that terminal state or return a recovery key whose acknowledgment now 404s. Serialize lifecycle transitions with a per-session async lock or phase/generation CAS so completion, cancellation, and expiry choose one atomic winner; add barrier tests for both orderings.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,991 @@
...
+        let (wallet, address, recovery_key, receipt) = {
+            let mut state = self.state.lock();
+            let session = state
+                .sessions
+                .get_mut(token)
+                .filter(|s| s.expires_at_ms > now_ms)
+                .ok_or_else(|| {
+                    AuthApiError::NotFound("unknown or expired registration session".into())
+                })?;
+            if session.winner.is_some() {
```
josh-richardson: **Medium-high — a lost successful `/complete` response is unrecoverable.** Once `winner` is committed, retrying the exact same request is rejected, but the recovery key and receipt exist only in the lost response. The browser can no longer acknowledge the registration, so it remains stuck until expiry. Make completion idempotent for an exact replay by storing the attempt ID plus a digest of the verified request and the one-time outcome; return the cached outcome only when that same request is retried.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,991 @@
...
+        token: &str,
+        receipt: &str,
+        now_ms: u64,
+    ) -> Result<String, AuthApiError> {
+        let (wallet, finalize_result) = {
+            let mut state = self.state.lock();
+            let session = state.sessions.get_mut(token).ok_or_else(|| {
+                AuthApiError::NotFound("unknown or expired registration session".into())
+            })?;
+            let winner = session.winner.take().ok_or_else(|| {
```
josh-richardson: **Medium-high — recovery acknowledgment is destructive before finalization is durable.** `winner.take()` transfers the only prepared wallet into a fallible rename. If rename fails, dropping the prepared value removes the temp directory and retry reports no completed attempt. Conversely, a DB failure or lost response after a successful rename leaves the real wallet installed while the token/status cannot be completed idempotently. Model this as retryable phases: preserve the prepared wallet on rename failure, retain a finalized result until `Completed` is durable, and return the same address for a repeated valid receipt; reconcile an already-installed matching wallet at startup.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,991 @@
...
+
+    async fn recovery_ack(
+        &self,
+        token: &str,
+        receipt: &str,
+        now_ms: u64,
+    ) -> Result<String, AuthApiError> {
+        let (wallet, finalize_result) = {
+            let mut state = self.state.lock();
+            let session = state.sessions.get_mut(token).ok_or_else(|| {
```
josh-richardson, line 790: **Medium — recovery acknowledgment has inconsistent expiry semantics.** This lookup does not check `expires_at_ms`, while entering `AwaitingRecoveryAck` does not extend the original five-minute deadline. A completion near expiry may give the user only seconds to save the key; an acknowledgment after the advertised expiry succeeds if it beats the 60-second sweeper and fails if it loses that race. Define and persist a separate recovery-ack deadline (or consistently enforce the original one), and make acknowledgment and sweeping compare-and-swap the same phase/deadline with boundary tests.

crates/bloom-daemon/src/wallet_registration.html
```diff
@@ -0,0 +1,627 @@
...
+  var attemptId='ack'; // path segment is unused server-side; token+receipt authorize this call
+  fetch(BASE+'/attempts/'+attemptId+'/recovery-ack',{
+    method:'POST',headers:{'Content-Type':'application/json'},
+    body:JSON.stringify({receipt:PENDING_RECEIPT})
+  }).then(jsonOrThrow).then(function(){
+    PENDING_RECEIPT=null;
+    showPanel('panel-done');
+  }).catch(function(e){
+    btn.disabled=false;
+    showTerminal('Could not finish registration',e.message||String(e));
```
josh-richardson: **Medium — a transient acknowledgment error removes the only retry path.** `showTerminal` hides the recovery panel, so although the receipt remains in memory, the recovery key and acknowledgment button become inaccessible; reloading loses both. Keep or restore the recovery panel on failure, preserve the key/receipt, show an inline error, and re-enable the button. A DOM test can reject the first fetch and confirm the second acknowledgment succeeds.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,991 @@
...
+                .filter_map(|token| {
+                    state.sessions.remove(&token).map(|session| {
+                        state.by_wallet.remove(&session.wallet);
+                        (token, session.wallet)
+                    })
+                })
+                .collect()
+        };
+        for (token, wallet) in expired {
+            if let Ok(Some(mut status)) = self.store.wallet_registration_status(&wallet).await
```
josh-richardson: **Medium — expiry persistence failures are silently converted to success.** The session is removed from memory first, then both status-read and status-write errors are ignored. A stale `awaiting_user` row can retain a dead ceremony URL and keep the partial unique live-registration index occupied until restart reconciliation. Persist/CAS the terminal transition before discarding the session, or retain a pending tombstone and propagate the error so the background task retries; cancellation should use the same transition mechanism.

crates/bloom-daemon/src/ceremony_server/wallet_registration.rs
```diff
@@ -0,0 +1,285 @@
...
+) -> Response {
+    let coordinator = match coordinator_or_404(&state) {
+        Ok(c) => c,
+        Err(status) => return status.into_response(),
+    };
+    match coordinator.complete(&token, &attempt, body, now_ms()).await {
+        // The only place `recovery_key`/`receipt` are ever serialized —
+        // never logged (this response body is not traced) and never
+        // written to any persisted status/session record.
+        Ok(outcome) => Json(serde_json::json!({
```
josh-richardson, line 209: **Low — the response containing the plaintext recovery key lacks an explicit no-store policy.** The HTML page's cache headers do not cover this POST response. Return `Cache-Control: no-store, no-cache, must-revalidate` (and, if supporting legacy intermediaries, `Pragma: no-cache`) with the completion JSON; apply the same policy to other token-bearing ceremony responses.

josh-richardson (commented):
Cross-functional review findings are attached inline.

crates/bloom/src/main.rs
GavinPacini: Confirmed — a real, high-severity path-traversal bug. `info_unverified` calls `validate_name` first, so an invalid name like `../x` returns `Err`, which `.is_ok()` treats identically to "not found," letting it flow into `prepare_passkey_wallet`'s and `finalize_passkey_wallet`'s filesystem-path joins. Fixed in f505de5: `Keystore::validate_name(&name)` is called explicitly before the existence check, with a CLI regression test for traversal names, path separators, empty, and absolute paths.

crates/bloom-daemon/src/lib.rs
GavinPacini: Confirmed — traced this to ~19 call sites that call `Daemon::from_home`, several genuinely read-only (`wallet list`, `status`, `vfs cat/ls/stat`), all unconditionally hitting this reconciliation. Fixed in f505de5: moved reconciliation out of `from_home_inner` and into `ceremony_server::spawn`, gated on this process having just proven exclusive listener ownership via a successful bind (right before `mark_listener_bound`). Added `WalletRegistrationCoordinator::reconcile_after_restart` and a subprocess regression test that stages via a live `bloom serve`, runs `wallet list` concurrently, and confirms the session stays `awaiting_user`.

crates/bloom-daemon/src/registration.rs
GavinPacini: Confirmed. The clearest instance: `cancel_by_token` and `sweep_expired` both removed the whole session (including an already-installed winner) unconditionally — a racing cancel against `/complete`, or a completion landing near the original session deadline, could destroy an already-completed, awaiting-ack prepared wallet after the browser already had a real recovery key. Fixed in 57b1e44: `cancel_by_token` now rejects once `session.winner.is_some()` instead of removing it, and `sweep_expired`/`recovery_ack` use a separate, later recovery-ack deadline (set when `/complete` installs the winner) instead of the original ceremony deadline, checked consistently by both. This doesn't add a generic phase/generation CAS across all four operations, but it closes the specific races you identified by making "has a winner" itself the thing cancel/expiry must respect. Full end-to-end barrier tests for both orderings would need a real completed registration, which needs a WebAuthn credential — no virtual-authenticator test harness exists anywhere in this codebase, so I added a unit test confirming cancellation is unaffected pre-completion plus the deadline-selection logic, and flagged the harness gap rather than building one as a side effect of this fix.

crates/bloom-daemon/src/registration.rs
GavinPacini: Confirmed — the recovery key/receipt only ever existed in the one HTTP response, and a retry after a lost response hit "already completed" with no way to recover them. Fixed in 57b1e44: a digest of the verified request (attempt id + credential + prf_output_b64) is stored alongside the installed winner, and a second `/complete` for the same attempt with a matching digest returns the same cached outcome (recovery_key/receipt are still sitting in the winner's `prepared` wallet, so no extra secret storage needed) instead of erroring. Added a unit test (`complete_body_digest_is_stable_and_sensitive_to_every_field`) covering the digest's sensitivity to every distinguishing field.

crates/bloom-daemon/src/registration.rs
GavinPacini: Confirmed — `finalize_passkey_wallet` took `PreparedPasskeyWallet` by value, so a failed rename dropped it inside the function before returning, and `Drop` unconditionally `remove_dir_all`s the temp dir regardless of whether the rename actually happened — destroying the still-valid wallet and its recovery key on a merely transient failure (disk full, cross-device, permissions). Fixed in 57b1e44: `finalize_passkey_wallet` now returns `Err((Box<PreparedPasskeyWallet>, KeystoreError))` on a failed rename, and `recovery_ack` reinstates `session.winner` on that path so a retried acknowledgment with the same receipt can still succeed. Added keystore-level unit tests for both the success path and the hand-back-on-failure path. Reconciling an already-installed matching wallet at startup (your last suggestion) isn't included — flagging as a follow-up rather than folding into this fix, since it's a separate reconciliation concern from the immediate data-loss bug.

crates/bloom-daemon/src/registration.rs
GavinPacini: Confirmed — `recovery_ack`'s session lookup had no expiry filter at all, while `sweep_expired` used the original `expires_at_ms` for every session regardless of whether a winner had been installed, so a completion landing near that deadline risked the sweeper deleting an awaiting-ack wallet within seconds. Fixed in 57b1e44: added `RECOVERY_ACK_TTL_MS`, a separate deadline set on the winner when `/complete` succeeds, and both `recovery_ack` and `sweep_expired` now check that deadline (falling back to `expires_at_ms` only when there's no winner) instead of the original ceremony deadline. Boundary tests for the exact expiry instant aren't included — flagging alongside the other winner-lifecycle tests that need a real completed registration (see the reply on the `/complete` race comment for why).

crates/bloom-daemon/src/wallet_registration.html
GavinPacini: Confirmed — `showTerminal` switches to the dead-end `panel-terminal`, which has no button back, even though `PENDING_RECEIPT` is still valid in memory and a reload loses it by design. Fixed in e4cdb1f: an ack failure now stays on the recovery panel, shows an inline error (`#ack-error`), and re-enables the Continue button so the same receipt can be retried. A DOM test rejecting the first fetch and confirming a second succeeds isn't included — no JS/DOM test harness exists anywhere in this codebase (confirmed absent), and standing one up felt disproportionate to this one fix; verified via a JS syntax check and by reading the resulting control flow instead.

crates/bloom-daemon/src/registration.rs
GavinPacini: Confirmed — the session was removed from memory first, then both the status read and write used `if let Ok(...)`/`let _ =`, so a transient store error left a stale row (dead `ceremony_url`, possibly blocking the wallet's unique live-registration index) with nothing left in memory to retry it. Fixed in e4cdb1f: the session now stays in memory until its persisted status is actually written `Expired`, so a failed persist gets picked up by the next sweep tick instead of being lost; failures are logged via `tracing::warn!` with wallet/reason, and `sweep_expired` now returns `Result<usize, _>` (mirroring `bloom_tx::Outbox::sweep_expired`) so the existing background-task warn branch has something to report. Cancellation already goes through the same `upsert_wallet_registration_status` path as expiry, though it doesn't yet share the retry-on-failure treatment — flagging as a smaller follow-up rather than folding into this pass.

crates/bloom-daemon/src/ceremony_server/wallet_registration.rs
GavinPacini: Confirmed — only `page()` set Cache-Control; `session.json`, `/attempts`, `/fallback-options`, `/complete` (the one carrying the plaintext recovery key), `/recovery-ack`, and `/cancel` had none. Fixed in e4cdb1f: applied `Cache-Control: no-store, no-cache, must-revalidate` uniformly across the whole token-keyed route group via a response-header layer (`no_store_cache_control`), rather than repeating it per handler — so a future route added to this group can't accidentally skip it. Didn't add `Pragma: no-cache` for legacy intermediaries since this is a loopback-only listener with no intermediary in the path, but happy to add it if you'd still like the belt-and-suspenders. Added an HTTP-level regression test asserting the header on both the page and a JSON endpoint against the real pre-bound listener.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,1268 @@
...
+            }
+        };
+
+        let url = Self::ceremony_url(&base, &token);
+        let status =
+            WalletRegistrationStatus::awaiting_user(wallet, created_at_ms, expires_at_ms, url);
+        self.writer
+            .upsert_wallet_registration_status(&token, &status, now_ms)
+            .await?;
+        Ok(status)
```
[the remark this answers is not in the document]
GavinPacini: Confirmed. Fixed in cc4ca26: stage() now distinguishes 'found an existing live session' from 'inserted a fresh one' and only constructs/persists an awaiting_user status for the fresh case — re-staging over an existing live session now returns the current persisted status unchanged instead of downgrading it.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,1268 @@
...
+        let mut status = self
+            .store
+            .wallet_registration_status(&wallet)
+            .await?
+            .ok_or_else(|| AuthApiError::Store("registration session status missing".into()))?;
+        status.state = WalletRegistrationState::AwaitingRecoveryAck;
+        status.address = None; // not committed until recovery is acknowledged
+        self.writer
+            .upsert_wallet_registration_status(token, &status, now_ms)
+            .await?;
```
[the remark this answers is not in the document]
GavinPacini: Confirmed. Fixed in cc4ca26: complete()'s persisted status now sets expires_at_ms = now_ms + RECOVERY_ACK_TTL_MS on the awaiting_recovery_ack transition, matching the in-memory recovery-ack deadline.

crates/bloom-daemon/src/registration.rs
```diff
@@ -0,0 +1,1268 @@
...
+                None => {
+                    // Terminal/expired in-memory session for this wallet, if
+                    // any: drop it before starting a fresh one.
+                    if let Some(stale) = state.by_wallet.remove(wallet) {
+                        state.sessions.remove(&stale);
+                    }
+                    Self::insert_fresh_session(&mut state, wallet, now_ms)
+                }
+            }
+        };
```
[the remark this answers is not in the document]
GavinPacini: Confirmed — this was the root cause behind several of these findings. Fixed in cc4ca26: extracted SecretSession::effective_deadline() (the recovery-ack deadline once a winner is installed, else the original expiry) and applied it uniformly across stage(), session_view(), complete()'s replay handling, and sweep_expired, so they all agree on how long an awaiting-ack session stays alive.

PR #151, opened 2026-08-09
Restore VFS ergonomics and require daemon-only CLI IPC

GavinPacini:
## Summary

This PR repairs the approved VFS and CLI regressions found during the triad-architecture audit:

- advertise wallet outbox `new.tx` consistently as a writable sink;
- restore the documented plain-name `/wallets/new` body while leaving the asynchronous Broker registration protocol unchanged;
- route every Petal CLI operation through the running daemon;
- make every CLI operation other than `init` and `serve` a strict IPC proxy, including `--version`, wallet cancel/replace, request planning, update checks, and completions;
- preserve request and staged-transaction identities atomically across IPC;
- preserve CLI stdout, stderr, exit behavior, caller-relative Petal paths, generated artifacts, and source-build output;
- reject missing and non-regular wallet outbox artifacts consistently across list, lookup, and read;
- document the daemon-only boundary in Interaction Modes.

The IPC expansion also closes the local trust boundary around daemon-owned commands: Unix peers must have the daemon's effective UID, sockets are securely staged and atomically published under a persistent endpoint lock, raw signer-consuming writes remain denied, and Machine commands/errors use closed typed contracts.

## Why

The triad rewrite exposed Broker-oriented schemas and daemon internals directly through some VFS and CLI surfaces. In particular, wallet registration began requiring JSON, several commands silently constructed local daemons or talked directly to Broker, IPC responses lost operation identities or output, and some outbox metadata advertised impossible files.

The intended boundary is now explicit: `bloom init` initializes state, `bloom serve` owns runtime authority and state, and every other command is a client of that running daemon.

## Base-branch reconciliation

This branch now includes `triad-architecture` through `43ed476`.

- The base's petname registration work is combined with this PR's ergonomic sink: `/wallets/new` still accepts a plain wallet name, while status, cancellation, and results are projected at `wallets/registrations/<petname>/...`. The random Broker operation identity remains internal to the projection.
- The base's approval-contract alignment is adopted unchanged, including the updated Petal contract and Broker/runtime revisions.

### EVM exact-approval expiry

EVM exact signing persists a ten-minute approval window together with its approval/signing operation IDs, request nonce, and approval ID in `ceremony.json`. When matching state is reloaded after expiry, the scalar and batch paths retain those terminal identities. Broker therefore continues rejecting an otherwise unchanged staged payload instead of starting a fresh exact-approval attempt.

The Petal exact-signing paths already demonstrate the recovery shape: rotate approval/signing operation IDs and the request nonce, refresh timestamps, and clear the old approval ID while retaining exact payload and provenance comparisons.

This PR deliberately does not change the EVM flow. Correcting it needs a dedicated scalar-and-batch identity-retirement design, atomic persistence rules, and protocol-level tests; folding it into CLI/VFS compatibility work would make the approval semantics much harder to review.

### Sealed Approval sinks

The Sealed Approval sinks are low-level Broker orchestration surfaces for reusable, bounded signing authority rather than ordinary transaction-staging endpoints:

- `sealed-approvals/new.json` accepts `ApprovalPrepareRequest` and prepares a new approval ceremony;
- `<approval-id>/renew` accepts `ApprovalRenewRequest` and prepares replacement terms;
- `<approval-id>/revoke` accepts `RevokeRequest` and revokes one approval with a reason;
- `revoke_all` accepts `WalletOperationRequest` and revokes wallet-scoped approvals.

They currently expose raw Broker request schemas. That is useful as a narrow orchestration layer, but it is not an ergonomic VFS builder API. This PR leaves those schemas unchanged so a later protocol/API change can address them deliberately.

### Petal contract alignment (resolved by the merged base)

The original audit found Bloom pinned to Petal commit `4f6fb570…` while newer payload-signing interfaces were being supported elsewhere, creating a risk that `contract_wit_digest()`, SDKs, builders, and conformance tooling would disagree about the canonical contract.

The latest `triad-architecture` now pins `bloom-petal-contract` to `eda6647c523bba161eaa22812aa0e75ec7782404` and updates the associated host-interface and approval-contract handling. This merge adopts that alignment, so the old-pin discrepancy is no longer deferred work in this PR. Keeping this note documents the audit finding and why no separate pin change appears in the feature commits.

## Validation

- `cargo test -p bloom --test cli` — 62 passed
- `cargo test -p bloom-daemon ipc::tests -- --test-threads=1` — 28 passed
- `cargo test -p bloom-vfs handlers::wallets::tests -- --test-threads=1` — 37 passed
- `git diff --check triad-architecture...HEAD`

## Baseline and merge verification context

Before the base update, full-suite runs reproduced the exact-Petal `PROVENANCE_MISMATCH` family in the daemon approval retry and VFS exact-signing retry tests. The newly merged base aligns the Petal, Broker, and runtime contract dependencies that govern this area.

After resolving the merge, `cargo check -p bloom-vfs -p bloom-daemon -p bloom`, the CLI integration suite (62/62), and daemon IPC suite (28/28) passed. Per request, the remaining suites were not rerun after conflict resolution; the earlier focused wallet VFS suite was 37/37 before the base merge.

## Checklist

- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — see `docs/architecture/Agent-native Documentation.md`

### DIFF
--- a/crates/bloom-daemon/src/ipc.rs
+++ b/crates/bloom-daemon/src/ipc.rs
@@ -11,14 +11,16 @@
 //! | `lookup`   | `{ "path": "/..." }`                  | `{ "name", "kind", ... }` |
 //! | `read`     | `{ "path": "/..." }`                  | `{ "bytes_b64": "..." }`  |
 //! | `write`    | `{ "path": "/...", "bytes_b64": "" }` | `null`                    |
+//! | `write_with_lookup` | write params plus `projection_path` | identity entry           |
 //! | `list`     | `{ "path": "/..." }`                  | `[ entry, ... ]`          |
 //! | `version`  | `null`                                | `"x.y.z"`                 |
 //! | `chains`   | `null`                                | `[ "ethereum", ... ]`     |
 //! | `confirm_batch` | `{ "wallet", "txs", "text" }` | batch result              |
-//! | `petals.install` | `{ "path", "ref"? }`             | package metadata          |
-//! | `petals.build` | `{ "package_dir", "out"? }`       | package metadata          |
+//! | `petals.install` | remote source or package transport | package metadata          |
+//! | `petals.build` | `{ "package_dir", "out"? }`           | package metadata       |
 //! | `petals.list` | `null`                              | `[ package, ... ]`        |
 //! | `petals.uninstall` | `{ "hash" }`                   | `{ "removed" }`          |
+//! | `machine.execute` | tagged [`MachineCommand`]        | [`MachineCommandOutput`] |
 //! | `shutdown` | `null`                                | `null`                    |
 //!
 //! Wire framing is one JSON document per line. Encoding/decoding errors
@@ -27,6 +29,7 @@
 
 use std::collections::BTreeMap;
 use std::future::Future;
+use std::io::Write as _;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::sync::Arc;
@@ -50,6 +53,10 @@ pub enum IpcError {
     Io(#[from] std::io::Error),
     #[error("home dir does not exist: {0}")]
     NoHome(PathBuf),
+    #[error("refusing insecure IPC socket {path}: {reason}")]
+    InsecureSocket { path: PathBuf, reason: String },
+    #[error("IPC endpoint is already served: {0}")]
+    EndpointBusy(PathBuf),
 }
 
 /// Default socket path under `<home>/run/bloom.sock`.
@@ -57,13 +64,17 @@ pub fn default_socket_path(home_root: &Path) -> PathBuf {
     home_root.join("run").join("bloom.sock")
 }
 
-/// RAII guard that removes the socket file on drop, covering normal exit
-/// and panics during [`IpcServer::serve`].
-struct SocketGuard(PathBuf);
-impl Drop for SocketGuard {
-    fn drop(&mut self) {
-        let _ = std::fs::remove_file(&self.0);
-    }
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct SocketIdentity {
+    device: u64,
+    inode: u64,
+}
+
+/// Keeps the persistent endpoint ownership record locked for the listener's
+/// full lifetime. The record is deliberately not removed on drop: unlinking a
+/// lock path would let a raced publisher lock a different inode.
+struct EndpointLock {
+    _file: std::fs::File,
 }
 
 #[derive(Debug, Deserialize)]
@@ -87,12 +98,56 @@ struct Response {
     error: Option<RpcError>,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Deserialize, Serialize)]
 struct RpcError {
     code: i32,
     message: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     data: Option<Value>,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MachineErrorKind {
+    InvalidParams,
+    PermissionDenied,
+    Unavailable,
+    Conflict,
+    NotFound,
+    Internal,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, Error, PartialEq, Serialize)]
+#[error("{message}")]
+pub struct MachineError {
+    pub kind: MachineErrorKind,
+    pub code: String,
+    pub message: String,
+}
+
+impl MachineError {
+    pub fn new(
+        kind: MachineErrorKind,
+        code: impl Into<String>,
+        message: impl Into<String>,
+    ) -> Self {
+        Self {
+            kind,
+            code: code.into(),
+            message: message.into(),
+        }
+    }
+
+    fn rpc_code(&self) -> i32 {
+        match self.kind {
+            MachineErrorKind::InvalidParams => -32602,
+            MachineErrorKind::PermissionDenied => -32007,
+            MachineErrorKind::Unavailable => -32003,
+            MachineErrorKind::Conflict => -32009,
+            MachineErrorKind::NotFound => -32004,
+            MachineErrorKind::Internal => -32603,
+        }
+    }
 }
 
 impl Response {
@@ -113,6 +168,19 @@ impl Response {
                 code,
                 message: message.into(),
                 data: None,
+            }),
+        }
+    }
+
+    fn machine_err(id: Value, error: MachineError) -> Self {
+        Self {
+            jsonrpc: "2.0",
+            id,
+            result: None,
+            error: Some(RpcError {
+                code: error.rpc_code(),
+                message: error.message.clone(),
+                data: Some(serde_json::to_value(error).expect("Machine error serializes")),
             }),
         }
     }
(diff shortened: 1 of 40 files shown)


crates/bloom/tests/cli.rs
```diff
@@ -42,16 +41,134 @@ fn bloom_cmd(home: &Path) -> Command {
...
+            .env("RUST_LOG", "error")
+            .arg("serve")
+            .stdout(std::process::Stdio::null())
+            .stderr(std::process::Stdio::piped());
+        if !automatic_update_checks {
+            command.env(bloom_update::DISABLE_AUTO_CHECK_ENV, "1");
+        }
+        let mut child = command.spawn().expect("start Bloom daemon for CLI test");
+        let socket = bloom_daemon::ipc::default_socket_path(home);
+        for _ in 0..200 {
```
josh-richardson: This gives the daemon only 2 seconds to start. The current Actions run already failed `raw_ipc_call_reports_invalid_machine_params_without_json_error_wrappers` and `request_cli_dry_run_uses_vfs_lifecycle_and_body_receipt_helpers` here because their sockets were not ready, and I reproduced the same load-dependent failure locally; the suite passes when rerun without competing compilation. Please use a materially longer deadline (while retaining the early child-exit/stderr path) so host load does not turn unrelated CLI tests red.

crates/bloom/src/github_source.rs
```diff
@@ -177,18 +229,31 @@ fn install_github_source_with_expectation(
...
+                }));
+            }
+            return Err(error);
+        }
+    };
 
-    println!("Validating Petal package...");
+    let completion_progress = vec!["Validating Petal package...".to_owned()];
     let package =
         PreparedPetalPackage::from_dir(tmp.path()).context("validate generated Petal package")?;
```
josh-richardson: `build_output` has already been captured here, but this `?` and every later validation/consent/install error return without it. The failing Polymarket E2E demonstrates the regression: the build succeeds, validation fails, and the CLI gets empty stdout instead of the build report that explains what was produced. Please carry progress plus bounded stdout/stderr through every post-build failure, not only `SourceBuildCommandFailure`.

crates/bloom/src/main.rs
```diff
@@ -2632,165 +2922,289 @@ async fn run(cli: Cli) -> Result<()> {
...
+            Err(error) => {
+                if let Some(failure) = github_source::source_build_failure(&error) {
+                    return Ok(serde_json::json!({
+                        "progress_lines": failure.progress,
+                        "build_stdout": failure.stdout,
+                        "build_stderr": failure.stderr,
+                        "operation_error": failure.message,
+                    }));
+                }
+                return Err(error.to_string());
```
josh-richardson: `anyhow::Error::to_string()` retains only the outer context. In the current external E2E this reduces the actual validation failure to `validate generated Petal package`; running the same package through the base path reveals the useful cause (`bloom:tx/outbox@0.1.0` has an invalid Bloom WIT interface shape). Please serialize the full chain (for example formatted with `{error:#}` or, preferably, structured causes) across IPC.

crates/bloom/src/github_source.rs
```diff
@@ -862,20 +935,29 @@ fn run_source_build(root: &Path) -> Result<()> {
     if !command.is_file() {
         bail!("build command missing: {}", build.command);
     }
-    let status = Command::new(&command)
+    let output = Command::new(&command)
         .current_dir(root)
-        .status()
+        .output()
```
josh-richardson: `Command::output()` buffers unlimited stdout and stderr in the long-running authority-owning daemon. Both buffers are then converted to strings and copied into one JSON IPC response, so an accidentally noisy build can exhaust daemon memory. Please stream/spool with explicit byte limits, or truncate each stream at a documented cap with a marker.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -2518,8 +2516,14 @@ impl WalletsHandler {
                     .outbox
                     .read_in_state(wallet, chain, id, st)
                     .map_err(err_be)?;
-                let path = entry.dir.join(fname.as_str());
-                let bytes = std::fs::read(&path).map_err(HandlerError::Io)?;
+                let path = regular_outbox_artifact(&entry.dir, fname)?;
+                let bytes = std::fs::read(&path).map_err(|error| {
```
josh-richardson: There is a check/open race here: `regular_outbox_artifact` verifies the pathname with `symlink_metadata`, returns it, and this call opens it again. The entry can become a symlink or another nonregular object between those operations, bypassing the rejection this change is intended to enforce. Please open once with no-follow semantics, validate the opened descriptor with `fstat`, and read from that same descriptor.

josh-richardson (changes requested):
Requesting changes for the inline findings: the daemon startup deadline is already causing CI flakes, source-install failures lose their actionable build output and error chain, build output is buffered without bounds in the daemon, and outbox artifact validation has a check/open race.

crates/bloom-daemon/src/ipc.rs
```diff
@@ -623,60 +1504,297 @@ impl IpcClient {
...
+            "bloom_protocol": client_protocol,
         });
         let mut out = serde_json::to_vec(&req).unwrap();
+        if out.len() > MAX_IPC_FRAME_BYTES {
+            return Err(IpcClientError::Protocol(format!(
+                "IPC request exceeds {MAX_IPC_FRAME_BYTES} bytes"
+            )));
+        }
         out.push(b'\n');
+        let stream = UnixStream::connect(&self.socket_path).await?;
```
josh-richardson, line 1946: The server authenticates connecting clients, but the client never authenticates the daemon before sending this request. With an explicit/shared endpoint, a foreign-UID process can pre-bind the path and forge wallet addresses, ceremony output, command output, or exit status. Please verify `stream.peer_cred().uid()` against the expected Machine UID before writing, validate endpoint metadata where applicable, and add a foreign-UID server test.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -517,24 +536,54 @@ impl WalletsHandler {
...
+            && projection
+                .ceremony_expires_at_ms
+                .as_ref()
+                .is_some_and(|expires_at| expires_at.get() <= now_ms_u64())
+        {
+            // Ceremony launch data is already bound to a trusted expiry. Do
+            // not turn an expired, persisted registration into filesystem EIO
+            // merely because the Broker no longer retains the operation (for
+            // example after restarting a local development triad).
+            projection.ceremony_state = bloom_broker_api::CeremonyState::Expired;
```
josh-richardson: This can permanently misclassify a completed registration as expired. If the owner completes just before the deadline and Machine polls just after it, this branch persists `Expired` without asking Broker; the state is then terminal, `result.json` never appears, and a retry can launch another registration. Please reconcile with Broker before committing a terminal state. If Broker is unavailable, clear the stale URL while retaining a retryable state, and add a completion-at-expiry race test.

crates/bloom-daemon/src/ipc.rs
```diff
@@ -175,26 +504,32 @@ impl IpcServer {
...
+        let _endpoint_lock = acquire_endpoint_lock(&target)?;
+        let staging = private_socket_staging_dir(&target)?;
+        // Keep the unpublished path no longer than a typical endpoint name;
+        // sockaddr_un has a small fixed path limit on macOS.
+        let staged_socket = staging.path().join("s");
+        let listener = std::os::unix::net::UnixListener::bind(&staged_socket)?;
+        use std::os::unix::fs::PermissionsExt as _;
+        std::fs::set_permissions(&staged_socket, std::fs::Permissions::from_mode(0o600))?;
+        let identity = verify_socket_security(&staged_socket, rustix::process::geteuid().as_raw())?;
+        let replacing_stale = validate_stale_socket(&target)?;
```
josh-richardson: `validate_stale_socket` treats every same-UID socket as stale, so the following rename can replace a live listener that does not participate in this new lock convention (for example an older daemon or a different-home daemon using the same explicit endpoint). That orphans the live process and creates split-brain ownership. Please probe and refuse connectable sockets, revalidate immediately before publication, and test with an independently bound listener that has no `.lock` file.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -215,17 +209,18 @@ impl WalletsHandler {
...
+        let Some(projections) = &self.wallet_projections else {
+            return Vec::new();
+        };
+        match projections.list_wallets().await {
+            Ok(wallets) => wallets,
+            // Root directory enumeration is navigation, not an authority
+            // decision. Prefer a previously authenticated cache when the live
+            // edge is unavailable, and keep `new`/`registrations` reachable
+            // even if no safe cached projection remains.
+            Err(_) => projections.cached_wallets().unwrap_or_default(),
```
josh-richardson: Falling back for every error masks protocol-integrity failures such as malformed, duplicate, or identity-invalid Broker projections. The production cached reader already limits fallback to `ServiceUnavailable`; this wrapper can instead list stale wallets while lookup/read rejects the same state. Please preserve the error type, fall back only for `ServiceUnavailable`, propagate other failures, and test a populated cache with a non-availability error.

crates/bloom-daemon/src/ipc.rs
```diff
@@ -260,6 +622,14 @@ impl IpcServer {
                 br#"{"jsonrpc":"2.0","id":null,"error":{"code":-32603,"message":"internal error"}}"#
                     .to_vec()
             });
+            if out.len() > MAX_IPC_FRAME_BYTES {
```
josh-richardson: This cap has no streaming path. `read` base64-encodes unrestricted VFS bytes into one response, so `bloom vfs cat`/`request body` fail at roughly 3 MiB; large writes and command output likewise fail now that there is no local fallback. Please add chunked/length-prefixed transport or documented per-method limits before work is performed, with tests immediately below and above the effective base64 threshold.

crates/bloom/src/github_source.rs
```diff
@@ -862,20 +961,81 @@ fn run_source_build(root: &Path) -> Result<()> {
...
+        let read = reader.read(&mut buffer)?;
+        if read == 0 {
+            break;
+        }
+        let remaining = SOURCE_BUILD_STREAM_LIMIT.saturating_sub(retained.len());
+        let keep = remaining.min(read);
+        retained.extend_from_slice(&buffer[..keep]);
+        omitted = omitted.saturating_add(read - keep);
+    }
+    let mut output = String::from_utf8_lossy(&retained).into_owned();
```
josh-richardson: This does not preserve the base CLI's stdout/stderr behavior: each stream is capped at 256 KiB, non-UTF-8 is replaced lossily, and the client later replays all stdout before all stderr only after completion. The base inherited the foreground descriptors and preserved bytes with live diagnostics. Please stream bounded base64 byte chunks to the corresponding descriptor and test non-UTF-8, oversized, and long-running output. The early `?` returns at lines 223-242 should also retain already-earned progress.

crates/bloom-daemon/src/ipc.rs
```diff
@@ -384,12 +809,136 @@ impl IpcServer {
...
+                request.package_dir.display()
+            ))
+        })?;
+        let output = request
+            .out
+            .as_deref()
+            .map(|path| validated_petal_archive_output(&package_dir, path))
+            .transpose()?;
+        let mutation = self.petal_mutation.clone().lock_owned().await;
+        tokio::task::spawn_blocking(move || {
```
josh-richardson, line 1236: Foreground cancellation is detached from this daemon-side mutation. The connection handler waits for dispatch before it can observe a closed client, so Ctrl-C can return control while this task continues regenerating caller artifacts, publishing an archive, or completing an install. Please add disconnect-aware cancellation with atomic cleanup, or explicitly model these as durable operations returning reconcilable IDs; test disconnect during a blocked build/install.

docs/architecture/Interaction Modes.md
```diff
@@ -9,7 +9,43 @@ Bloom exposes actions through three client interaction modes:
 2. the foreground `bloom vfs` facade; and
 3. a VFS mounted by a long-running Bloom Machine.
 
-All three modes use the same production authority plane. Machine stages,
+The `bloom` CLI has exactly two local lifecycle commands: `bloom init` creates
```
josh-richardson, line 12: The implementation resolves `BLOOM_RPC_ENDPOINT` before command dispatch, so the two lifecycle namespaces described here still depend on unrelated client configuration. For example, `BLOOM_RPC_ENDPOINT=tcp:invalid bloom init` fails before initialization, and hidden installer/service submodes under `init`/`serve` are affected too. Please defer client endpoint resolution until after excluding those branches and add lifecycle tests with invalid client endpoint configuration.

crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -517,24 +536,54 @@ impl WalletsHandler {
...
             ceremony_kind: prepared.ceremony_kind,
             ceremony_state: bloom_broker_api::CeremonyState::AwaitingUser,
             ceremony_url: Some(prepared.ceremony_url),
             ceremony_expires_at_ms: Some(prepared.ceremony_expires_at_ms),
             signer_contribution_digest: prepared.signer_contribution_digest,
         };
-        write_atomic_json(&registration_path, &projection)
+        write_atomic_json(&registration_path, &projection)?;
+        if let Some(legacy_path) = legacy_registration_path {
+            std::fs::remove_file(legacy_path)?;
```
josh-richardson, line 592: The canonical projection is durable before the legacy operation-ID file is deleted. A crash or deletion failure here leaves two identical records for one petname, after which duplicate rejection prevents registration list, lookup, and retry. Please treat identical-operation duplicates as recoverable, prefer the canonical record with best-effort cleanup, and add a dual-file recovery test.

josh-richardson (changes requested):
Requesting changes for the nine inline findings. The highest-risk issue is that CLI clients do not authenticate the daemon peer; the remaining findings cover registration-state correctness, socket publication, projection integrity, framing/output compatibility, cancellation semantics, lifecycle command isolation, and crash-safe registration migration.

PR #167, opened 2026-08-17
fix: bump wasmtime 26 -> 36.0.13 to clear advisories

0xdewy:
## Summary

Bump the workspace `wasmtime`/`wasmtime-wasi` pin inherited by
`bloom-petals` from 26 to 36.0.13 — the lowest release clearing the critical
sandbox-escape and component-model/WASI advisory batch — and update
`crossbeam-epoch` from 0.9.18 to 0.9.20.

The Wasmtime API adaptation is one line: the in-memory stdio pipes moved from
`wasmtime_wasi::pipe` to `wasmtime_wasi::p2::pipe`.

The upgrade also revealed an upstream-default change: Wasmtime 36 accepts core
`memory64` modules by default, whereas Wasmtime 26 rejected them. Bloom keeps
`memory64` disabled and now tests the accepted core-Wasm proposal profile inside
actual Petal components.

## Runtime profile vs. Petal ABI

`memory64` is not a `bloom:route@0.1.0` ABI upgrade. The Petal ABI is the
versioned WIT interface and digest shared by the SDK, builder, package validator,
and host. Wasmtime's accepted core-Wasm proposals are a separate host runtime
profile.

Enabling `memory64` here would be premature:

- Wasmtime still tracks component-model memory64 support as incomplete in
  [upstream issue #4311](https://github.com/bytecodealliance/wasmtime/issues/4311).
- Changing the WIT package version would not complete that engine support and
  would not make a memory64 Petal compatible with Bloom hosts still using
  Wasmtime 26.
- Bloom has no package-level runtime-feature or minimum-host-version negotiation
  for Petals yet. Accepting memory64 silently would let a package pass on a new
  host and fail on an older one.
- The existing `MemLimiter` applies the same byte cap to 32-bit and 64-bit
  memories, so this is a compatibility/support boundary rather than a newly
  discovered memory-exhaustion hole.

The regression test covers `memory64`, threads/shared memory, wide arithmetic,
multi-memory, and tail calls inside components. A future engine bump must now
make changes to that profile explicit.

No Petal packages need rebuilding for this change. Petals consume the versioned
WIT/SDK contract and produce portable components; they do not link Wasmtime.
The Solana driver also has no Wasmtime dependency, so its committed component is
unchanged. Enabling memory64 later should be a separate feature-design change
with engine conformance coverage, package/runtime feature negotiation, builder
validation, and a stated minimum Bloom host version.

## Advisory list — before → after

| Crate | Before | After |
|---|---|---|
| wasmtime | 26.0.1 | **36.0.13** |
| wasmtime-wasi | 26.0.1 | **36.0.13** |
| crossbeam-epoch | 0.9.18 | **0.9.20** |

Cleared from the Wasmtime/Cranelift/WASI set:

- **RUSTSEC-2026-0095** (9.0) — Winch sandbox escape
- **RUSTSEC-2026-0096** (9.0) — aarch64 Cranelift miscompile
- **RUSTSEC-2026-0093** (6.9) — component-model UTF-16→latin1 OOB read
- **RUSTSEC-2026-0091 / -0094** (6.1) — component-model string OOB / `table.grow` mask
- **RUSTSEC-2026-0085 / -0089 / -0092** (5.6–5.9) — component-model lifting panics
- **RUSTSEC-2026-0086 / -0087 / -0088 / -0222** — low/medium
- **RUSTSEC-2025-0046 / -0118**, **RUSTSEC-2026-0020 / -0021** — WASIp1/WASI-http
- wasmtime-wasi **RUSTSEC-2026-0149 / -0182 / -0188** — WASI filesystem permissions
- crossbeam-epoch **RUSTSEC-2026-0204**

The 24.x line is insufficient: the two critical advisories and several others
have no 24.x patch branch and require `>=36.0.7`; RUSTSEC-2026-0222 requires
`>=36.0.13`. Version 36.0.13 is the lowest single release clearing the batch.

## Current audit residuals

`cargo audit` reports 5 vulnerabilities, none in the Wasmtime set and none
introduced by this PR:

| Crate | Advisory | Patched version |
|---|---|---|
| h2 0.4.14 | RUSTSEC-2026-0258 | >=0.4.16 |
| quinn-proto 0.11.14 | RUSTSEC-2026-0185 | >=0.11.15 |
| rkyv 0.7.46 | RUSTSEC-2026-0235 | >=0.8.17 |
| ruint 1.18.0 | RUSTSEC-2026-0220 | >=1.20.0 |
| tracing-subscriber 0.2.25 | RUSTSEC-2025-0055 | >=0.3.20 |

These predate the change and remain follow-up work rather than blockers for the
Wasmtime advisory fix.

## Verification after rebase

- `cargo fmt --all -- --check` — clean.
- `cargo check --workspace --locked` — passed.
- `cargo clippy -p bloom-petals --all-targets --locked -- -D warnings` — passed.
- `cargo test -p bloom-petals --locked` — 169 unit tests and 2 integration fixtures passed.
- `cargo audit --file Cargo.lock` — no Wasmtime/Cranelift/WASI findings; 5 unrelated residuals above.

## Checklist

- [x] Tests added or updated for behavior changes — component-level runtime-profile regression test added; full `bloom-petals` suite passes.
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A; the WIT/route ABI is unchanged and the runtime profile preserves pre-upgrade behavior.
- [x] Sealed Approval invariants respected — N/A; no authorization or signing behavior changed.
- [x] Agent Documentation and affected Petal READMEs updated — N/A; no Petal authoring or agent-facing route behavior changed.

### DIFF
--- a/crates/bloom-petals/src/vm.rs
+++ b/crates/bloom-petals/src/vm.rs
@@ -39,7 +39,7 @@ use parking_lot::Mutex;
 use wasmtime::component::{Component, Linker as ComponentLinker, Val as ComponentVal};
 use wasmtime::{Caller, Config, Engine, Linker, Memory, Module, Store, StoreContextMut};
 use wasmtime_wasi::WasiCtxBuilder;
-use wasmtime_wasi::pipe::{MemoryInputPipe, MemoryOutputPipe};
+use wasmtime_wasi::p2::pipe::{MemoryInputPipe, MemoryOutputPipe};
 use wasmtime_wasi::preview1::{self, WasiP1Ctx};
 
 use crate::abi::{
--- a/crates/bloom-petals/src/vm.rs
+++ b/crates/bloom-petals/src/vm.rs
@@ -196,6 +196,18 @@ impl PetalVm {
         config.cranelift_nan_canonicalization(true);
         config.wasm_relaxed_simd(true);
         config.relaxed_simd_deterministic(true);
+        // Keep memory64 outside the Petal runtime profile. This is separate
+        // from the `bloom:route@0.1.0` WIT ABI: changing that interface would
+        // not make 64-bit linear memories safe or compatible with older
+        // hosts. Wasmtime 36 enables the core proposal by default even though
+        // its component-model integration is still incomplete (tracked by
+        // upstream issue #4311). The `MemLimiter` cap applies to either
+        // address width, so this is a compatibility/support boundary rather
+        // than a memory-exhaustion fix. Enable it only with an explicit Petal
+        // runtime-feature contract and component-level conformance coverage.
+        // `accepted_petal_component_surface_is_pinned` below catches future
+        // Wasmtime default changes.
+        config.wasm_memory64(false);
         let engine = Engine::new(&config).map_err(|e| PetalError::vm(e.to_string()))?;
         Ok(Self {
             engine,
@@ -2971,6 +2983,46 @@ mod tests {
             after > before,
             "dropping a Wasmtime store must drop its owned memory limiter"
         );
+    }
+
+    /// The core-Wasm proposals accepted inside a Petal component form a host
+    /// runtime profile, separate from the component's `bloom:route@0.1.0` WIT
+    /// ABI. Most are inherited from moving Wasmtime defaults. Upgrading
+    /// Wasmtime 26 -> 36 silently flipped `memory64` from rejected to accepted;
+    /// this test makes the next such change an explicit compatibility choice.
+    ///
+    /// Adding a row here is fine. Flipping one changes the runtime profile.
+    #[test]
+    fn accepted_petal_component_surface_is_pinned() {
+        let vm = PetalVm::new().expect("engine");
+        // (proposal, core module body, accepted inside a component?)
+        let cases: &[(&str, &str, bool)] = &[
+            ("memory64", r#"(memory i64 1)"#, false),
+            ("threads/shared-memory", r#"(memory 1 1 shared)"#, false),
+            (
+                "wide-arithmetic",
+                r#"(func (result i64 i64)
+                     (i64.const 1) (i64.const 2) (i64.mul_wide_s))"#,
+                false,
+            ),
+            ("multi-memory", r#"(memory 1) (memory 1)"#, true),
+            ("tail-call", r#"(func $a) (func $b (return_call $a))"#, true),
+        ];
+        for (name, module_body, want_accepted) in cases {
+            // A proposal can be refused either by the text parser or by
+            // component validation; for this profile both count as rejected.
+            let component = format!("(component (core module {module_body}))");
+            let accepted = wat::parse_str(&component)
+                .ok()
+                .is_some_and(|bytes| Component::from_binary(&vm.engine, &bytes).is_ok());
+            assert_eq!(
+                accepted, *want_accepted,
+                "accepted Petal component surface changed for the {name} proposal: \
+                 expected accepted={want_accepted}, got accepted={accepted}. \
+                 If this is intended, update the runtime profile and its \
+                 compatibility documentation."
+            );
+        }
     }
 
     /// Petal that writes to stdout via `fd_write`. We do this by
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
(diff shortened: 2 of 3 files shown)


0xdewy:
Reviewed. The bump does what it says — I verified the advisory claim rather than taking it on trust, and found one issue, now fixed in 75756d7.

## The advisory claim checks out

`cargo audit` on each branch:

| | vulnerabilities |
|---|---|
| `master` | 25 |
| this branch | 5 |

All 5 remaining are unrelated crates — `h2`, `quinn-proto`, `rkyv`, `ruint`, `tracing-subscriber`. Every wasmtime/cranelift advisory is cleared, and `36.0.13` satisfies the tightest constraints in the set (RUSTSEC-2026-0085 and -0086 both require `>=36.0.7, <37.0.0`).

Worth noting separately: **`quinn-proto` RUSTSEC-2026-0185 is 7.5 HIGH** (remote memory exhaustion, fixed in `>=0.11.15`) and predates this PR. Also, **nothing in CI runs `cargo audit`** — there is no advisory job in any workflow, so this whole class of problem regresses silently. Both are follow-ups, not blockers here.

## The issue: the bump silently widened the sandbox

Ten majors moved a default that this file depends on. Same probe against both engines:

| proposal | wasmtime 26 (`master`) | wasmtime 36 (this PR) |
|---|---|---|
| **memory64** | **rejected** | **ACCEPTED** |
| threads / shared memory | rejected | rejected |
| wide-arithmetic | rejected | rejected |
| multi-memory | accepted | accepted |
| tail-call | accepted | accepted |

`wasmparser`'s `WasmFeatures` default for `memory64` flipped to `true` between 0.218.1 and 0.236.1, and `Config::new()` inherits it. Wasmtime's own doc comment on `wasm_memory64` still reads *"This is `false` by default"* in 36.0.13, so reading the docs does not catch this — I only found it by compiling a 64-bit-memory module against both engines.

**Impact is narrower than it first looks.** `MemLimiter::memory_growing` works in bytes, so the 16 MiB cap applies to a 64-bit memory exactly as it does to a 32-bit one — this is not a memory-exhaustion hole. What it is: the set of modules a Petal may contain is part of the pinned ABI, and it changed as a side effect of a dependency upgrade rather than as a decision. That directly cuts against the determinism rationale already in the config block above it — a Petal built against a memory64-capable host would fail on a host still on 26.

## Fix (75756d7)

- `config.wasm_memory64(false)` — restores the pre-bump surface.
- `accepted_module_surface_is_pinned` — a table test over the five proposals above. I verified it fails with the pin removed (`expected accepted=false, got accepted=true`) and passes with it, so it will actually catch the next silent move. Adding a row is fine; flipping one is an ABI decision.

`cargo fmt`, `clippy -p bloom-petals --all-targets`, and the full `bloom-petals` suite (167 + 2 tests) all pass.

One thing I noticed but deliberately did **not** touch, since it is pre-existing on `master` and out of scope here: `MemLimiter::table_growing` returns `Ok(true)` unconditionally, so table growth is uncapped while memory growth is capped. Worth its own issue.

crates/bloom-petals/src/vm.rs
```diff
@@ -196,6 +196,17 @@ impl PetalVm {
...
+        // upstream default for memory64 is not stable across Wasmtime
+        // versions: it was off under Wasmtime 26 and on under 36, so a
+        // routine version bump would otherwise start accepting 64-bit
+        // linear memories that no Petal ABI was pinned against. The
+        // `MemLimiter` cap still applies to such a memory, so this is a
+        // surface-widening concern rather than a memory-exhaustion one,
+        // but widening it is a decision to take deliberately and not a
+        // side effect of a dependency upgrade. `accepted_module_surface_
+        // is_pinned` below fails if a future bump moves this again.
+        config.wasm_memory64(false);
```
GavinPacini, line 210: Let's maybe upgrade the Petal ABI so we don't need to disable this?

0xdewy: I dug into this before rebasing, and I do not think a Petal ABI bump is the right fix here. `bloom:route@0.1.0` versions the WIT interface; `memory64` is a separate core-Wasm/runtime feature. Changing the WIT version would neither complete Wasmtime support nor make a memory64 component work on older Bloom hosts. Wasmtime still tracks component-model memory64 as incomplete in bytecodealliance/wasmtime#4311 (with current panic cases even on newer releases), and Bloom has no package-level runtime-feature/minimum-host negotiation yet. I kept it disabled, rewrote the comment to make that distinction explicit, and changed the regression test to validate the proposal matrix inside actual Petal components rather than standalone modules. No Petal rebuild is needed because Petals link the WIT SDK, not Wasmtime. If we want memory64 later, I think it should be a separate feature-design PR covering engine conformance, builder/package feature declaration, and minimum host compatibility. Rebased update: d6ce0d4.

PR #169, opened 2026-08-18
BIP-39 wallet.accounts projection, CLI, VFS, and derived-child signing

### LINKED ISSUE #163: feat(wallet): make BIP-39 multi-curve HD wallets the default
## Summary

Make newly created local Bloom wallets deterministic, multi-chain HD wallets
backed by one Signer-owned BIP-39 root entropy value. Passkeys remain the normal
unlock and authorization UX: each WebAuthn credential's PRF output independently
wraps the same wallet key-encryption key (WKEK), which encrypts the root. The
blockchain keys are derived from the root, not directly from a passkey.

This is the prerequisite for adding Solana without forcing users to create and
recover an unrelated Ed25519 wallet. It blocks #156.

## Why

- Users should manage one Bloom wallet identity, not one private key per chain.
- Adding, replacing, or recovering a passkey must not change any chain address.
- EVM secp256k1 and Solana Ed25519 require different deterministic derivation
  schemes, but can share one BIP-39 seed.
- BIP-39 import/export gives a portable recovery format while remaining absent
  from normal daily UX.
- OWS-compatible paths and CAIP identifiers improve interoperability without
  adopting OWS's password vault or bearer-token custody model.

## Architectural decisions

1. **Passkeys wrap; they do not define the wallet.** Signer generates 256 bits of
   BIP-39 entropy. WKEK encrypts it, and each credential-specific WebAuthn PRF
   output wraps the same WKEK. Direct `seed = KDF(passkey_prf)` derivation is
   prohibited because every credential has different PRF material.
2. **The root is not a signing key.** Introduce a typed wallet seed/root reference
   distinct from a signable `KeyRef`. Every chain account is a derived, typed
   child `KeyRef` pinned to its derivation profile and public key.
3. **Use standard derivation.** Initially support BIP-32/BIP-44 secp256k1 for EVM
   and hardened SLIP-10 Ed25519 for Solana. Paths are allocated by a Signer-owned
   registry; callers cannot supply arbitrary paths.
4. **BIP-39 is the default for new local wallets.** Generate 24 English words
   (256-bit entropy). Use the empty BIP-39 passphrase in the default interoperable
   profile; any advanced passphrase mode must be explicit and versioned.
5. **Recovery is optional but first-class.** The mnemonic is shown/imported only
   in a Broker-hosted, Signer-verified custody ceremony. Machine, Petals, logs,
   audit records, and agent capabilities never receive it.
6. **Do not regress triad authority.** Agent operations receive scoped child
   references and bounded approvals, never a seed-decrypting bearer token.

### Contracts and storage

- Define versioned `WalletSeedProfile`, `DerivationProfile`, `WalletSeedRef`, and
  derived-child metadata across Broker/Signer APIs.
- Store encrypted BIP-39 entropy, derivation-profile version, allocated paths,
  tombstones, public fingerprints, and recovery metadata in the Signer backup set.
- Bind root/child relationship, path, profile, key spec, and public fingerprint
  into canonical operation and audit identities.
- Preserve the existing multi-passkey WKEK wrapping and recovery invariants.

### Derivation

- EVM default: BIP-44 coin type 60 using BIP-32 secp256k1, with one canonical
  account/address allocation policy.
- Solana default: BIP-44 coin type 501 using hardened SLIP-10 Ed25519,
  `m/44'/501'/<account>'/0'`.
- Add official BIP-39, BIP-32, SLIP-10, Ethereum-address, and Solana-address
  golden vectors plus cross-profile negative vectors.
- Make derivation authority-extending: allocation requires a custody ceremony or
  an explicitly budgeted policy capability.

### Ceremonies and UX

- Make HD creation the default local-wallet registration flow.
- Add mnemonic import, guarded export, recovery, and verification ceremonies.
- Support adding/replacing/removing passkeys without changing the root or child
  addresses.
- Project chain accounts and CAIP-2/CAIP-10 identifiers without exposing secret
  material.

### Migration

- Inventory the exact secret material retained by existing secp256k1 wallets.
- Preserve all existing EVM addresses and approvals.
- If an existing wallet lacks recoverable source seed entropy, do not reinterpret
  an extended secp256k1 private key as a SLIP-10 seed. Keep it as a legacy profile
  or perform an explicit owner-approved migration to a new root.
- Make migration restart-safe, idempotent, backup-aware, and rollback-safe before
  committing the new profile.

### Compatibility

- Support standard BIP-39 mnemonic import/export and documented derivation paths.
- Adopt CAIP identifiers where account/chain identities cross generic boundaries.
- Treat Open Wallet Standard as an interoperability adapter only; do not adopt its
  `~/.ows` storage authority or API-token-encrypted copies of wallet secrets.

## Out of scope

- Deriving blockchain keys directly from WebAuthn credential keys or PRF output.
- Persisting plaintext mnemonics, seeds, WKEKs, PRF outputs, or child private keys.
- Hardware-wallet derivation, threshold custody, social recovery, or arbitrary
  user-supplied derivation paths in the first release.
- Automatically moving funds from legacy addresses.

## Acceptance criteria

- [ ] A fresh local wallet creates one encrypted 256-bit BIP-39 root and exposes
      the canonical first EVM child without showing a mnemonic by default.
- [ ] The same root reproduces official BIP-39/BIP-32 and SLIP-10 vectors and
      deterministically projects typed EVM and Solana child accounts.
- [ ] Two independently registered passkeys unlock the same root and addresses;
      replacing either credential does not change them.
- [ ] Mnemonic export/import and loss-of-passkeys recovery require explicit
      Broker custody ceremonies and never expose the mnemonic to Machine/Petals.
- [ ] Derivation paths are registry-allocated, collision-free, tombstoned after
      removal, and included in backup/restore.
- [ ] Existing EVM wallets preserve their address and remain usable under an
      explicit legacy or migrated profile.
- [ ] Cross-curve, wrong-path, altered-profile, replay, missing-registry, and
      wrong-public-key cases fail closed.
- [ ] Agent capabilities cannot decrypt or export the wallet root.
- [ ] Architecture docs, recovery guidance, audit inventory, and release gates
      describe the final model.

## Delivery ordering

1. Freeze profiles, schemas, vectors, migration classification, and threat model.
2. Implement Signer storage/derivation/backup primitives.
3. Extend Broker/Signer custody ceremonies and public projections.
4. Make HD creation the default and add mnemonic recovery UX.
5. Migrate or classify legacy wallets and complete release verification.
6. Unblock #156 only after the Ed25519 child derivation edge is released and pinned.

0xdewy:
## What this PR does

This PR adds BIP-39 hierarchical-deterministic (HD) wallet support to Bloom's Machine process: importing a recovery phrase produces a wallet with multiple derived chain accounts (not one single key), those accounts are projected to the CLI and the virtual filesystem, additional child accounts can be allocated through an owner ceremony, and signing requests can name the exact child they intend to spend from — failing closed instead of guessing when the choice is ambiguous.

## Background

Bloom is a cryptocurrency wallet split into three cooperating processes (the "triad"):

- **Machine** (this repository): the CLI, the mounted virtual filesystem (VFS), public read-only projections of wallet state, unsigned transaction staging, broadcast, and reconciliation. It holds no private keys.
- **Broker**: hosts WebAuthn/passkey "ceremonies" (the browser pages where the owner types secrets and approves actions), interprets policy, and authorizes operations.
- **Signer**: stores encrypted key material and produces signatures. Machine never talks to Signer directly; everything goes through Broker.

Before this PR, a wallet had exactly one signable root key (`root_key_ref` in the wallet projection), either generated at registration or imported as a raw secp256k1 scalar. There was no notion of a seed with derived children, no way to represent several accounts under one wallet, and `bloom wallet import` meant "import a raw private key".

This PR depends on matching BIP-39 custody changes in the sibling Broker and Signer repositories, which are consumed as exact-revision Git dependencies pinned in `Cargo.toml`/`Cargo.lock` (see the `bloom-broker-api` pins). The Machine-side changes here are meaningful only against those pinned revisions.

### CLI (`crates/bloom/src/main.rs`, `crates/bloom-daemon/src/ipc.rs`)

- `bloom wallet import <name>` now starts a **BIP-39 mnemonic import ceremony**. The recovery phrase (12/15/18/21/24 English words, no BIP-39 passphrase — passphrase-protected mnemonics are unsupported) is typed only into the Broker-hosted ceremony browser. There is deliberately **no mnemonic command-line argument, flag, file, or environment variable**; Machine never sees the phrase.
- `bloom wallet import <name> --raw-private-key` is the explicit path for the old raw secp256k1 key import. The key is likewise entered only in the browser. The result is an `imported-secp256k1-scalar` wallet: a single-key EVM wallet that **cannot derive Solana or other child accounts**.
- `bloom wallet accounts <name>` prints the wallet's derived-account projection (JSON).
- `bloom wallet account-allocate <name> --profile bip44-solana-slip10-ed25519-v1` starts a Broker-hosted `AccountAllocate` ceremony that derives one new Solana (SLIP-10 Ed25519) child under the wallet's BIP-39 root. Requesting the EVM profile explicitly fails: import already creates the canonical EVM child, and further EVM children stay unavailable until EVM transaction surfaces can carry an explicit account selector.
- The import mode is bound to a matching input-class/seed-profile pair (`CustodyInputShape`), so a mnemonic ceremony and a raw-scalar ceremony cannot be silently interchanged mid-flight.

### Protocol / projection (`crates/bloom-machine-client`, `crates/bloom-daemon`)

- `WalletPublic.root_key_ref` becomes optional. A BIP-39 wallet has **no signable root** (`None`); its seed is not a key, and only its derived children can sign. Projection validation is updated accordingly (a BIP-39 projection claiming a root key is rejected as inconsistent).
- New Broker requests `WalletAccounts` and `AccountAllocatePrepare`, with client methods `wallet_accounts()` and `account_allocate()`.
- `CustodyPrepareRequest` gains `wallet_seed_profile`, `derivation_request`, and `account_terms`. Allocation terms (`AccountTerms`) bind wallet id, seed profile, derivation profile, path template, key spec, allowed crypto suites, policy version, revocation epoch, a fresh replay id, and a 30-minute expiry; their digest is what Broker authorizes. A fresh authoritative Broker wallet projection is fetched when constructing terms, rather than reusing a presentation cache.
- Import always creates the canonical EVM child at `m/44'/60'/0'/0/0`. The top-level EVM address resolves **only** from that path; Broker list order is never a fallback (pinned by `bip39_primary_key_never_falls_back_to_broker_list_order` in `projection.rs`).

### Exact signing (`crates/bloom-machine-client/src/lib.rs`)

- `ExactPayloadSignRequest` / `ExactPayloadBatchSignRequest` gain `account_key_ref: Option<KeyRef>` (a `KeyRef` is the full public identity of a key: backend, locator, key spec, public-key fingerprint, and derivation path).
- `verified_wallet_root` becomes `verified_signing_key`: root-signing wallets keep their old behavior; BIP-39 wallets resolve among derived children matching the requested crypto suite. With exactly one matching child it resolves implicitly; with several, omission or an unrecognized selector **fails closed** with an error naming every candidate's fingerprint and derivation path, so the caller can act on the error. A selector on a root-signing wallet is refused.
- The selected `KeyRef` is bound into `SealedApprovalTerms::key_ref` when an approval is prepared and into `SignOperationIdentity::key_ref` when it is spent, so an approval issued for one account can never authorize a signature from another. Existing EVM call sites pass `None` in this PR (v1 never exposes a second EVM child).

### VFS (`crates/bloom-vfs/src/handlers/wallets.rs`)

- New read-only leaf `wallets/<wallet>/accounts.json`: the authenticated Broker projection of a wallet's derived accounts (fingerprints, derivation paths, lifecycle state, supported suites, chain address projections). It contains no secrets. A regression test pins the mounted list → lookup → read contract for it.
- New `outbox/latest` symlink (EVM chains) pointing at the most recent pending outbox entry, giving an atomic "current staged transfer" identity.

### Custody hygiene and boundary checks

- Retired Machine-local wallet-secret test harnesses are removed: `BLOOM_TEST_WALLET_PASSPHRASE` and friends disappear from the Docker fork/mempool harness, `scripts/acceptance.sh` is rewritten as a thin custody acceptance entrypoint, and `scripts/play.sh` no longer provisions passphrase wallets.
- `packaging/triad/release/check-machine-authority-boundary.sh` gains `require_no_legacy_wallet_secret_inputs`, which fails the release boundary scan if any production source or script reintroduces passphrase-style inputs (`--passphrase`, `--passphrase-file`, passphrase environment markers, ...). `test-machine-authority-boundary.sh` mutation-tests it with a synthetic `--passphrase-file` input.

### Documentation

- `docs/architecture/Wallet.md` records the BIP-39 profile, import rules, and the legacy-wallet migration boundary; `docs/plans/2026-08-14-bip39-multicurve-hd-wallets.md` captures the design; the mounted agent guidance (`crates/bloom-vfs/src/docs/agent-guidance.md`) documents `accounts.json` and the selection rules.

## How it works: mnemonic import and first send

1. `bloom wallet import vacation` asks Broker to prepare a `WalletImport` ceremony with input class `bip39-mnemonic` and seed profile `bip39-multicurve-v1`, and prints the ceremony URL.
2. The owner opens the URL and types the phrase into the Broker-hosted browser page. Broker binds the ceremony; Signer stores the encrypted root and derives the canonical EVM child `m/44'/60'/0'/0/0`. Machine receives only the public result.
3. `bloom wallet accounts vacation` (or `cat wallets/vacation/accounts.json` on the mount) now shows the derived child: fingerprint, path, suites, chain projections.
4. Existing staging/signing flows work unchanged for the single EVM child: `verified_signing_key` resolves it implicitly. If a future wallet held two EVM children, the same call would refuse to guess and would name both candidates.

## Breaking changes

- **`bloom wallet import` changed meaning.** On `master` it launched the raw-private-key import ceremony; after this PR it launches the BIP-39 mnemonic ceremony. Scripts that used it to import raw keys must add `--raw-private-key`.
- Legacy wallets (raw BIP-32 seed or imported scalar) **cannot** be converted to BIP-39 in place — relabeling would change the key tree, and there is no transparent conversion. The supported migration is: export the old key with the old offline tooling, `wallet import <name> --raw-private-key`, and move assets deliberately. Raw-key imports preserve the EVM address but cannot derive Solana children.

## Testing

- Unit tests in `crates/bloom-machine-client`: derived-child selection (omitted selector with two children fails naming both, single child resolves, each selected child binds its own key, foreign selector refused, root-signing wallet refuses a selector) and projection invariants (no list-order fallback for the primary key).
- VFS handler tests: `accounts.json` listed/looked-up/read consistently; `outbox/latest` advertised and resolving.
- CLI parse tests: import defaults to BIP-39, `--raw-private-key` is explicit, input-class/seed-profile pairs are bound.
- `scripts/test-triad-projection-fidelity.sh` runs the full three-process launcher, registers a wallet, imports the public BIP-39 test fixture through a `0600` protected mnemonic file delivered to the ceremony, asserts the canonical EVM child in CLI output, and compares it byte-for-byte with the mounted `accounts.json`.
- Authority-boundary scripts: extended scan plus a mutation test proving a reintroduced wallet-passphrase input fails the check.

## Notes for reviewers

- The important security property to review is **fail-closed account selection**: nothing may resolve a key by projection order, list position, or cache. Every selection error must name the candidates (fingerprints + paths) so it is actionable rather than a dead end.
- Allocation and (later) retirement are authority ceremonies: Machine projects their public result but cannot mutate custody; projections only reflect them after the Broker ceremony completes.
- This branch is the base for the stacked native-Solana PR, which uses `wallet.accounts`, `account-allocate --profile bip44-solana-slip10-ed25519-v1`, and `account_key_ref` to spend from a chosen Ed25519 child.

## Checklist

- [x] Tests added or updated for behavior changes — `bloom-machine-client` selection/correlation suites, `bloom-vfs` outbox-latest suites, and the cross-process `test-raw-key-import-transfer.sh` / `test-bip39-import-transfer.sh` acceptance lanes.
- [x] Architecture docs — `docs/architecture/Wallet.md` and the BIP-39 plan/spec updated for derived accounts and the imported-scalar profile.
- [x] Sealed Approval invariants respected — account selection binds the exact `KeyRef` into approval and signing identities; no custody material crosses into Machine.
- [x] Agent Documentation updated — `crates/bloom-vfs/src/docs/agent-guidance.md` covers the derived-account surfaces.

Closes #163


crates/bloom-machine-client/src/lib.rs
```diff
@@ -1299,42 +1330,63 @@ impl MachineBrokerClient {
...
+                    .key_refs
+                    .iter()
+                    .filter(|key| key.key_spec == suite.key_spec())
+                    .collect();
+                match matching.as_slice() {
+                    [child] => ((*child).clone(), KeyRole::Derived),
+                    _ => {
+                        return Err(ProtocolError::new(
+                            ProtocolErrorCode::KeyrefMismatch,
+                            "BIP-39 wallet must have exactly one derived child for the requested CryptoSuite",
```
0xdewy: This makes the advertised HD multi-account flow self-defeating: after allocating a second child with the same key spec (for example, two EVM accounts), `matching` has length 2 and every exact/batch/reusable wallet-level signing call fails with `KeyrefMismatch`. Those request types carry only `wallet_id` + suite, so the caller cannot select the intended derived account. Please either add an exact account/key selector and bind it into the approval/signing identity, or explicitly enforce one active derived child per suite (which would materially narrow the multi-account feature). A regression test should allocate two same-suite accounts and sign with each.

0xdewy: Confirmed and fixed. You were right that the flow was self-defeating, and the ordering problem was worse than the `KeyrefMismatch` you found here: two other surfaces resolved the same ambiguity by silently taking the first active child.

**Where it stood.** `verified_signing_key` already refused to guess and already validated an explicitly passed `KeyRef` against wallet membership, key spec, suite, and Broker's public projection. Two things were missing: every caller passed `None`, and the error told you nothing about how to proceed.

**What changed.**

*This PR (`6062db3`, `crates/bloom-machine-client/src/lib.rs`)* — both the "multiple children" and "selected account is not a child" errors now name every compatible child by public-key fingerprint and derivation path, so the error is actionable instead of a dead end. Selection stays fail-closed and never positional: one compatible active child resolves implicitly as before, zero fails, two or more fail.

*#170 (`60a9612`)* — the two surfaces that were selecting by list order:

- `resolve_solana_child` in the VFS wallet handler, which produces the fee payer and transfer source;
- `wallet address --profile solana`.

Both had grown their own `.find(...)` independently, which is *how* they came to disagree with the signing layer that already refused to guess. Rather than fix the same bug twice, selection now lives in one shared `bloom_solana_tx::account` that both call.

**On the selector.** We went with public-key fingerprint (or a unique prefix) rather than account number, per your suggestion of a `KeyRef`/fingerprint. The fingerprint is what `SealedApprovalTerms::key_ref` and `SignOperationIdentity::key_ref` already bind, so the value the user names is the value the approval commits to, with no intermediate mapping that could drift. Account number and list index are rejected with reasons in `docs/design/solana-account-selection.md`. Interface: `account_fingerprint` in `new.tx`, `--fingerprint` on `wallet address`.

**On durability.** `StagedSolanaTransfer` pins the selected child's full fingerprint (never the user's prefix, which could later resolve differently). Confirm, restage, and reconcile re-select *that exact account* rather than resolving the wallet again — otherwise a second active child could sign a message staged for the first. Entries staged before selection existed carry no pin and still work; they cannot have been ambiguous.

**Regressions** (15 across both PRs):

- `bloom-machine-client`: `each_selected_child_binds_its_own_key_and_never_the_other` (asserts the *approval* binds the selected child, not the first), `omitted_selector_with_two_active_children_fails_and_names_both`, `a_single_active_child_still_resolves_without_a_selector`, `a_foreign_selector_is_refused_and_names_the_real_children`, `a_selector_is_refused_on_a_wallet_that_signs_with_its_root`, `candidate_description_names_every_fingerprint_and_path`.
- `bloom-solana-tx::account`: `two_active_children_can_each_be_selected_independently`, `a_unique_prefix_selects_and_a_shared_one_refuses`, `omitting_the_selector_with_two_active_children_fails_and_names_both`, `a_retired_child_is_neither_active_nor_selectable`, `a_foreign_fingerprint_is_refused_and_names_the_real_children`, `a_different_profile_is_never_an_active_solana_candidate` (EVM unchanged).
- `bloom-solana-tx::engine`: `a_pinned_account_must_be_reselected_exactly_to_sign` (covers substitution after staging), `an_entry_staged_before_selection_existed_still_signs`.

Your point that the earlier validation only demonstrated account 1 *after retiring account 0* is addressed directly: `a_retired_child_is_neither_active_nor_selectable` keeps that lifecycle assertion, and `two_active_children_can_each_be_selected_independently` covers the case that was actually missing — both children active at once.

CI green on both PRs (#169 11/11, #170 12/12 including the pinned agave validator job). Leaving this thread open for you to resolve.

0xdewy: Follow-up: the multi-account case you identified is now covered end to end against a real validator, not just by unit tests.

**New evidence** — `bloom-it/tests/solana_multi_account.rs`, `two_active_solana_children_select_sign_and_reconcile_independently` (bloom#170 `6661544`), run against a pinned agave v3.0.0 validator with **both children active for the whole test**:

- staging with no selector fails closed, naming both fingerprints *and* both derivation paths;
- a fingerprint belonging to no active child is refused;
- account 1 is selected while account 0 stays active, and the signing call names account 1's key (`wallet/derived/solana-1`);
- the signature verifies over the raw message against account 1, **does not** verify against account 0, and **does not** verify over SHA-256(message);
- the transfer reaches `confirmation_status: finalized`, the destination is credited, and account 0's balance is untouched;
- a replayed confirm neither signs again nor produces a second sent entry;
- account 0 remains independently selectable afterwards.

Two design points worth flagging for review:

**The fixture signs with whichever key the request's `key_ref` names, and only account 1 is funded.** A selection bug therefore cannot pass by accident — the wrong child produces a signature that fails verification against the staged fee payer, and the validator would reject the transfer too.

**Restart is checked by re-reading durable state through a fresh outbox, not by building a second Daemon.** A `Daemon` holds the home write permit for the life of the process by design, so an in-process restart cannot mean a second Daemon over the same home. What a restarted process actually recovers from is the on-disk record: it still pins account 1, still resolves as `sent`, and cannot be re-read as `pending`, so a restarted process has nothing to confirm and cannot resend.

**Boundary, stated plainly:** the Broker in this suite is an in-process fixture, as in `solana_workflow` and the EVM `anvil_e2e`. The Daemon, VFS, wallet handler, engine, reconciler, and validator are real. This is *not* the separate-process bloom-broker/bloom-signer triad — that remains outstanding and is called out in the module docs rather than left implicit.

CI is green on both PRs (#169 11/11, #170 12/12 including `transfer lifecycle on pinned agave validator`). Still leaving this thread open for you to resolve.

0xdewy:
Requesting **provisional independent approval** of exact head
`21cbaa207e306bf7d79841191aa3aaa7c9a818f5` for the BIP-39 account projection,
CLI/VFS, and derived-child signing layer.

All current CI checks are green. Please treat approval as provisional because
the stack must merge and repin bottom-up. After merge I will provide a
`git range-diff` from this approved head to the final head. Limited
post-merge confirmation is valid only for expected ancestry, immutable pins,
lockfiles, compatibility catalogs, and approved conflict resolution; any
semantic code/policy delta requires full re-review. The implementation
author/operator cannot self-satisfy that confirmation.

0xdewy:
## Review: simplification / code size / code smells

Reviewed `c68db31` against `master`. Claims below verified against the tree.

Framing: of the +1215, ~470 is a design doc, ~200 is tests on one function, ~160 is mechanical `Some(...)` fixture churn. Real new code is ~250 lines and is appropriately sized. The issues are correctness, not size.

### 1. `accounts.json` is listed but not resolvable (blocking)

`crates/bloom-vfs/src/handlers/wallets.rs:2039`

The readdir entry was added (`:1616`) and the read handler was added (`:2195`), but the lookup arm was not extended:

```rust
"address" | "address.qr.png" | "address.qr.svg" | "addresses.json" | "public_key"
| "kind" | "projection.json" => Ok(Entry::file(&segs[1])),   // accounts.json absent
```

`bloom-mount/src/adapter.rs:1127` routes FUSE `lookup` straight to `vfs.lookup`, so under the mount the file appears in `ls` and fails on `stat`/`cat`. One-word fix — but note neither the new CLI commands nor this VFS entry have any test, which is why it survived.

### 2. `account: Some(0)` blocks multi-account and can deadlock signing

`crates/bloom/src/main.rs:829`

The Signer treats `account: None` as "the registry chooses" and, for Solana, advances the account index because the Solana path template carries no address index (`derivation_registry.rs:399-427`). Pinning `Some(0)`:

- every Solana allocation resolves to `m/44'/501'/0'/0'`, so the **second allocation fails** ("canonical derivation path is already allocated");
- for EVM, two successful allocations give two secp256k1 children — and since no caller in this repo sets `account_key_ref`, `verified_signing_key` then returns `KeyrefMismatch` for *every* wallet-level signing path.

Which leads to the structural point:

### 3. The new selector is fail-closed with no way to unblock it

`account_key_ref` is `None` at all five production call sites — `crates/bloom-tx/src/tx_engine.rs:3351`, `:4005`, `crates/bloom-vfs/src/exact_signing.rs:327`, `:586`, `:723` — while the new CLI can create exactly the multi-child state that makes those sites fail permanently. The plumbing is one PR short of usable. Suggest landing the selector with #170 (where a caller exists) rather than merging a path nothing can resolve.

### 4. `account-allocate` output is JSON-encoded twice

`crates/bloom/src/main.rs:1221` wraps `serde_json::to_string_pretty(&prepared)` around a value that is already the human-readable multi-line summary returned by `launch_account_allocation`, so the output is a quoted blob with literal `\n`. Every sibling arm (`WalletCustody` `:1268`, `WalletMigrate` `:1290`) passes the string through. Dropping the wrapper also removes a line.

### 5. `primary_key` guesses, contradicting the PR's own stated rule

`crates/bloom-machine-client/src/projection.rs:92` falls back to `.or_else(|| evm_children.into_iter().next())` — an arbitrary EVM child in Broker listing order — when the canonical `m/44'/60'/0'/0/0` is absent. This feeds `bloom wallet address`, `wallet list`, and the VFS `address`/`address.qr.*` files, while `lib.rs:1409-1412` explicitly documents "Never fall back to the first match". Deleting the fallback collapses the branch to a single `find`.

### Size findings (reviewer estimates)

- **~70 LOC** — `bip39_multi_account_signing_requires_and_honours_an_account_selector` (`lib.rs:3114-3183`) is fully subsumed by the later block at `:4113-4295`; both assert ambiguity-without-selector, per-account approval binding, and foreign-selector refusal. Only the `requests.len() == 1` assertion is unique — worth folding into `a_foreign_selector_is_refused_and_names_the_real_children`.
- **~35 LOC** — `verified_signing_key` (`lib.rs:1357-1455`) is a 100-line nested match that flattens to ~25. `matching.iter().copied().find(|key| *key == selected)` at `:1382` is just a membership test, and the post-checks at `:1425-1436` are unreachable for the BIP-39 branch (the key was selected *from* `wallet.key_refs` filtered on that exact `key_spec`).
- **~25 LOC** — `launch_account_allocation` (`main.rs:779-890`) repeats `launch_custody_ceremony` (`:680-775`): the 32-byte `OperationId` block, the `configured_broker_client` wrapper, and a byte-identical 6-line projection/format tail.
- Related: it reads via `daemon.wallet_projections.get_wallet` (`:805-810`), which deliberately serves stale cached data when the Broker edge is down — and a stale `policy_version`/`revocation_epoch` is exactly what Broker rejects at `verify_account_terms_baseline`. Prefer the `client.wallet(wallet_id)` it already has.
- **~15 LOC** — the hand-written profile parser at `:787-803` duplicates serde's own kebab-case names (`DerivationProfile` derives `Deserialize` with `rename_all = "kebab-case"`), and `Token::new("bloom.account_terms.v1")` at `:833` hardcodes the exported `ACCOUNT_TERMS_SCHEMA`.

### Cross-repo

`AccountTerms` is now hand-built in three places: here (`main.rs:832-848`), re-derived and re-checked in `AccountTerms::validate`, and built a third time in Broker's `allocation_terms()` fixture. An `AccountTerms::allocate(...)` constructor in `bloom-broker-api` would make the frozen shape unrepresentable-if-wrong instead of validated-after-the-fact.

---
*Automated review (Claude), findings verified against the tree by hand.*

0xdewy:
## Disposition of the 2026-08-30 architecture review

All five correctness findings are closed on exact head `15da5d7d75a465c69ffb9bb1f85adaf5503e2a33`:

- `accounts.json` now satisfies list → lookup → read, with regression coverage.
- Account allocation delegates index choice to the Signer (`account: None`), so repeated Solana allocations advance canonically.
- Multi-child signing fails closed with actionable selectors; additional EVM allocation is refused until its transaction surfaces accept an explicit selector.
- Account-allocation output is no longer JSON-encoded twice.
- Primary-key projection never falls back to Broker list order.

The separate-process rehearsal now imports a public BIP-39 fixture through a protected mnemonic file, asserts the canonical EVM child, and compares CLI projection with mounted `accounts.json`. It contains no passphrase field or mnemonic process argument.

Exact-head CI is green. The host used for this cleanup lacks the user-systemd/NFS prerequisites for that final mounted triad rehearsal, so that acceptance lane still must run on the intended release host. This is a self-authored disposition, not independent approval.

0xdewy:
Final import-contract follow-up on exact head `3984d0917296f1300fb23ecc992da4cf82b71191`:

- `bloom wallet import <name>` now starts the passphrase-free BIP-39 mnemonic ceremony by default.
- `--raw-private-key` is the explicit old-EVM-wallet migration mode; the secret is still entered only in the Broker-hosted browser.
- One typed input shape binds each ceremony input class to its matching seed profile, preventing mnemonic/raw reinterpretation.
- There is no BIP-39 passphrase field or CLI input. Passkey/PRF wrapping encrypts the stored root.
- The migration docs now distinguish address-preserving raw-scalar import from move-funds migration into a BIP-39 wallet for Solana derivation.

Local exact-head gates: Bloom CLI 51, daemon 79 + 3, strict Clippy; zero failures. Fresh GitHub CI is running. This is a self-authored remediation record, not independent approval.

0xdewy:
Final dependency-graph cleanup is complete.

- Base remains `master`; exact head is `3c96c496171fbe9c2a09e964c1e1364ab68fd211`.
- Exact Broker #26 pin: `210ed16d7d64a319de84489a27819d8d609028c0`; that Broker revision pins Signer #20 at `18fb12ddea1113478c5ca926f2f711e212ba839e`.
- `bloom wallet import <name>` is the passphrase-free BIP-39 path. `--raw-private-key` is the explicit old-EVM-key migration path. There is no mnemonic-passphrase flag or field.
- Strict local Clippy and the focused CLI/daemon/Machine/VFS suites are green on this head. Exact-head GitHub jobs are the remaining automated merge gate.

The previous apparent divergence came from obsolete temporary review bases and aliases; the active PR graph is now linear and exact-pinned. This is a self-authored disposition note, not independent approval.

josh-richardson:
Superseded by #170, which now contains the full BIP-39 wallet foundation and native Solana stack rebased onto master.

PR #164, opened 2026-08-17
feat(evals): add Harbor Hyperliquid agent task

violet-agent:
## Summary

Reopens the Harbor evaluation work from closed PR #160 on top of the now-current `master` branch.

- adds the Harbor Hyperliquid agent task, Python harness, lifecycle verification, and CI/static-test entrypoint
- enforces exact installed-package, policy, route, custody-capability, and active-lineage preauthorization before temporary authority is opened
- makes cleanup fail closed on orphaned sessions, venue agents, orders, fills, positions, and custody state
- adds the Linux triad developer harness needed to exercise the eval against the authenticated Machine/Broker/Signer topology
- makes Linux developer Petal enrollment emit deterministic, signed active lineage records rather than lineage-free provenance

The live eval remains deliberately fail-closed while the separately versioned Hyperliquid Petal is corrected. The latest bounded attempt reached Broker and exposed an installer-provenance class mismatch; it created no session, key scope, venue agent, order, fill, or position. This PR tracks the Machine/harness side without weakening provenance or custody enforcement.

Supersedes the tracking role of #160; this PR targets `master` directly.

## Verification

- `scripts/test-harbor-evals.sh` — 24 tests passed; Harbor static tests passed
- `cargo fmt --all -- --check`
- `git diff --check origin/master...HEAD`

## Checklist

- [x] Tests added or updated for behavior changes
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A: this PR adds an eval harness and developer-runtime support without changing production architecture contracts; the affected operational flow is documented in `evals/harbor/README.md` and `docs/local-mainnet-integration.md`
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — the harness requires exact package authority, Broker-owned ceremonies, bounded derived-key scope, and deny-by-default restoration
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — N/A: no production agent-facing VFS contract changes; eval operator guidance is contained in `evals/harbor/README.md`

PR #174, opened 2026-08-20
feat(evals): add Harbor Hyperliquid agent task, and make it reach the agent

### LINKED ISSUE #87: [ci]: automated agentic testing
End-game automated testing as part of CI for us probably looks like sets of:
 - prompt <> expected chain state <> expected bloom state

This will allow us to perform somewhat blackbox testing across agents (and various LLMs), while ensuring they can achieve their goals via Bloom. This tests for much more than our existing test stack ever could, namely "can agents successfully use this version of Bloom to achieve what their user wants".

It will also allow us to actually test prompts.

This could potentially just be done on `master` or `on_demand` as part of some / select CI jobs as it will be costly.

### LINKED ISSUE #172: Venue-backed Petal reads serialise and degrade super-linearly under concurrency, starving unrelated reads
## Summary

Petal reads that make a venue call serialise and degrade super-linearly under
concurrency. While any are outstanding they also block unrelated reads that make
no network call at all. An idle venue read costs 0.09s; four concurrent ones cost
97.8s; eight do not finish inside 180s.

## Measurements

Same triad, same wallet, same package hash, taken back to back. `N` is the number
of concurrent readers, each on a distinct path.

| Path | idle | N=2 | N=4 | N=8 |
|---|---|---|---|---|
| local VFS (`wallets/…/addresses.json`, no Petal, no network) | 0.01s | 0.01s | 0.01s | 0.02s |
| Petal **static** (`petals/hyperliquid/README.md`, no venue call) | 0.02s | 0.02s | 0.12s | 0.14s |
| Petal **venue** (`users/…/clearinghouse.json`, enters `http_fetch`) | **0.09s** | **11.90s** | **97.77s** | **>180s, timed out** |

Two further data points:

- A direct `curl` to `api.hyperliquid.xyz/info`, bypassing Bloom entirely, takes
  0.50s. An idle read through Bloom takes 0.09s, so the Petal path is not slow in
  itself — it is faster than the network, because it caches.
- A local VFS read issued while venue reads were still draining took **7.30s**,
  against 0.01s once they cleared. Local VFS reads make no network call.

## What this shows

1. **The defect is confined to the venue-backed read path.** VFS dispatch, the
   mount, and Petal dispatch all stay flat to N=8. Only reads entering
   `http_fetch` degrade.
2. **The degradation is worse than serialisation.** Pure serialisation at N=4
   would predict ~4 × 6.9s ≈ 28s. Measured 97.8s, roughly 3.5× worse, so
   something amplifies beyond a simple queue.
3. **Outstanding venue reads starve unrelated readers**: 0.01s → 7.30s for a
   local VFS read that never touches the network.

## Mechanism: not yet identified

Three candidates were tested and none accounts for it. All sweeps below are
single runs on a freshly restarted triad with a cold cache, same protocol.

| N | unmodified (control) | daemon lock removed | audit fsync removed |
|---|---|---|---|
| 1 | 5.41s | 5.07s | 5.23s |
| 2 | **86.65s** | 44.76s | 66.02s |
| 4 | **150.46s** | 170.88s | 153.11s |

The unmodified control is the worst of the three at N=2, so neither change
helped. The spread between arms is noise, and single runs cannot distinguish
these: any conclusion here needs repeats, ideally an in-process benchmark rather
than `cat` over the mount.

Ruled out as the dominant cost:

- `bloom-daemon/src/lib.rs:982`, a process-wide mutex held across the whole
  round-trip. Removing it changed nothing outside noise.
- `bloom-proto/src/audit.rs:650`, an `fsync` on a signed journal append performed
  synchronously on the async runtime, twice per fetch. Removing it changed
  nothing outside noise.

Both remain real smells worth fixing on their own merits — blocking, fsync'ing
file I/O on a Tokio worker thread in the hot path of every Petal fetch is not
good regardless — but neither explains the collapse.

Still unexamined: per-request wasm component instantiation, the mount adapter's
task and blocking-pool handling, and the reqwest client configuration shared by
Petal fetches.

## Correction

An earlier revision of this issue claimed that removing the daemon lock made
concurrency 2-4x worse and reasoned that the lock was inadvertently protecting
the runtime. That comparison was invalid: the baseline was measured on a warm
aged triad and the experiment on a cold fresh one, so cache state and the code
change moved together. The controlled cold-vs-cold comparison above shows no
such effect. Disregard that claim.

## Impact

This is the binding constraint on the Harbor Hyperliquid eval. The agent explores
`/bloom` with ordinary filesystem idioms — `find`, `ls`, speculative `cat` — which
generate exactly the concurrent venue reads that trigger this, and it exhausted
its entire 900s budget without placing an order. Three further runs were lost to
read and write timeouts. Timeouts in the harness were raised as mitigation, but no
fixed timeout is correct against latency that varies by three orders of magnitude
with load.

## Disproven along the way

- **Hyperliquid-specific**: no. `next.md` makes no Hyperliquid call and was
  equally slow under load.
- **Rate limiting**: no. 453 fetches, all `status=200`, zero 429s, address budget
  `nRequestsUsed: 5` of `10000`.
- **RPC health contention**: no. Probes live in `bloom-rpc` and never touch this
  path; the failing chains in that environment were coincidental.
- **Degrades with process age**: no. That was load correlation. Fresh triads
  looked fast because they were idle, and restarts appeared to help only because
  they cleared load.
- **A constant multi-second overhead per read**: partly. A *cached* read is
  0.09s, but an uncached fetch costs ~5s against 0.50s for a direct curl, and
  that gap is the audit path rather than the network.
- **`http_audit_lock` is the bottleneck**: no. Removing it changed nothing
  outside measurement noise.
- **The audit `fsync` is the bottleneck**: no. Same result.

## Reproduce

Concurrent `cat` of distinct venue-backed routes under
`petals/hyperliquid/mainnet/users/<addr>/` at N = 1, 2, 4, 8, timing each. Compare
against the same sweep over static Petal paths and local VFS paths, which stay
flat.

## Suggested direction

The next step is measurement, not a fix. `cat` over an NFS mount adds its own
variance, and single runs at N=2 ranged 44-87s across arms that should have been
equivalent. A repeatable in-process benchmark that drives the Petal read path
directly would make the remaining candidates separable.

Getting the signed journal append off the async runtime is still worth doing on
its own merits, but the measurements above say it will not fix this.

GavinPacini:
Supersedes #164, which is based on a fork and cannot receive further commits from this repository.

## Summary

Adds Bloom's first end-to-end Harbor agent evaluation. A Codex or Claude agent uses a live Bloom triad and the Hyperliquid Petal to create bounded signing authority, place a small post-only BTC order on Hyperliquid mainnet, prove that it rested, cancel it, and leave no residual venue or Bloom session state.

Getting that black-box flow working also exposed and fixed two underlying stack problems: venue-backed Petal reads became progressively slower as the signed audit journal grew, and a derived Petal key did not yet receive reusable authority spanning the bounded set of routes needed for a workflow.

### Harbor Hyperliquid evaluation

- Adds the Harbor task, container environment, verifier, fixtures, CI/static checks, documentation, and a Python harness for running the evaluation against a local Bloom triad.
- Preauthorizes the exact package, account, route set, operation classes, key lifetime, and operation/signature limits needed by the task.
- Drives agent creation/adoption, WebAuthn ceremonies, key derivation, reusable approval, session creation, leverage, order placement, live venue-state verification, cancellation, and cleanup.
- Fails closed when preauthorization, live-state verification, session cleanup, or host cleanup is incomplete.
- Makes repeated local runs reliable through stable agent identity, monotonic WebAuthn counters, bounded retries/timeouts, and improved diagnostics.

### Reusable multi-route Petal authority

- Derives one reusable sealed approval for a Petal key from the key request's complete allowed-route set, rather than requiring a separate approval for every action.
- Binds that authority to one package, Petal key, sorted and deduplicated routes, operation classes, lifetime, operation/signature budgets, policy version, and revocation epochs.
- Keeps the key pending until the reusable approval is active, then reuses that approval for matching typed requests without another user ceremony.
- Preserves deterministic active provenance and rejects stale, mismatched, expired, exhausted, or revoked authority.

### Venue-read and audit performance

- Fully authenticates the signed audit journal when it is opened at daemon startup.
- Replaces full-history journal verification before every append with constant-time file identity and metadata checks plus incremental in-memory state.
- Locks and rechecks the journal during append so external growth, truncation, replacement, and concurrent mutation still fail closed.
- Removes the super-linear audit cost paid by each Petal HTTP intent/result, preventing concurrent venue-backed reads from starving unrelated VFS work.

### Developer and runtime reliability

- Extends the triad development harness and Linux service/socket handling needed by Harbor.
- Reports mount ownership as the serving process so unprivileged action sinks remain writable.
- Uses the Bloom wallet identifier for session routes while retaining the on-chain address for venue account reads.
- Improves host-side Petal error logging without exposing additional diagnostics to guests.

## Verification

- `scripts/test-harbor-evals.sh` passes all 32 harness tests.
- Harbor static checks pass.
- A live Codex run completed with reward `1.0`: it placed and observed a resting BTC post-only order, cancelled that exact order, stopped the Bloom session, and finished with no open orders or positions.

## Checklist

- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — see `docs/architecture/Agent-native Documentation.md`

Closes #87
Closes #172


evals/harbor/harness/hyperliquid_order_cancel.py
```diff
@@ -0,0 +1,1004 @@
...
+        )
+
+    def cleanup(self) -> None:
+        if self.session_base is None or self.session_id is None:
+            return
+        failures: list[str] = []
+        status_data = self._read_session_status()
+        if status_data is None:
+            try:
+                if self._pending_petal_key_ceremony() is not None:
```
0xdewy: Cleanup can return successfully with the second owner ceremony still open. If provisioning has completed key derivation and staged `approve_agent`, but the durable session has not appeared, `status_data` is `None` and `_pending_petal_key_ceremony()` is also `None`; `_require_empty_wallet()` passes because the venue agent has not been registered yet. The matching `_pending_agent_approval_ceremony()` is never checked here, so an awaiting owner-approval ceremony is left behind despite the fail-closed cleanup contract. Please check both ceremony projections in this branch (and fail cleanup if either exact match remains).

evals/harbor/harness/hyperliquid_order_cancel.py
```diff
@@ -0,0 +1,1004 @@
...
+        self.session_created = True
+        stopped = isinstance(status_data, dict) and status_data.get("stopped") is True
+        if not stopped:
+            cancel = self._write_route(
+                self.session_base / "cancel_all", b"host-cleanup", 30
+            )
+            if cancel.returncode != 0:
+                failures.append("session cancel_all failed")
+
+        try:
```
0xdewy, line 1010: These cleanup writes need bounded completion polling. The provisioning path correctly notes that a zero mounted-write exit code means only “accepted for dispatch,” but cleanup immediately reads `open_orders`, positions, and later `status.json` after `cancel_all`, `close_all`, and `stop`. On the real asynchronous mount those reads can observe the pre-write state, record a failure, skip `stop` because `failures` is non-empty, and leave the session active. The unit test masks this by making each mocked read advance synchronously. Please poll the relevant postcondition after each accepted write (no orders, no positions, then `stopped == true`) before deciding cleanup failed.

josh-richardson:
Addressed the review findings in 6e68483 and 8cac0de.

- The Harbor verifier now obtains independent Hyperliquid orderStatus evidence using the host-generated CLOID and the venue-assigned OID. It requires a canceled BTC ALO order and binds the venue OID, price, and original size to the report. Added HTTP-boundary and adversarial tests.
- Nonempty audit journals must now end with a newline delimiter. A truncated signed tail degrades the journal and blocks further mutation; added a restart regression test.
- Updated the runtime and Broker API pins to the fixed zero-clock and multi-route authority revisions.

The complete PR CI suite is green, including Clippy, unit tests, local integration, live-network e2e, Harbor checks, and the locked release build.

PR #135, opened 2026-07-29
Implement triad process architecture

josh-richardson:
## What changed

- implements the Machine/Broker/Signer protocol and process boundaries
- adds ceremony, custody, policy-update, revocation, trusted-time, audit-checkpoint, and release packaging behavior
- extracts Broker and Signer into independently versioned public repositories:
  - https://github.com/bloom-directory/bloom-broker
  - https://github.com/bloom-directory/bloom-signer
- adds Linux principal isolation and fail-closed release/installer gates
- implements the root-requiring macOS Unix-principal profile with hidden service accounts, LaunchDaemons, authenticated Unix RPC, UID-scoped packet filtering, transactional enrollment/upgrades/config and identity rotation, recoverable uninstall, and retain/restore custody
- preserves the rootless macOS architecture as an explicit future profile; no App-Group production claim is enabled

## Why

The triad architecture moves authorization and custody out of the Machine process, removes direct/hash-only signing paths, and makes process, protocol, replay, audit, packaging, and operating-system isolation boundaries testable.

## Validation

Local validation passed after merging the current default branch:

- `cargo fmt --all -- --check`
- `cargo check --workspace --all-targets --locked`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`
- shell syntax and ShellCheck for the macOS installer and W0 harness

The non-privileged Linux W0 and macOS listener-retry CI lanes pass. The production `macos-unix-principals` claim remains deliberately disabled until the privileged destructive W0 workflow runs after this workflow is available on the default branch and its conformance evidence is recorded.

## Checklist

- [x] Tests added or updated for behavior changes
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — see `docs/architecture/Sealed Approvals.md`
- [x] Agent Documentation updated — N/A: this changes triad custody/packaging and host isolation, not VFS agent-facing behavior or Petal interfaces

PR #175, opened 2026-08-21
feat(triad): add Linux enrollment and packaging

0xdewy:
## Why this is necessary

Master can already *run* the triad on Linux but cannot *install* it there.

`packaging/triad/linux/` on master carries the systemd units, socket
activation, sysusers, tmpfiles, and chrony config, and `install-linux.sh` has
been present since the triad process architecture landed (#135). What is
missing is the step that produces the per-login identities those units
consume. There is no Linux equivalent of the macOS enrollment path, so the
service files on master reference identity material that nothing generates.

This PR supplies that half: enrollment generation, a journaled installer with
crash recovery, and the release-bundle and CI wiring that keeps both honest.
Without it Linux packaging on master is inert.

## What it adds

- `run_linux()` in `crates/bloom/src/triad_enrollment.rs` — root-only
  generation of per-login Linux triad identities against FHS paths, alongside
  the existing macOS path (`run` is renamed `run_macos` so the two platforms
  are named symmetrically rather than one being implicit).
- `packaging/triad/release/enroll-linux.sh` — the enrollment generator.
- `packaging/triad/release/install-linux.sh` — reworked into a journaled,
  atomic install/upgrade/uninstall flow that recovers from interruption.
- `packaging/triad/linux/config/` — broker, signer, edge-manifest, and
  unsigned provenance-catalog templates the units consume.
- `packaging/triad/linux/tests/` — fresh-install and SIGKILL crash-recovery
  fixtures, plus an isolated CI job to run them.
- `packaging/triad/release/check-docker-container-safety.sh` and a
  release-gate check, so the container used for verification cannot silently
  acquire privilege.

## Scope

Linux enrollment, packaging, and release plumbing only. It touches no BIP-39,
wallet-account, machine-client, or Solana code.

**This PR previously targeted `feat/bip39-hd-wallets`.** That stacking was
historical, not functional — nothing here depends on BIP-39. It has been
retargeted to `master` and rebased onto it, so it can be reviewed and merged
independently of the Signer/Broker/Machine BIP-39 stack rather than queueing
behind three repositories of unrelated review.

## Relationship to #174

#174 landed the developer-side platform policy on master, factoring the root
and platform checks into `validate_developer_caller(uid, os)` and relaxing
them from macOS-only to Linux-or-macOS. This PR originally made the same
relaxation inline. That duplicate has been dropped during the rebase in
favour of master's helper; the two call sites now use
`validate_developer_caller` directly.

What remains here is the part #174 did not cover: the root-only **production**
Linux enrollment path (`run_linux`), which is separate from the developer
harness.

## Verification

- `cargo test -p bloom --bin bloom triad_enrollment` — 5 passed, including
  master's `developer_enrollment_platform_policy_accepts_non_root_linux_and_macos_only`
  and this PR's `linux_material_uses_fhs_paths_and_kernel_enforced_containment`,
  which together confirm the deduplicated guard and the new Linux path coexist.
- `cargo test -p bloom-it --test triad_release linux` — 2 passed
  (`linux_installer_upgrade_rotation_and_confirmed_uninstall_are_staged_safely`,
  `triad_developer_launcher_supports_linux_without_weakening_root_boundary`).
- `cargo fmt --all -- --check` and
  `cargo clippy --workspace --all-targets -- -D warnings` both clean.

## Safety

The tracked Linux crash-recovery test uses an ordinary Docker container with a
Bash entrypoint and read-only mounts. It does not use privileged mode, host
cgroups, `/sbin/init`, host namespaces, or host TTYs, and the release gate now
asserts that. Full production-binary smoke testing remains a VM/CI-only
follow-up and should not be run on an active desktop.

## Checklist

- [x] Tests added or updated for behavior changes — `triad_enrollment` gains
      `linux_material_uses_fhs_paths_and_kernel_enforced_containment`,
      `triad_release` gains
      `linux_installer_upgrade_rotation_and_confirmed_uninstall_are_staged_safely`
      and `triad_developer_launcher_supports_linux_without_weakening_root_boundary`,
      plus the fresh-install and SIGKILL crash-recovery fixtures under
      `packaging/triad/linux/tests/`
      behavior changed — N/A, no architectural contract changes; the Linux
      install and enrollment procedures are documented in
      `packaging/triad/linux/README.md` and `packaging/triad/release/README.md`,
      both updated here
      bytes) — N/A, this PR performs no signing and touches no approval path;
      it generates transport identities and installs service units
      and affected Petal READMEs) — N/A, no VFS surface, agent-facing
      behavior, or Petal contract changes


josh-richardson:
Superseded by #165, which already implements the Linux enrollment and packaging path and has since advanced the installer architecture substantially. The crash-recovery ideas here remain useful, but should be ported onto #165 rather than merging this older implementation.

PR #184, opened 2026-08-25
perf(evals): use prebuilt Harbor agent image

GavinPacini:
## Summary

- pin the Hyperliquid Harbor task to the published multi-architecture `bloom-eval-agent-base` manifest
- remove the redundant task-local Ubuntu Dockerfile
- pre-pull the immutable agent image before authority provisioning and batch exact Bloom reads sequentially
- tolerate transient malformed JSON snapshots while NFS projections are replaced
- retire durable Petal ceremony projections whose Broker sessions were lost on restart, incorporating and superseding #186
- consume immutable action-and-CLOID-scoped Petal receipts so order acknowledgement cannot be overwritten by a later cancel response

## Dependencies

This PR is the Bloom/Machine portion of the ceremony-stability stack and should be tested with:

- [bloom-signer#13](https://github.com/bloom-directory/bloom-signer/pull/13) — merged as `161bc8e6b5c435128faa7f15b23a501ec99f9f05`; durably terminalizes rejected proofs
- [bloom-broker#14](https://github.com/bloom-directory/bloom-broker/pull/14) — merged as `c88876addd6bb4dd05d44943e371da8a7c07a8c2`; preserves, reconciles, and releases rejected proof state

## Benchmark

Measured on Apple Silicon with Docker 29.4.0 using the exact Harbor 0.21 provisioning commands in fresh disposable containers. The reusable-base measurements include container startup and Harbor-equivalent CLI detection.

| Agent | Current image | Prebuilt base | Median savings |
| --- | ---: | ---: | ---: |
| Codex | 59.92s, 86.17s, 85.47s | 0.43s, 1.26s, 0.28s, 0.28s, 0.29s | 85.18s (99.7%, ~295x) |
| Claude | 90.25s, 70.30s, 68.33s | 0.91s, 0.28s, 0.28s | 70.02s (99.6%, ~251x) |

Bare container startup measured 0.24–0.29s, so provisioning from the base is effectively reduced to Docker startup. The current path repeats `apt-get`, Node installation, and agent installation inside every ephemeral trial, so Docker build caching cannot retain that work.

The tradeoff is a larger first pull: Docker reports 1.07GB unpacked for the reusable image versus 150MB for the current task image. That cost is paid once per eval host; the 70–85 second saving recurs per trial.

The pinned manifest was produced by the successful [eval-images publication run](https://github.com/bloom-directory/eval-images/actions/runs/32868380982). Anonymous registry inspection confirms the digest is public and contains both `linux/amd64` and `linux/arm64` manifests.

## Validation

- `uv run --isolated --no-project --python 3.12 bash scripts/test-harbor-evals.sh`
  - 39 harness/verifier tests passed
  - Harbor 0.21 job-plan API validation passed
- `cargo test -p bloom-daemon --lib` — 79 passed for the incorporated restart-reconciliation change
- sibling image arbitrary-non-root smoke test passed for Codex 0.149.1 and Claude Code 2.1.245
- anonymous inspection of the pinned GHCR manifest passed for amd64 and arm64
- integrated mainnet eval with the linked Broker/Signer branches retired five stale ceremony projections on startup and converged with no stuck ceremonies, open orders, or positions
- full correlated-receipt mainnet eval with the linked Broker/Signer branches and local Hyperliquid Petal commit `ed9d770` passed at reward `1.0`, with zero errors/retries; immutable order and cancel receipts independently proved the resting-order lifecycle, cleanup stopped the session, and no positions or open orders remained
- timed full run: `182.03s`, versus `321.66s` for the prior integrated run — `139.63s` saved (`43.4%`). This end-to-end comparison includes the combined image, batching, transient-read, restart-reconciliation, and correlated-receipt changes; the isolated image benchmark above measures image savings specifically.
- repeated full eval after fast-forwarding Signer #13 to `2175f05f4814dedb40b00daa4ce6683a90d0c39d`: reward `1.0`, zero errors/retries, `220.99s`; the immutable order/cancel receipts passed, the session stopped, no positions or open orders remained, and the wallet was restored to deny-by-default
- `git diff --check`

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A; eval documentation is updated and restart reconciliation is an implementation fix
- [x] Sealed Approval invariants respected — stale sessions retire fail-closed; image and batching changes do not widen authority
- [x] Agent Documentation updated — N/A; no production agent-facing filesystem contract changed


0xdewy (commented):
Independently verified: ran `scripts/test-harbor-evals.sh` locally (38 tests + Harbor 0.21.0 job-plan validation pass), inspected the pinned digest via anonymous GHCR (valid 64-hex OCI index with `linux/amd64` + `linux/arm64`), confirmed the image's final layer installs `bash ca-certificates coreutils curl git jq openssh-client procps python3 ripgrep` plus codex 0.149.1 / claude-code 2.1.245 (a superset of what `tests/test.sh` and the stdlib-only `verify_result.py` need), and confirmed `preflight()` → `_pull_eval_image()` runs before `provision()` creates the 30-minute session (`harness/core.py:185` vs `:197`). No bugs found. Three non-blocking notes:

**1. Benchmark semantics change not in the PR summary.** `instruction.md:6-9` now allows batching multiple exact-path reads into one shell invocation ("one command at a time" before). This is a deliberate turn-economics change (commit ebea010, "batch discovery") but the PR description only covers the image swap, so eval results before/after this PR are not directly comparable. Worth an explicit line in the PR body (or wherever results are tracked) so consumers of the numbers don't miss it.

**2. No local fallback if GHCR is unreachable.** With the task Dockerfile deleted, Harbor's `should_use_prebuilt_docker_image` (`harbor/environments/definition.py:26`) returns True whenever `docker_image` is set and no Dockerfile exists — so even `force_build` can't fall back to a local build. That looks deliberate (the digest is authoritative), but it means a host without outbound GHCR access cannot run the eval at all, and a cold pull of the ~1.07GB image that exceeds the 600s `EVAL_IMAGE_PULL_TIMEOUT_SECONDS` fails the run. Both failure modes are fail-closed before any authority exists and are mitigable by pre-pulling, so this is an operational note, not a defect.

**3. `build_timeout_sec` is not dead — please don't remove it as cleanup.** Despite the name and the deleted Dockerfile, it still bounds the entire `environment.start()` phase: `harbor/trial/trial.py:1215-1224` wraps `start()` in `asyncio.wait_for(timeout=...)` derived from `task.config.environment.build_timeout_sec` (`trial.py:1136-1139`), and with a prebuilt image `start()` still does compose file writes, image-OS validation (`docker.py:893-900`), and `compose up` inside that budget. Keeping it at 600 also matches the harness pull timeout.

PR #186, opened 2026-08-25
fix(petals): retire stale ceremony projections on restart

### LINKED ISSUE #173: Triad restart silently invalidates pending ceremony projections, which keep advertising dead tokens
## Summary

Broker ceremony sessions live only in Broker memory. Their owner-visible
projections live on disk under the Machine's cache. Restarting the triad clears
the former and leaves the latter, so every pending ceremony projection becomes a
durable advertisement for a session that no longer exists. Nothing reconciles
them, and nothing tells a consumer that a restart is why its ceremony stopped
working.

## What a consumer sees

A projection under `state/machine/cache/petal-key-requests/` or
`petal-signing-requests/` still reads:

```json
{ "status": "awaiting_owner_approval", "ceremony_url": "http://localhost:18734/ceremony/…" }
```

Completing that URL fails, and how it fails depends on timing:

- `HTTP 403` with an empty body, from `authorize_browser_token`, when the token
  resolves to nothing
- `HTTP 400 {"code":"CEREMONY_REPLAY","retry":"never","message":"ceremony is absent, expired, or already consumed"}`

Neither mentions a restart, and the projection continues to claim the ceremony is
awaiting the owner. Diagnosing this took several hours and a temporary
`eprintln` in Broker's `authorize_browser_token`, because Broker has no tracing
at all: the only line that identified the cause was one we added, reporting
`token not resolvable (rate_limited=false, live_sessions=12)`.

## Impact

Three Harbor eval runs failed against stale projections after a restart. One of
them burned three WebAuthn counters, because the consumer retried an error the
Broker had already marked `retry: "never"`.

Worse, a stale projection is indistinguishable from a live one, so a consumer
scanning for work will pick it up and act on it. Pruning the two cache
directories by hand is currently the only remedy, and it is not documented
anywhere as part of restarting.

## Suggested direction

Roughly in order of value:

1. **Invalidate on startup.** If Broker's session store is empty, the Machine's
   pending ceremony projections cannot be live. Mark them terminal, or delete
   them, rather than leaving them to advertise dead tokens.
2. **Distinguish the failure.** `CEREMONY_REPLAY` is accurate but conflates
   "already consumed" with "the process that held this restarted". A consumer
   cannot tell whether to give up or to restage.
3. **Log it.** Broker currently emits nothing on any of its ceremony rejection
   paths. Even one line naming which precondition failed would have made this
   immediate. See also the diagnosability point raised in the Petal signing
   denial work.

## Reproduce

Stage a ceremony, restart the triad without pruning
`state/machine/cache/petal-*-requests/`, then complete the ceremony URL still
advertised by its projection.

violet-agent:
## Summary

Fixes #173 by invalidating durable Petal ceremony projections that point to Broker sessions lost on restart. Stale URLs are cleared, completed approvals remain reconcilable, and key ceremonies can restage rather than becoming permanently blocked.

## Verification

- `cargo test -p bloom-daemon --lib` (79 passed)

## Checklist

- [x] Architecture docs updated if contracts or behavior changed — N/A: restart reconciliation implementation
- [x] Sealed Approval invariants respected — stale sessions are retired fail-closed
- [x] Agent Documentation updated — N/A: no agent workflow change

### DIFF
--- a/crates/bloom-daemon/src/lib.rs
+++ b/crates/bloom-daemon/src/lib.rs
@@ -236,6 +236,19 @@ struct DaemonPetalHost {
 const PETAL_KEY_STATE_SCHEMA: &str = "bloom.machine.petal-key-request.v2";
 const PETAL_KEY_INPUT_CLASS: &str = "petal-key-scope-v2";
 
+/// Status written over any Petal ceremony projection that still advertised an
+/// owner ceremony when this process started.
+///
+/// Broker holds ceremony sessions in memory only, so a URL staged by an
+/// earlier run resolves to nothing once the triad restarts: completing it
+/// fails with an unexplained `403` or `CEREMONY_REPLAY`. A durable projection
+/// that keeps claiming `awaiting_user` / `awaiting_owner_approval` is
+/// indistinguishable from a live one, so a consumer scanning for work picks it
+/// up and burns real authenticator counters on it. Machine cannot vouch for a
+/// ceremony it staged before its own start, so it stops advertising it and
+/// names the reason. Both projections restage on the next Petal call.
+const PETAL_CEREMONY_INVALIDATED_STATUS: &str = "ceremony_unavailable_after_restart";
+
 /// Machine-owned public reconciliation record. The ceremony URL is retained
 /// here for an owner-readable status projection, but is never returned across
 /// the Petal host boundary.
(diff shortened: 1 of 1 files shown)


GavinPacini:
Superseded by #184. The restart-reconciliation commits have been incorporated into that branch so the eval performance work and the full Bloom/Broker/Signer stability stack can be tested together. Thank you, Violet.

PR #192, opened 2026-08-27
Make Harbor evals reproducible and recoverable

### LINKED ISSUE #187: Require explicit Bloom mount selection for Harbor evals
## Problem

The Hyperliquid Harbor harness currently defaults an omitted `BLOOM_EVAL_BLOOM_MOUNT` to `/bloom`. When a developer triad uses a custom `--mount` path, the omission is discovered only during full-run preflight as:

```
Bloom Harbor eval: Bloom is not mounted at /bloom
```

This does not explain that mount selection was implicit or how to get the correct value. More importantly, if `/bloom` exists but belongs to another local instance, the harness could select the wrong VFS.

## Proposed solution

Make mount selection explicit and fail closed:

1. Require `BLOOM_EVAL_BLOOM_MOUNT` for every full eval; remove the implicit `/bloom` fallback.
2. Fail before provisioning with an error that names the missing variable.
3. Keep `--preauthorization-only` mount-independent.
4. Make `scripts/triad-dev-launch.sh` export its exact `--mount` value as `BLOOM_EVAL_BLOOM_MOUNT` in generated `triad.env`.
5. Show the variable explicitly in every full-eval README example, using `/bloom` for a conventional installation.
6. Retain validation that the configured path is a real mount with the expected Bloom projections.

Avoid automatic mount discovery: multiple Bloom mounts may exist, so guessing is less safe than requiring an explicit path.

## Acceptance criteria

- A full run without `BLOOM_EVAL_BLOOM_MOUNT` fails immediately and names the variable.
- The failure occurs before provisioning or external work.
- Generated `triad.env` exports the launcher’s exact custom mount.
- `--preauthorization-only` still works without a mount.
- Tests cover omitted, invalid, and valid custom mount paths.
- Documentation no longer calls the mount optional for full evals.

### LINKED ISSUE #190: Make Harbor evals reproducible from a context-free operator runbook
## Problem

The Hyperliquid Harbor eval documentation explains the trial and its safety boundaries well once an operator already has a prepared environment. It is not currently sufficient for a new agent or operator with no prior run context to reproduce a full, timed eval safely.

The missing operational state and setup steps include:

- how to start or verify the Bloom triad, establish the Machine mount, and install the intended Hyperliquid Petal;
- how to discover the dedicated wallet identifiers without embedding them in the repository;
- how to derive the owner-record, Petal-store, and provenance-catalog paths from the configured triad home;
- how to verify the exact Bloom, Broker, Signer, Petal contract, and Hyperliquid Petal revisions used by the running environment;
- how to recover and persist the next valid WebAuthn signature counter across successful and interrupted runs;
- how to apply and restore the package-only wallet policy without relying on temporary files or manual repetition;
- how to honor `BLOOM_EVAL_BLOOM_MOUNT` consistently, since the current manual policy examples still use literal `/bloom` paths; and
- how to report total elapsed time in a consistent machine-readable result.

This creates two risks: a new operator cannot reproduce a prior result from repository documentation alone, and an interrupted mainnet run can leave ambiguous counter or wallet-policy state that requires undocumented recovery knowledge.

No credentials, wallet values, user names, or machine-specific filesystem paths should be committed as part of the solution.

## Proposed solution

Add a single operator-facing eval command and supporting runbook that owns setup discovery, validation, execution, and recovery.

1. Introduce a gitignored, mode-`0600` eval-state file with a documented schema. Store only the dedicated wallet identifiers, paths to secret material, triad root, selected model, and next candidate authenticator counter. Never store the seed or model credential itself.
2. Add an initialization or discovery command that derives triad paths, validates the Machine mount, locates the installed Petal owner record and provenance catalog, and confirms the wallet is dedicated and empty.
3. Record and validate component lineage before provisioning: Bloom, Broker, Signer, Petal contract/tooling, installed Hyperliquid Petal package hash, and source revision where available. Emit this lineage in the run result.
4. Move package-only policy activation and deny-by-default restoration into an automated host-side lifecycle. Persist the original canonical policy in protected state before mutation and restore it from an unconditional cleanup path after success, failure, or interruption.
5. Persist ceremony-counter progress after each confirmed completion using an atomic write. On restart, reconcile pending ceremonies and documented Broker state before selecting the next counter; never infer advancement solely from process exit.
6. Make every mounted path derive from `BLOOM_EVAL_BLOOM_MOUNT`; remove literal `/bloom` assumptions from executable examples.
7. Add monotonic timing around preflight, image pull, authority provisioning, Harbor execution, cleanup, and total wall time. Write a machine-readable summary alongside the Harbor job result.
8. Keep `--preauthorization-only` as the read-only first step and add a dry-run/status mode that reports missing or stale prerequisites without creating authority or making external calls.

## Acceptance criteria

- A new agent can follow repository documentation and run the eval without access to prior chat logs or shell history.
- The workflow requires an explicit mainnet acknowledgement and fails closed before authority creation when setup or lineage is incomplete.
- No personally identifiable information, wallet value, authenticator seed, or model credential is committed or printed.
- An interrupted run can be resumed or cleaned up deterministically, including policy restoration and ceremony-counter reconciliation.
- Non-default Machine mount paths work throughout the documented and automated workflow.
- Each run reports component revisions, package hash, Harbor/model configuration, reward, errors/retries, phase timings, total elapsed time, and final open-order/position/session/policy state.
- Static tests cover state-file permissions, atomic counter updates, interruption recovery, path overrides, policy restoration, and redaction.

GavinPacini:
Closes #190
Closes #187

## Problem

The Hyperliquid Harbor eval could only be run safely by an operator carrying prior shell history and undocumented state. Wallet policy staging/restoration, authenticator counters, triad paths, component lineage, and elapsed timing were manual, making interrupted runs difficult to reconcile and cold-start reproduction unreliable.

Repeated cold-start testing also found that ignored Petal packaging artifacts could become stale after route compilation. Full environment-driven runs also implicitly fell back to `/bloom` when `BLOOM_EVAL_BLOOM_MOUNT` was omitted, which could select the wrong local Bloom instance.

## Solution

- add a single `init` / `status` / `run` / `recover` operator command using isolated `uv` and Python 3.12; `uv` is the only Python prerequisite
- define the protected, gitignored default handoff at `evals/harbor/operator-state.json`, with self-describing JSON, field guidance, safe-field and secret-path declarations, and explicit resume instructions
- discover and persist the exact triad, mount, wallet, package, source, binary, and recovery bindings required to resume safely without shell history
- revalidate Bloom, Broker, Signer, Petal contract, Hyperliquid Petal, and triad binary lineage before authority creation
- stage only the installed package in wallet policy and restore the original deny-by-default bytes from a mode-`0600` backup in unconditional cleanup/recovery paths
- atomically persist the next safe authenticator counter after every Broker ceremony attempt, including ambiguous transport failures
- reject unresolved matching owner ceremonies and prevent initialization from replacing pending recovery state
- refresh ignored integration-Petal package artifacts before developer provenance enrollment, preventing source/artifact drift across repeated launches
- require both host orchestration and the evaluated agent to use the real NFS mount through ordinary filesystem operations; `bloom vfs`, Machine RPC, and alternate transports are explicitly prohibited and mount failures fail the eval environment closed
- require explicit `BLOOM_EVAL_BLOOM_MOUNT` selection before any full-run external work, retain mount-independent preauthorization, and export the launcher's exact custom mount in generated `triad.env`
- gitignore the transient single-operator lock so the workflow cannot invalidate its own clean-lineage check
- emit mode-`0600` machine-readable summaries with outcome, reward/errors/retries, lineage, timings, and cleanup postconditions while omitting secrets and wallet identifiers
- document cold launch, initialization, status, run, recovery, state schema, explicit mount selection, and repeat-agent resume

## Validation

- `scripts/test-harbor-evals.sh` (54 tests plus Harbor 0.21 API smoke), including omitted/invalid/valid custom mount coverage, a regression check that rejects fallback symbols in the Python harness, and the mounted-only instruction
- `cargo test -p bloom-it triad_developer_launcher_exports_its_machine_connection`
- focused developer-enrollment tests and `cargo check -p bloom --features triad-dev-harness`
- isolated uv/Python 3.12 launcher and standard-library `json.load` ingestion checks
- Ruff, shell syntax, gitignore, and `git diff --check`
- multiple fresh agents, without prior wallet/path/counter context, independently discovered and operated the default handoff
- before the mounted-only follow-up, one full bounded Hyperliquid mainnet Harbor eval validated the operator lifecycle and handoff: reward `1.0`, 0 errored/cancelled/retried trials, zero final orders/positions, stopped session, reconciled counter, restored deny-by-default policy, and recovery clear
- the mounted-only follow-up is statically validated and deliberately has not reused that earlier live result as evidence for its removed transport behavior

## Checklist

- [x] Operator documentation updated
- [x] Sealed Approval invariants preserved; policy and session authority remain owner-approved, bounded, recoverable, and fail closed


evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,898 @@
...
+        Path(paths["signer_repo"]),
+        Path(paths["hyperliquid_repo"]),
+    )
+    if actual != state["lineage"]:
+        raise EvalError("source lineage changed; re-run init before creating authority")
+    for name, expected in state["binaries"].items():
+        if sha256_file(Path(expected["path"])) != expected["sha256"]:
+            raise EvalError(
+                f"{name} binary changed; re-run init before creating authority"
+            )
```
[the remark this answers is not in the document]
GavinPacini: Fixed in 7d13d5ca. Binary hashing now converts OSError into a path-redacted EvalError, with a regression test for an unavailable binary.

evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,898 @@
...
+            ) from last_read_error
+        raise EvalError(
+            "completed policy could not be committed by bounded byte-identical replay"
+        )
+
+    def apply(self, target: bytes) -> None:
+        expected = json.loads(target)
+        expected_bytes = canonical_json(expected)
+        if target.rstrip(b"\n") != expected_bytes:
+            raise EvalError("policy update target is not canonical JSON")
```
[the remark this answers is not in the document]
GavinPacini: Fixed in 7d13d5ca. Policy payload decoding now maps JSON/Unicode/type failures to EvalError and rejects non-object or unsupported policy shapes before any mutation, with regression coverage.

evals/harbor/harness/hyperliquid_order_cancel.py
```diff
@@ -96,7 +102,8 @@ def __init__(self, repo_root: Path, environ: dict[str, str] | None = None) -> No
             "BLOOM_EVAL_PROVENANCE_CATALOG", ""
         )
         self.provenance_catalog = Path(self.provenance_catalog_value)
-        self.bloom_mount = Path(self.env.get("BLOOM_EVAL_BLOOM_MOUNT", "/bloom"))
+        self.bloom_mount_value = self.env.get("BLOOM_EVAL_BLOOM_MOUNT", "")
+        self.bloom_mount = Path(self.bloom_mount_value)
```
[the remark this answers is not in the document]
GavinPacini: Fixed in 7d13d5ca. The mount environment value is stripped on ingestion, so whitespace-only input follows the required-variable failure path; regression coverage verifies it fails before external work.

josh-richardson (commented):
P1 recovery-path findings

evals/harbor/harness/hyperliquid_order_cancel.py
```diff
@@ -907,6 +939,8 @@ def provision(self, agent_name: str) -> EvalRunContext:
             output += (completed.stdout + completed.stderr).decode(errors="replace")
             counter += 1
             self.next_sign_count = counter
+            if self.counter_committed is not None:
```
josh-richardson: [P1] Persist the next counter when ceremony completion has an ambiguous transport failure. `counter_committed` is only called after `subprocess.run` returns, so a 45-second timeout after the Broker accepted the WebAuthn assertion takes the exception path with the state file still pointing at the consumed counter. A subsequent `recover`/`run` will reuse it, which is exactly the unsafe ambiguity this workflow is intended to prevent. Advance and atomically persist `counter + 1` before re-raising `TimeoutExpired`/other transport errors (as `PolicyLifecycle.apply` already does), and add a timeout regression test.

evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,909 @@
...
+
+
+def run_or_recover(
+    args: argparse.Namespace, repo_root: Path, recover_only: bool
+) -> None:
+    store = StateStore(args.state)
+    state = store.read()
+    if args.ack != MAINNET_ACK:
+        raise EvalError(f"explicit acknowledgement must equal {MAINNET_ACK}")
+    validate_lineage(state, repo_root)
```
josh-richardson: [P1] Do not gate restoration of a pending wallet policy on full source/binary lineage. This call also runs for `recover`, so after an interrupted activation leaves the package allow-listed, any branch switch, source edit, or binary rebuild makes the supported recovery command exit before reading and restoring the protected backup. `init` also refuses while recovery is pending, leaving the wallet stuck outside deny-by-default until the operator reconstructs the old checkout/binaries or edits state manually. Keep the full lineage gate for `run`/authority creation, but let `recover` restore the exact digest-bound backup using only the prerequisites needed for that fail-closed operation.

evals/harbor/harness/hyperliquid_order_cancel.py
GavinPacini: Addressed in 55632227. The harness now durably persists the next unused WebAuthn counter before invoking the ceremony driver, in both session provisioning and policy-update completion. If persistence fails, no assertion is sent; if the driver times out or is interrupted after Broker accepts it, the attempted counter cannot be reused. The new timeout regression verifies that the persisted counter is already advanced while subprocess.run is failing, and the focused harness suite passes.

evals/harbor/harness/operator.py
GavinPacini: Addressed in 55632227. Recovery no longer runs the full source/binary lineage check, so a branch change or rebuilt binary cannot block restoration; normal eval runs still require complete lineage validation. The recovery path remains constrained to the protected backup: activation binds its SHA-256 digest before policy mutation, subsequent recovery markers preserve that binding, and restore requires a regular mode-0600 file containing canonical bytes for the exact expected deny-by-default policy and matching the bound digest. Regression tests cover recovery with stale lineage plus rejection of digest mismatches and non-deny-policy backups; the focused harness suite passes.

evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,954 @@
...
+            expected = json.loads(target)
+        except (json.JSONDecodeError, UnicodeDecodeError, TypeError) as error:
+            raise EvalError("policy update target is invalid JSON") from error
+        if not isinstance(expected, dict) or set(expected) != POLICY_KEYS:
+            raise EvalError("policy update target has an unsupported shape")
+        expected_bytes = canonical_json(expected)
+        if target.rstrip(b"\n") != expected_bytes:
+            raise EvalError("policy update target is not canonical JSON")
+        _, previous_bytes = self._read_policy()
+        if previous_bytes == expected_bytes:
```
josh-richardson, line 688: [P1] Recovery can report success while the staged allow-policy ceremony remains live. If activation has prepared the package-allow update but the process dies before that update is committed, the visible policy is still the deny-by-default backup. `recover` therefore takes this early return; `restore()` then clears `pending_policy_recovery` and deletes the backup even though the original allow-policy operation is still pending against the unchanged baseline. That operation can subsequently be completed and committed, broadening wallet authority after recovery/status have declared the boundary clean. Please reconcile or cancel the stored `operation_id` and verify that no matching live policy-update challenge remains before clearing recovery state. A regression test should cover interruption after policy preparation but before ceremony completion/commit.

GavinPacini: Fixed in latest commits.

josh-richardson (commented):
P1 recovery-path findings

evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,1078 @@
...
+                if store.backup_path.exists() or state.get("pending_policy_recovery"):
+                    raise EvalError(
+                        "interrupted policy lifecycle exists; run recover first"
+                    )
+                # Source and binary lineage protect a new eval run, but are not
+                # prerequisites for restoring the exact digest-bound policy
+                # backup after a branch switch or rebuild.
+                validate_lineage(state, repo_root)
+                phase = time.monotonic()
+                policy.activate()
```
josh-richardson: [P1] Include activation in the automatic restoration boundary. `policy.activate()` runs before the `try/finally` that calls `policy.restore()`, so an activation error skips restoration entirely. This is reachable after authority has already broadened: for example, the byte-identical replay can commit the package-allow policy and then `_wait_for_policy_commit` can exhaust transient mounted reads and raise. `run` then exits with the recovery marker intact but the allow policy still active until someone notices and invokes `recover`. Please make any activation attempt that created the protected backup enter an unconditional restoration path, and cover an activation exception after policy mutation.

evals/harbor/harness/operator.py
```diff
@@ -0,0 +1,1078 @@
...
+                raise EvalError("stored confirmed policy-update projection is invalid")
+            return
+
+        cancel_path = (
+            self.definition.wallet_root
+            / "policy-updates/pending"
+            / operation_id
+            / "cancel"
+        )
+        cancelled = self.definition._write_route(cancel_path, b"cancel\n", 30)
```
josh-richardson: [P1] Reconcile a succeeded ceremony that is still in the pending projection instead of always cancelling it. After the debug driver succeeds, the operation remains under `policy-updates/pending` until the byte-identical replay observes `SUCCEEDED`, commits it, and moves the projection to `confirmed`. An interruption in that window makes recovery reach this cancel write, but Broker rejects cancellation of terminal ceremonies (`terminal ceremony cannot be cancelled`), so every subsequent `recover` stops here. If the policy commit reached Broker before the local transition, this can strand the package-allow policy active with no supported recovery path. Handle the pending-but-succeeded state by safely reconciling the exact staged bytes before restoring the backup, and add an interruption regression for driver success before replay/transition.

josh-richardson (approved):

PR #196, opened 2026-08-29
evals: guarded mainnet SOL canary and Harbor evaluation

0xdewy:
## In plain English

Bloom normally refuses to broadcast SOL on Solana mainnet. This PR adds a deliberately restricted canary build that can send **one pre-authorized transfer**, then adds a Harbor evaluation that exercises that path and independently checks what happened on-chain.

The goal is to test the real mainnet broadcast and recovery path without creating a general-purpose mainnet switch.

## What changes

- Adds the non-default `mainnet-canary` build capability.
- Requires an out-of-band authorization tied to one built Machine binary.
- Binds the authorization to one wallet account, source, destination, amount, fee ceiling, balance ceiling, network, and expiry.
- Consumes a `.spent` record before broadcasting, so the authorization cannot be reused.
- Revalidates every bound value immediately before the network send.
- Adds release checks that keep canary artifacts separate from production artifacts.
- Adds a Harbor task that asks an agent to make a native SOL transfer through Bloom's mounted VFS.
- Binds retries to one staged transfer and always attempts the destination-wallet cleanup sweep, even if mounted cleanup fails.
- Reuses the configured ceremony-URL pattern for both Solana and Hyperliquid evaluations.

## Safety boundary

Production builds still reject Solana mainnet. Building the canary requires both the compile-time feature and an explicit artifact label. Even a canary binary cannot broadcast without a valid, unexpired, single-use authorization matching the exact transaction and binary.

The normal Broker policy checks, semantic verification, and passkey approval still run.

## Stack

- Based on #170, which adds native Solana support.
- Uses bloom-directory/bloom-broker#32 and bloom-directory/bloom-signer#25 for isolated development ceremony ports.
- Uses bloom-directory/eval-images#9 for a pinned OpenCode runtime.

## Validation

- Default and labeled-canary configuration tests.
- Canary authorization, artifact-binding, expiry, cap, and single-use tests.
- Solana mainnet admission-gate tests.
- Harbor Solana harness tests, including multiple staged transfers and cleanup failures.
- CI release checks that reject canary capability in ordinary production artifacts.

## Checklist

- [x] Architecture docs updated — the isolated design is documented in `docs/design/solana-mainnet-canary.md` and the operator flow in `docs/operations/solana-mainnet-canary.md`.
- [x] Sealed Approval invariants respected — the normal Broker/passkey approval remains required, and the additional canary authorization is exact, bounded, expiring, and single-use.
- [x] Agent Documentation updated — N/A; this does not change the general VFS agent contract. The evaluation-specific instructions and runbook are included with this PR.


0xdewy:
Follow-up image reproducibility work: bloom-directory/eval-images#9 pins OpenCode 1.18.26 in the Harbor base image. After that image is merged and published, this PR should update its task image digest to the resulting immutable multi-architecture manifest.

0xdewy:
Closing after re-evaluating the scope and safety model. The base Solana work already supports broadcast to a pinned genesis when explicitly enabled, while the exact transaction remains subject to Broker/Signer approval. This PR first adds a mainnet prohibition and then adds a large one-shot exception around it, so the added canary, release, and Harbor machinery is not justified for the bounded mainnet test. The current unit-test regression also confirms it is not merge-ready. Any generally useful Harbor, reconciliation, or release-hardening pieces can return as small independent PRs.

PR #202, opened 2026-09-02
feat(config): add Arc Testnet to default networks

jer117:
Add Arc Testnet (chain ID 5042002, https://rpc.testnet.arc.io) to default_chains(), bringing the read-ready default set to 14 entries.

Arc meters gas in USDC rather than ETH, so the entry sets native_symbol = "USDC" while keeping native_decimals = 18: the native balance is still an ordinary 18-decimal EVM quantity, confirmed against the live testnet, whose base fee reads ~20 gwei. TxPreview::render already takes symbol and decimals as parameters, so plan previews denominate in USDC with no rendering change.

Chain ID and RPC endpoint were verified against the live network (eth_chainId returns 0x4cef52) and the head advances through the VFS at /chains/arc-testnet/head/number.

Arc support is testnet-only: Arc mainnet has not launched, so there is no `arc` entry. The three chain lists that enumerate the defaults are updated to say so — README.md, docs/AGENTIC_WALLET.md, and the agent-facing crates/bloom-vfs/src/docs/README.md, which additionally notes the USDC gas denomination so agents do not assume ETH.

Tests:
- local_default_shape asserts the full Arc spec (RPC, display name, broadcast, op_stack, symbol, decimals), not just the chain ID
- arc_ships_testnet_only_until_mainnet_launches fails if an `arc` mainnet key appears, naming the three docs to update alongside it
- default_chain_names_are_valid_protocol_tokens checks every chain key against the bloom-rpc-wire Token grammar, since chain names cross the triad in Broker and Signer policy and approval terms

No Broker or Signer change is required: `chain` is an opaque Token whose grammar admits the hyphen in "arc-testnet", and neither service holds a chain registry or allowlist.

Known gap, not addressed here: Config::load_or_init loads an existing config.toml verbatim and never merges newly added default chains, so installs that already ran `bloom init` will not see Arc without a migration.

## Checklist

      behavior changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,10 @@ Bloom gives an agent a safe wallet workspace:
 
 Bloom ships read-ready RPC defaults for major EVM networks — Ethereum,
 Base, Tempo, Robinhood Chain, Arbitrum, Optimism, Polygon, BNB Smart Chain,
-Avalanche, Gnosis, Linea, and HyperEVM — plus local Anvil. Per-chain
-broadcasting is enabled by default; set `allow_broadcast = false` on a chain to
-disable it.
+Avalanche, Gnosis, Linea, and HyperEVM — plus Arc Testnet and local Anvil.
+Arc coverage is testnet-only for now; an Arc mainnet entry is pending Arc
+mainnet launch. Per-chain broadcasting is enabled by default; set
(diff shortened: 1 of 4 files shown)


GavinPacini:
Closing as resolved by #271. - thanks for the work on this @jer117!

PR #204, opened 2026-09-02
fix(petals): enforce manifest-derived key scopes

### LINKED ISSUE #20: Replace generated route IDs in derived-key scopes with manifest-declared named scopes
## Problem

Petal authors currently have to pass package-internal route IDs in `PetalKeyRequest.allowed_routes` when deriving a scoped key. Those IDs are assigned by sorting canonical route patterns and numbering them sequentially (`r000001`, `r000002`, ...).

As a result, adding an otherwise unrelated route can renumber every route that sorts after it. The Hyperliquid Petal exposed this when four immutable receipt routes were added: its session-action scope and package-contract tests had to change IDs such as:

- `schedule_cancel`: `r000015` -> `r000019`
- `update_leverage`: `r000019` -> `r000023`
- session creation: `r000021` -> `r000025`

CI first failed on a stale route-count assertion and then on stale exact route IDs. More importantly, this means an opaque packaging implementation detail leaks into Petal application code and authorization tests.

Route IDs remain useful as internal artifact/provenance identities, and the package hash prevents an old scope from silently applying to a different package. This is therefore not presently a demonstrated privilege-escalation vulnerability. It is, however, a brittle and error-prone authoring interface for security-sensitive policy.

## Proposed design

Define immutable, named derived-key scopes in `petal.toml` using canonical route patterns rather than generated IDs. For example (syntax illustrative):

```toml
[[key.scope]]
name = "agent-session"
derive_route = "[network]/agent_sessions/[wallet]/new.json"
allowed_routes = [
  "[network]/agent_sessions/[wallet]/[session]/order.json",
  "[network]/agent_sessions/[wallet]/[session]/cancel.json",
  "[network]/agent_sessions/[wallet]/[session]/schedule_cancel.json",
  "[network]/agent_sessions/[wallet]/[session]/update_leverage.json",
  "[network]/agent_sessions/[wallet]/[session]/cancel_all",
  "[network]/agent_sessions/[wallet]/[session]/close_all",
]
operation_classes = ["hyperliquid.agent_action"]
```

Petal code would request the named scope and runtime bounds without supplying route IDs:

```rust
derive_key(PetalKeyRequest {
    scope: "agent-session".into(),
    wallet_id,
    key_slot,
    maximum_lifetime_ms,
})
```

At package/install time, the toolchain should:

1. Normalize and validate every declared pattern against the complete package route graph.
2. Resolve patterns to the package's opaque internal route IDs.
3. Include the derivation route in its own scope automatically, or enforce an explicit and documented rule.
4. Validate that scoped action routes have compatible capabilities/signing metadata.
5. Commit the resolved scope into authenticated package/provenance metadata.

At runtime, Bloom should resolve the requested scope from installed authenticated metadata and reject attempts to widen its routes, operation classes, crypto suites, or lifetime constraints.

The derivation route's macro could optionally reference the scope name for locality and compile-time checking, but the cross-route list should have one authoritative definition in the package manifest.

## Why not stabilize sequential IDs?

Hash-derived or otherwise stable IDs would reduce renumbering, but would still expose artifact identity to application authors and require a compatibility migration. Canonical patterns and named scopes provide a cleaner boundary:

- canonical patterns / scope names: developer-facing policy identity;
- route IDs: opaque package and runtime identity.

## Acceptance criteria

- Petal application code does not hardcode generated `rNNNNNN` IDs for derived-key scopes.
- Adding an unrelated route does not require editing existing key-scope application code or expected IDs.
- The packager rejects missing, malformed, duplicate, or incompatible scoped routes.
- The installed/authenticated package metadata contains the exact resolved scope enforced by the host.
- Runtime requests can select and narrow a declared scope, but cannot widen it.
- Package-contract tests cover route insertion without authorization drift.
- A migration/backward-compatibility plan exists for the current `allowed_routes` request field.

This likely requires coordinated changes in the Petal manifest schema/builder/SDK and Bloom's package ingestion and key-derivation host implementation.

0xdewy:
## Summary

Closes bloom-directory/petal#20.

A declared `[[key.derive]]` scope is now complete and authenticated: canonical `allowed_routes`, `operation_classes`, `allowed_crypto_suites`, and `maximum_lifetime_ms`. Packaging resolves route patterns to immutable route IDs, includes the derivation route, and rejects missing, unknown, duplicate, unsigned, unsupported-suite, zero-lifetime, or operation-class-incompatible declarations.

At runtime Bloom replaces guest route IDs with the packaged scope and permits only strict narrowing of operation classes, crypto suites, and lifetime. Attempts to widen any bound fail before the host derivation call. The package-contract regression inserts an unrelated earlier-sorting route and proves the semantic scope is unchanged despite route-ID renumbering.

For migration compatibility, packages that omit all new manifest scope bounds retain the legacy guest-supplied request. Once any new bound is present, all are required. Bloom also validates wallet IDs at the daemon signing boundary so malformed IDs fail as input errors instead of opaque permission denials.

Derived-key authority remains bound to one immutable package hash. Succession is intentionally deferred: upgrades request a new key/session with fresh user authorization.

## Validation

- `cargo test -p bloom-petals` (170 unit tests plus integration tests)
- focused bloom-daemon approval test
- clippy for bloom-petals and bloom-daemon with warnings denied
- rustfmt and diff checks
- exact built 70-route Hyperliquid archive contract via bloom-directory/bloom-petal-hyperliquid#17

## Checklist

- [x] Architecture docs updated for complete scopes and package-bound succession
- [x] Sealed Approval invariants respected: authenticated bounds can only narrow authority
- [x] Agent Documentation updated: Petal authoring guidance documents every scope field

### DIFF
--- a/crates/bloom-daemon/src/lib.rs
+++ b/crates/bloom-daemon/src/lib.rs
@@ -1473,6 +1473,8 @@ impl PetalHost for DaemonPetalHost {
         let broker = self.broker.as_ref().ok_or_else(|| {
             HostError::Backend("SERVICE_UNAVAILABLE: Broker client is not configured".into())
         })?;
+        let wallet_id = bloom_broker_api::Token::new(req.wallet.clone())
+            .map_err(|error| HostError::Invalid(format!("wallet: {error}")))?;
         let context = req.context.as_ref().ok_or_else(|| {
             warn!(
                 wallet = %req.wallet,
@@ -1672,8 +1674,7 @@ impl PetalHost for DaemonPetalHost {
             .transpose()
             .map_err(|error| HostError::Invalid(error.to_string()))?;
         let trusted_request = TrustedPetalSignRequest {
-            wallet_id: bloom_broker_api::Token::new(req.wallet)
-                .map_err(|error| HostError::Invalid(error.to_string()))?,
+            wallet_id,
             preimage: req.preimage,
             claimed_hash: bloom_broker_api::Digest32::from_bytes(req.claimed_hash),
             crypto_suite,
@@ -1737,6 +1738,8 @@ impl PetalHost for DaemonPetalHost {
         let broker = self.broker.as_ref().ok_or_else(|| {
             HostError::Backend("SERVICE_UNAVAILABLE: Broker client is not configured".into())
         })?;
+        bloom_broker_api::Token::new(req.wallet.clone())
+            .map_err(|error| HostError::Invalid(format!("wallet: {error}")))?;
         let context = req.context.as_ref().ok_or_else(|| {
             HostError::Denied("payload batch signing requires trusted Petal provenance".into())
         })?;
@@ -5273,6 +5276,15 @@ mod tests {
             key_ref: None,
             context: Some(context),
         };
+
+        let mut malformed_wallet = request.clone();
+        malformed_wallet.wallet = "0x0000000000000000000000000000000000000001".into();
+        let error = host
+            .sign_payload_outcome(malformed_wallet)
+            .await
+            .unwrap_err();
+        assert!(matches!(error, HostError::Invalid(_)));
+        assert!(error.to_string().contains("wallet"));
 
         let SignOutcome::ApprovalPending(pending) =
             host.sign_payload_outcome(request.clone()).await.unwrap()
--- a/crates/bloom-petals/src/package.rs
+++ b/crates/bloom-petals/src/package.rs
@@ -148,6 +148,14 @@ struct KeyPolicyToml {
 struct KeyDerivePolicyToml {
     route: String,
     operation_classes: Vec<String>,
+    #[serde(default)]
+    allowed_routes: Vec<String>,
+}
+
+#[derive(Debug)]
+struct ValidatedKeyDerivePolicy {
+    operation_classes: Vec<String>,
+    allowed_route_ids: Vec<String>,
 }
 
 #[derive(Debug, Default, Deserialize)]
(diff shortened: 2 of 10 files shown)


GavinPacini:
@0xdewy What are the possible options for `allowed_crypto_suites`? I'm not sure where they're actually coming from. Thanks!

0xdewy:
@GavinPacini Three values, from the `CryptoSuite` enum in bloom-broker (`crates/bloom-broker-api/src/crypto.rs`, pinned here at `a4711e8`):

- `secp256k1-keccak256-recoverable`
- `secp256k1-sha256-recoverable`
- `ed25519-message`

The suite picks which hash runs over the payload before the key signs it. It matters because an EVM wallet's key can do **both** keccak and sha256 (`DerivationProfile::frozen_crypto_suites()`), and Ethereum only accepts keccak — so a key limited to sha256 cannot sign anything a chain will take.

Honest caveat: it is not restricting anything today. Polymarket and hyperliquid both declare keccak, the more capable of the two. What it does do is put the suite list inside `PetalKeyScope`, whose digest is the key's identity — so a different suite list is a different key — and move the choice out of the running wasm into the installed manifest.

(Edited: an earlier version of this comment said an unbounded key could move the user's funds. That was wrong — the derived key is a child key and cannot touch the parent wallet balance.)

Two small things: `package.rs` checks these names against hardcoded literals while `bloom-daemon/src/lib.rs:959` already parses them via `serde_json::from_value::<CryptoSuite>`, worth matching so the enum stays the one source of truth. And I can add the three values to `docs/petals/authoring-petals.md` — nothing in `petal.toml` hints at them, which is presumably why you asked. Say the word on either.

GavinPacini:
@0xdewy That makes sense, thanks! Yes, worth aligning against the enum and adding something into `authoring-petals.md` IMO.

GavinPacini (approved):

PR #214, opened 2026-09-04
Fix EVM signing recovery and pending-nonce selection

### LINKED ISSUE #217: A multi-step action stages its next transaction on a nonce the previous one already spent
## What happened

A Morpho deposit is two transactions: approve, then deposit. Confirming the approve broadcast it and the Petal immediately staged the deposit. Both got **nonce 5**.

```
0017-43235: nonce=5 status=sent    to=0xA0b8...eB48 data=0x095ea7b3   (approve)
0018-49500: nonce=5 status=pending to=0xBEEF...64CB data=0x6e553f65   (deposit)
```

By the time I looked, the approve had mined and the account was at nonce 6. The staged deposit could never be broadcast: it would be rejected as nonce too low, after an owner had been asked to approve it in a ceremony.

## Cause

`Outbox::highest_pending_nonce` scans only the `Pending` directory:

```rust
let dir = self.state_dir(wallet, chain, OutboxState::Pending)?;
```

and the next nonce is

```rust
pending_high.map_or(chain_nonce, |h| chain_nonce.max(h + 1))
```

There is a window where a transaction is in neither term. Once it broadcasts it moves to `sent`, so `highest_pending_nonce` stops counting it, and until it mines `chain_nonce` does not include it either. Anything staged in that window gets the nonce that transaction is already using.

For a single user action the window is invisible. For a multi-step Petal action it is the normal path: step N+1 is staged the moment step N broadcasts, which is exactly when the gap is open.

## `nonce_conflicts.json` does not catch it

The same wallet reported:

```json
{ "outbox_sent_nonces": [0,1,2,3,4,5], "conflicts": [] }
```

~~It compares the chain nonce against *sent* nonces.~~ **Corrected:** it compares mempool-index observations for the address against the union of pending **and** sent outbox hashes, skipping any mempool hash that is one of ours (`bloom-vfs/src/handlers/wallets.rs:3242-3260`).

The conclusion is the same but the mechanism matters for anyone fixing it: because the in-flight transaction's hash *is* one of ours, it is filtered out, so an internally-duplicated pending-vs-sent nonce can never appear as a conflict. It only ever detects a *foreign* transaction at one of our nonces. A guaranteed-invalid transaction sat in the queue while the surface built to report nonce problems said there were none.

## Neither recovery path works

- `cancel` is refused before the approval is dispatched — #213.
- `replace` is documented as *"the original nonce is preserved"*, so it cannot fix a transaction whose defect **is** its nonce.

So the only way out was deleting the outbox entry and its `central_outbox` twin by hand, which I did on a live mainnet wallet. Re-staging then picked up nonce 6 correctly and the deposit went through (`0xa2dcc5b7b14ca06c7d3714e0a2ac6c172e3bbcb346941207ca6edc3bb8a49a2f`).

## Fix

> **Corrected 2026-09-04.** I ranked these the wrong way round, on a premise the code contradicts. Fixed in bloom-directory/bloom#214.

~~Count sent-but-unmined entries when deriving the next nonce, or read the chain nonce at the `pending` block tag rather than `latest`. Either closes the window. The first is probably better, since it does not depend on the node's mempool view.~~

The second option was not a design question at all — it was a one-line change, and the codebase had already decided it. Staging read its nonce from the pinned read session, and `Session::nonce` documents itself as *"the historical transaction count at the pinned block, not the pending nonce — the engine's pending semantics live on `ChainClient::nonce`"* (`bloom-rpc/src/session.rs:105-113`). `ChainClient::nonce` already calls `.pending()`, and `assert_nonce_not_ahead_of_chain` already uses it at broadcast time.

So the two nonce reads in the engine disagreed with each other, and staging was using the one its own documentation warns against. Staging now uses the same source as the broadcast guard.

My reason for preferring the other option — not depending on the node's mempool view — does not survive contact with that: the broadcast-time gap guard already depends on exactly that view, and a wallet whose node cannot see its own mempool has a worse problem than nonce selection.

The stale doc comment on `highest_pending_nonce` is corrected in the same PR. It claimed to count *"every pending entry ... whether or not it has broadcast yet"*, which is false — broadcasting moves the entry to `sent/` — and is what made this gap read as impossible.

Covered by `anvil_stage_does_not_reuse_the_nonce_of_a_broadcast_but_unmined_tx`, which disables anvil's auto-mining to hold the window open and reproduces the duplicate nonce without the fix.

**Still open:** `nonce_conflicts.json` should compare our own pending nonces against our own sent ones. Had it done so, this would have been visible before the owner was asked to approve anything, instead of after. The fix above prevents the collision; it does not make the reporting surface able to see one.

## Repro

1. Any two-step Petal action on one wallet and chain — a Morpho deposit needing an approve does it.
2. Confirm step 1 and let it broadcast.
3. Read the nonce on the automatically staged step 2, before step 1 mines.

Observed on mainnet, wallet at nonce 5, via `bloom-petal-morpho`.

0xdewy:
## Summary

The second transaction of a multi-step action (for example an approve followed by a deposit) could be staged on the nonce the first had already spent, and a refused signature left a staged transaction stuck until its state files were deleted by hand. Staging now reads the chain's pending nonce so a broadcast-but-unmined nonce is not reused, a permanent refusal resets the signing lineage so the transaction can simply be retried, and an approval-required error now names the owner ceremony to complete instead of answering "permission denied".

## Why

- **Live incident, nonce reuse:** a Morpho deposit was staged at the just-broadcast approve's nonce and could never broadcast; recovery meant re-drafting the transaction by hand (#217). Staging read the nonce at the pinned historical block, which cannot see a transaction still in the mempool, and the outbox scan only covers `pending/` — a broadcast transaction has already moved to `sent/`. Between broadcast and mining, neither source sees it.
- **Live incident, refused signature:** after the Broker refused a swap signature, the staged row kept its dead approval and operation IDs. Confirm re-reserved the same approval, cancel was refused, and the only way out was deleting `ceremony.json` by hand on a live mainnet wallet. A batch whose signing operation reached a terminal state wedged the same way, permanently.
- **Illegible approval step:** confirming, cancelling, or replacing a transaction that needed owner approval answered "permission denied" — telling the caller they may not act, when the engine already held the ceremony URL they needed to act next.

## What changed

- **Nonce selection:** staging reads the chain nonce from the pending block — the same source the broadcast-time gap guard already uses — instead of the pinned historical block. A doc comment that wrongly claimed `highest_pending_nonce` covered already-broadcast entries is corrected.
- **Refusal recovery, single and batch:** when the Broker's error contract says the refusal is permanent with no durable effect, the signing lineage is replaced, so the next confirm mints a fresh ceremony. Retryable errors and ambiguous outcomes (such as a lost response) keep the existing lineage, because a signature may exist and must be reconciled rather than re-signed.
- **Terminal batch operations:** denied, cancelled, and failed operations release the row with a fresh lineage. A quarantined batch keeps its identity (a signature may exist) but stops offering a ceremony URL that can no longer do anything.
- **Approval errors:** a new `ApprovalRequired` error carries the reason, action id, expiry, and ceremony URL. IPC keeps the existing `-32007` code with the detail in the message, the CLI reports permission-denied with the URL in the message, and Petals receive the full detail. NFS still maps to access-denied — a filesystem has no errno for a pending ceremony; the ceremony stays discoverable in the projection beside the staged action.

## Not in this PR

- `nonce_conflicts.json` still cannot report internal pending-vs-sent nonce duplication (the still-open reporting gap in #217). This PR prevents the collision; it does not make the reporting surface able to see one.
- Petal package-policy enforcement requested in #215, and Solana approval recovery in #234.
- The single-payload terminal-operation arm keeps its current behavior: its escape is the `action_id` gate, which requires the dispatch marker to stay set. Changing that asymmetry wants its own change.
- Earlier revisions of this PR also carried token pricing, wallet export, and the per-chain policy view; those were removed and remain recoverable from the branch history. Robinhood network configuration already landed in #145.

## How to verify

- `cargo test -p bloom-tx -p bloom-vfs --lib` — 158 transaction tests and 350 VFS tests pass.
- `cargo test -p bloom-it --test anvil_e2e` — all 3 pass, including the new `anvil_stage_does_not_reuse_the_nonce_of_a_broadcast_but_unmined_tx`, which disables anvil's auto-mining to hold the window open and reproduces the duplicate nonce without the fix.
- New engine tests: a refused single or batch signature resets the lineage; a lost response keeps the dispatch marker; a refusal can be superseded by cancellation; definite terminal batch states release the row while quarantine holds it. Removing the retry-class guard fails both refusal tests on `ApprovalRearmRequired`.
- CI on the head is green: fmt, clippy, doctests, unit, local integration, external e2e, locked release build, Linux W0, and the Solana validator transfer lifecycle. macOS and docker jobs are skipped.
- Not run: live triad/browser testing.

## Risk and rollout

- No wire, config, or dependency changes. IPC keeps the `-32007` code and NFS keeps access-denied, so existing clients and mounts behave as before. Each fix is independent and confined to its recovery path; rollback is a plain revert.

## Reviewer notes

- The substantive change is the error-contract gates in `triad_sign_evm_payload` and `triad_sign_evm_batch` in `crates/bloom-tx/src/tx_engine.rs`, plus the one-line nonce-source swap in `stage`. Most of the `tx_engine.rs` diff is regression tests and fixture plumbing; the remaining files are one-to-few-line error adapters, and `outbox.rs` is a doc-comment correction.

## Checklist

      behavior changed — N/A: no authority or wire-contract change; recovery follows the existing Broker error contract
      bytes) — fresh identities require fresh approval; ambiguous outcomes retain their existing identities
      and affected Petal READMEs) — N/A: the stage/approve/confirm flow is unchanged; errors now expose the required next step, and no new VFS path is added

Fixes #217


0xdewy:
Added `21a1826`: clear `sign_dispatched` when the Broker refused to sign.

The marker is written before `sign_exact_payload` on purpose — a Machine that dies mid-flight must assume a signature may exist — but it was left set even when the Broker answered with an explicit refusal, recording a dispatch that provably never happened.

The protocol already draws this line. Every error code carries an `ErrorContract`, and `durable_effect: none` is the Broker stating it refused before anything durable occurred. Anything else still means a signature might be out there, so the marker stays.

This is not cosmetic: the stale marker stranded the row in both directions. Confirm re-reserved the same dead approval, and cancel was refused by the `action_id` gate in `triad_sign_evm_payload` *because* `sign_dispatched` claimed a signature might exist. Recovering it meant deleting `ceremony.json` by hand on a live mainnet wallet.

Seen live as `CLAIM_INVALID: approval is not valid at reservation time`, after an owner completed a ceremony and the commit was not issued before the approval's validity window closed.

Two tests, since the narrowness is the whole point:
- `a_refused_signature_clears_the_dispatch_marker_it_never_earned`
- `a_lost_response_keeps_the_dispatch_marker_because_a_signature_may_exist` — `ServiceUnavailable` after a commit, where clearing would let a retry sign the same nonce twice

`cargo test -p bloom-tx`: 153 passed, 0 failed.

Full context on this failure mode and the three others in the same family is in #216.

0xdewy:
Added host-verified canonical ERC-20 Petal classification in 552136f. Paired Petal hardening PR: https://github.com/bloom-directory/bloom-petal-robinhood/pull/1

0xdewy:
Synced with base; pins current; CI failing.

0xdewy:
Correction to my last comment: CI is green. Merged master forward (`44efb1b`), which brings in #170 (BIP-39 HD accounts and native Solana) and #220 (obsolete VFS surfaces removed). There were no textual conflicts. The one semantic conflict was two new BIP-39 account-key tests from master that build `TriadBrokerFixture` without the `deny_sign` and `operation_terminal` fields this branch adds; `3a9b892` initializes both to `None`. No production code changed.

On `3a9b892`, every check passes, including clippy, unit, local integration, external e2e (anvil), the release build and the Solana validator lifecycle. Locally, `cargo clippy --workspace --all-targets -D warnings` is clean, and the bloom-tx, bloom-vfs, bloom-petals, bloom-mount and bloom-daemon tests all pass.

That replaces the "Full CI/clippy … not run" line in the description. Live Triad/browser testing still hasn't been done. Ready for review; start with the error-contract gates in `tx_engine.rs`.

0xdewy:
Merged master (`9dc4039d`) to clear the conflict with the v0.3.0 release. The one conflict was in `tx_open_err`: kept this PR's `ApprovalRequired` mapping, and took master's `PolicyDenied` arm because master removed `TxEngineError::BroadcastDisabled`. Tests run on the merged head:

- `cargo check --locked --workspace --all-targets`: pass
- `cargo fmt --all -- --check`: pass
- `cargo clippy --locked -p bloom-vfs -p bloom-tx -p bloom-daemon -p bloom-mount --all-targets -- -D warnings`: pass
- `cargo test --locked -p bloom-vfs --lib`: 410 passed
- `cargo test --locked -p bloom-tx --lib`: 158 passed
- `cargo test --locked -p bloom-daemon --lib`: 99 passed
- `cargo test --locked -p bloom-mount --features mount --lib`: 78 passed

0xdewy:
The merge above broke one test in CI (`local integration tests`), fixed in `5ffa99e0`.

`anvil_stage_does_not_reuse_the_nonce_of_a_broadcast_but_unmined_tx`, which this PR adds, staged through `wallets/alice/chains/anvil/outbox/…`. Master's bloom#282 denies wallet-level outbox writes, so after the merge the first stage failed with `permission denied`. Git saw no conflict because master never had this test. Its five paths now use `wallets/alice/0/chains/anvil/outbox/…`, like the other tests in `anvil_e2e.rs`.

`cargo test --locked -p bloom-it --test anvil_e2e`: 3 passed, including this one. `cargo fmt --check` passes. No other added line in this PR uses a wallet-level chain path.

jer117:
These overlap a little but its still a separate issue. 

https://github.com/bloom-directory/bloom/issues/320

PR #220, opened 2026-09-05
fix: remove obsolete VFS surfaces and bound paid HTTP

violet-agent:
## Summary

Resolves review findings 1, 3, and 4:

- removes the router-wide mutation lock from awaited handler work while preserving the exclusive `write_then_lookup` identity-capture boundary;
- serializes only correlation-ID allocation plus audit-intent publication, adds bounded audited-effect admission, and gives each correlation ID a nonce so concurrent identical writes cannot collide;
- bounds paid-HTTP clients to a 30-second request timeout, with a deterministic timeout regression test;
- removes unused native Enso configuration and the inert active-capabilities VFS/model surface, including stale guidance and specification references.

The obsolete capabilities directory is absent from listing and lookup, with a
regression test covering both. Mounted guides point to Sealed Approvals instead.

Paid confirmations and cancellations share a per-request lock. A per-wallet
lock spans payment checks through receipt publication, while unrelated work
remains concurrent. Before merchant transmission, the existing receipt records
the full possible charge as unresolved and is atomically persisted with directory
syncs. Completion updates that same record without double counting. Interrupted
requests cannot execute again, but other requests can use the remaining budget;
there is no blanket wallet block or new public VFS path. Unresolved charges remain
reserved without automatic expiry; this PR does not add a reconciliation API.
Legacy interrupted markers retain retry protection and accounting.

Preflight failures and incomplete local credential preparation remain retryable.
Tests cover interruption/restart, remaining-budget boundaries, completion and
timeout accounting, legacy markers, and corrupt records. Current canonical Broker
policy does not expose configurable legacy HTTP daily caps, so cap tests cover
the accounting seam. Malformed accounting records fail closed.

Cancellation now records an audit result with an explicitly unknown external
effect outcome before releasing admission. If that audit append fails, admission
remains occupied and audit degradation fails closed.

The Tempo session RPC timeout P2 is deferred to [#274](https://github.com/bloom-directory/bloom/issues/274).
The temporary preparation deadline has been reverted. The recommended follow-up
upgrades Bloom's reqwest from 0.12 to 0.13 to align with Alloy and reuse a bounded
client. The Tempo timeout issue remains open; this PR resolves the audit-cancellation P2.

## Verification

- `cargo fmt --all -- --check` and `git diff --check`
- `cargo test -p bloom-vfs --locked` — 400 unit tests and 14 integration tests passed
- `cargo test -p bloom-proto --locked --lib config::tests` — 37 passed
- `cargo clippy -p bloom-vfs --all-targets --locked -- -D warnings` — passed
- Removing either scoped lock independently makes the concurrency regression fail.
- `cargo check -p bloom-daemon --locked` — passed
- obsolete-surface reference sweep for `EnsoConfig`, `default_enso_url`, `CapabilityViewEntry`, and `capabilities/active.{json,md}` — no matches

- Cancellation regressions cover 80 aborted writes/side-effecting reads and audit-append failure.

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A: the only removed public-surface reference was in a dated implementability specification, updated in this PR; no architecture contract was changed.
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — N/A: this change removes inert configuration/discovery code and narrows VFS synchronization; it does not alter signing, approvals, or grants; payment-attempt receipts are durably persisted before transmission.
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — updated the VFS guidance to remove the retired active-capabilities surface; no Petal README referenced it.


0xdewy:

josh-richardson (changes requested):
Two P2 issues block approval:

1. [P2] Cancellation releases the semaphore without resolving the durable audit intent (`crates/bloom-vfs/src/router.rs:73-79`). `PendingEffect` drops its `OwnedSemaphorePermit` when a write future is aborted, but no `machine.effect.result` is appended. A new operation can then take the released slot. I reproduced 64 sequential cancellations leaving 64 pending correlations even though the limit is 16; the next write hits `MAX_PENDING_MACHINE_EFFECTS` and permanently latches mutation degradation. The admission permit must remain coupled to the durable intent until a result/reconciliation is recorded, including cancellation.

2. [P2] The 30-second timeout does not cover Tempo session RPC (`crates/bloom-vfs/src/handlers/requests.rs:45-63`). `RealMppBackend` receives the bounded client, but `prepare_session_credential` constructs an Alloy provider with `connect_http(rpc_url)`, which creates its own default `reqwest::Client`; `RealMppBackend.client` is not used for that provider. A silent Tempo RPC can therefore keep confirmation, the wallet lock, and an audited-effect slot open indefinitely. Build the provider from the bounded client or wrap the whole preparation in an explicit deadline, and add a hanging-RPC regression test.

GavinPacini:
Update on the two P2 findings:

- **Audit cancellation P2 resolved** in 49e1d0e6. Cancelling a write or side-effecting read now records an audit result with an explicitly unknown external effect outcome before releasing its admission slot. If the audit append fails, the slot remains occupied and mutation degradation fails closed. Regression tests cover 80 sequential cancellations and audit-append failure.
- **Tempo RPC timeout P2 remains open**, tracked in #274. We reverted the temporary Tempo preparation deadline in de29a4b2 while retaining the audit fix. The recommended follow-up is to upgrade Bloom’s reqwest from 0.12 to 0.13, aligning it with Alloy/Tempo so the provider can reuse a bounded HTTP client, with a hanging-RPC regression test. This PR no longer claims to resolve the Tempo finding.

Validation after the Tempo revert: 400 VFS unit tests and 14 integration tests passed, along with clippy, formatting, and diff checks.

josh-richardson (changes requested):
Re-review of `de29a4b2`:

The cancellation/admission P2 is resolved. `PendingEffect::drop` now records a cancellation result before releasing admission, retains the permit if that append fails, and the new cancellation regressions pass locally.

The Tempo session RPC P2 remains unresolved and is explicitly deferred to #274. `prepare_session_credential` still uses Alloy `connect_http`, whose separate default reqwest client has no request timeout. A silent RPC can therefore hold confirmation, the per-wallet lock, and an audited-effect slot indefinitely. Tracking the defect is useful, but it does not make the current behavior safe to merge under the stated no-P1/P2 criterion. Please land #274’s bounded provider/deadline before approval.

josh-richardson (approved):
Approved. The audit-cancellation blocker is fixed and covered by regression tests. The Tempo session RPC timeout is accepted as explicit follow-up work in #274.

Issue #78, opened 2026-07-08
[hyperliquid]: enable by default


GavinPacini:
Similar to #75, `config.toml` values are required for Hyperliquid to be enabled. We should default to it being enabled.

0xdewy:
Related: #75 asks for the same default-enabled behavior for DeFi. These can likely share one configuration/defaults pass for first-party petals; Hyperliquid may have its own endpoint/account defaults to verify separately.

PR #222, opened 2026-09-06
Deploy from Foundry, Hardhat and Ignition with a Bloom wallet

### LINKED ISSUE #206: Transaction approval page cannot state the amount, destination or fee
## Summary

An owner approving an ordinary transaction is not shown what it moves. The
ceremony page renders `asset_amounts: []` and the disclosure:

> Bloom has not established the execution effects of these opaque payload
> digests and hashes.

The amount, destination and fee are absent — not because the design lacks a
place for them, but because an existing channel is unconnected at both ends.

## Observed

Confirming a 0.0003 ETH send on Base from a BIP-39 wallet. The review showed
wallet, account (`m/44'/60'/0'/0/0`), expiry, operation class
`transaction.confirm`, and a one-signature cap — all correct and useful — but
nothing identifying the transfer. `plan.md` beside the staged transaction has
the facts; the page does not.

## Cause

Three facts together:

1. `SystemUseClaim` already declares `declared_debits`, `declared_destinations`
   and `declared_fee`, exactly as `PetalUseClaim` does.
2. The transport already carries it: `PayloadSignRequest.system_use_claim:
   Option<SystemUseClaim>`.
3. Machine has **no references to `SystemUseClaim` anywhere**. It never builds
   one, so the field is always `None` for a transaction confirmation.

Separately, the Broker's `canonical_review_plan` populated `asset_amounts` only
from the Petal claim; `system_claim` was consulted solely for disclosures. So
even a populated system claim would not have rendered.

The digest chain is intact throughout — `canonical_plan_facts_digest` binds the
facts. They are simply never transmitted in a form the page can display.

## Suggested fix

Broker half (done in bloom-broker, branch `fix/ceremony-review-legibility`):
read `declared_debits` / `declared_fee` from the system claim as well as the
Petal claim. Inert until the Machine half lands.

Machine half (not attempted):

- add `system_use_claim` to `ExactPayloadSignRequest` and
  `ExactPayloadBatchSignRequest`, threading it to `PayloadSignRequest`;
- populate it for `transaction.confirm` from the staged transaction's value,
  recipient and fee.

Care needed: `service.rs` folds the claim into `approved_claim_digest` via
`jcs_digest`, so the claim must be byte-identical at approval time and at
signing time, which means persisting it in the triad signing state. An
inconsistent claim will fail confirms closed.

## Impact

Not blocking — transactions succeed. But the review page is the only point
where a human authorises value, and it currently cannot state the value. An
approval surface people cannot read is one they learn to click through, which
defeats its purpose.

## Related

The same shape appeared twice more while testing and both are fixed:

- custody ceremonies carried no review manifest at all outside legacy passkey
  migration and Petal key derivation, so eight kinds — `WalletDelete` included
  — rendered as raw base64 (bloom-broker, same branch);
- `TxEngineError::ApprovalRequired` was flattened to a bare `permission
  denied`, discarding the `ceremony_url` the engine already held (bloom, branch
  `feat/robinhood-chain-support`).

### LINKED ISSUE #221: Deploy EVM contracts with Bloom wallets from Foundry and Hardhat without exposing signing keys
## Implementation delivered

The EVM release is implemented in [Machine PR #222](https://github.com/bloom-directory/bloom/pull/222) and [Broker PR #36](https://github.com/bloom-directory/bloom-broker/pull/36). Machine pins Broker commit `2030c38ab8d3056ab19fb6fee55a9682792caacf`; both services require Machine/Broker protocol 1.5. Signer protocol and Petal WIT are unchanged.

**Scope refinement:** use the existing native wallet/outbox for deployment jobs, exposed through `bloom deploy rpc/list/status/resume`, rather than adding a WASM Petal that duplicates these surfaces. Foundry, Hardhat remote accounts, and Ignition share one authenticated local RPC adapter. The agent explicitly continues the same submission after each exact owner approval.

[Runnable multi-contract examples and recovery guide](https://github.com/bloom-directory/bloom/blob/agent/root/evm-deploy/examples/evm-deploy/README.md) include wallet selection and the required numeric-chain policy opt-in. The supplied multicall-scripting project passed on local Anvil, including factory CREATE2 deployments, runtime-code checks, and a second run without additional transactions.

Validation passed: 784 relevant Machine library tests, 165 Broker tests, 37 Linux release-packaging tests, workspace compile/strict Clippy/formatting, the actual CLI compatibility test for Foundry/Hardhat/Ignition, and a separate legacy/EIP-1559 creation test. Broker CI is green. Machine's new compatibility CI job uses pinned Foundry and npm dependencies.

The deployment harness simulates owner approval with a test Broker. It verifies response loss, same-request retry, fresh engine/outbox reload, unmined nonces, framework resume/reconciliation, authentication rejection, and cached receipts during an RPC outage. It does **not** claim live passkey onboarding, a full production service restart, or a real two-minute approval timeout/expiry exercise. Keep the combined lifecycle and real-user acceptance items open until those release exercises are recorded. Market demand and repeat use also remain unvalidated; no mainnet transaction was performed.

## Problem and intended outcome

A developer using a coding agent should be able to run an existing Foundry deployment script with a Bloom wallet, review the concrete transactions, and deploy without putting the signing key in the project, an environment variable, or the agent process. The useful first release must handle multiple contract deployments and dependent initialization/configuration calls, not just a Counter deployment.

Bloom retains key custody and authorization; Foundry and Hardhat retain compilation and deployment scripting. Public wallet addresses and network settings can remain in project configuration. RPC/explorer credentials, where necessary, are separate from wallet custody and should use existing protected configuration mechanisms.

## Product rationale and evidence limits

The stronger value proposition is “let my coding agent finish deployment without giving it my private key.” Avoiding `.env` alone is a modest convenience: Hardhat already offers an encrypted keystore. Likely early users are agent-assisted Solidity developers, builders deploying prototypes/testnets, and existing Bloom users. Production protocol teams with established release procedures are a later audience.

Technical feasibility has been investigated; market demand has not been validated. First validation: use this for a real Bloom developer project, then have 3–5 developers try their own existing projects. Measure successful setup, deployment, recovery, and voluntary repeat use. Do not treat demo enthusiasm as repeat-use evidence.

### EVM: one JSON-RPC adapter for Foundry and Hardhat

- Foundry `forge script --unlocked --sender ADDRESS --broadcast` submits unsigned requests via `eth_sendTransaction`. Scripts using `vm.envUint("PRIVATE_KEY")` must switch to `vm.startBroadcast()` or the public-address overload.
- A local experiment with Forge 1.7.1 and Anvil successfully deployed a contract and executed a dependent setter through a forwarding JSON-RPC proxy. Forge's explicitly constructed environment contained only PATH and a thread-count setting, with no wallet credentials. The final contract value was checked on-chain. This validates the upstream integration point, not an implemented Bloom adapter or production signing.
- The experiment observed `eth_chainId`, `eth_feeHistory`, `eth_gasPrice`, account/balance/code/block/nonce/receipt reads, `eth_sendTransaction`, and Foundry node-detection/optimized account methods. Unsupported optional methods need correct JSON-RPC errors and fallback behavior.
- Hardhat supports `accounts: "remote"`. Provider-backed scripts and Ignition can use the same adapter; scripts that explicitly construct local private-key wallets require adaptation.
- Keep Ignition's dependency graph, journal, and reconciliation. Bloom records its own authorization, transaction identity, broadcast attempts, and receipts; it must not impersonate Ignition's deployment state.
- A generic RPC adapter cannot discover every future transaction of an arbitrary script in advance. Receipt-dependent scripts must use exact per-transaction authorization initially. Do not promise one approval for arbitrary scripts without a separately verified bounded authorization design.

### Solana: defer implementation, preserve the findings

- Anchor and native Rust compile programs to `.so` artifacts (`anchor build`, `cargo build-sbf`). A future Bloom driver should accept these artifacts and preserve existing build tools.
- Solana `sendTransaction` takes an already signed transaction. Agave's signer parser has file/prompt/stdin/hardware/presigner sources, not a generic Bloom HTTP/Unix signing source. An EVM-style RPC proxy alone cannot supply Solana signing.
- Deployment requires account creation, chunk uploads, finalization, multiple signing roles, rent funding, blockhash refresh, and resumable state. Program identity, fee payer, buffer identity/authority, and upgrade authority must be represented separately; long-lived keys remain in Signer.
- Bind the public program ID to Anchor configuration and `declare_id!`; do not silently trust a generated local keypair as the funded wallet or upgrade authority.
- Current English Solana docs describe loader-v3; some translated pages recommend v4, and installed Agave 3.0.0 exposes both. Loader availability must be checked against a pinned, tested toolchain and target cluster rather than inferred from docs or CLI presence.
- Agave v4 exposes offline options, but source inspection shows `sign_only` can return one initial/final transaction rather than a complete upload plan. It is not a ready-made generic remote deployment signer.
- Existing Bloom Solana transfer work accepts a single signer and a single System Program transfer. Deployment requires a separate semantic verifier and multi-signature assembly. Do not weaken the transfer verifier into arbitrary Ed25519 signing.

### 1. Native EVM contract creation

- Represent direct creation distinctly from calls all the way through intent parsing, staged persistence, simulation, policy review, unsigned encoding, signing, replacements, broadcast, and receipts.
- Encode creation using `TxKind::Create` / absent RPC `to`. The zero address is an ordinary recipient and must never stand in for creation.
- Support constructor initcode including ABI-encoded arguments and native endowment; use existing build artifacts/tools for linking and encoding.
- Support legacy and EIP-1559 transactions on existing compatible Bloom chain profiles.
- Show creation explicitly in the review, with sender/network/nonce, value, initcode commitment, predicted direct-CREATE address, and fees. Predictions are conditional; receipts and deployed code establish the outcome.
- Preserve old persisted call transactions and the existing Petal WIT interface. A WIT shape change must have an explicit version migration, not an in-place @0.1.0 change.

### 2. Developer-tool adapter

- Expose a local authenticated, wallet-and-chain-scoped EVM JSON-RPC endpoint and straightforward CLI startup instructions.
- `eth_accounts` returns only the selected public wallet. Read methods use the configured chain; `eth_sendTransaction` stages and signs through the existing Machine → Broker → Signer path.
- Support Foundry scripts with several deployments and dependent calls, including constructor arguments/value and calls through already-deployed factories. Factory calls are not interchangeable with direct CREATE: preserve their calldata and do not invent a direct-CREATE address.
- Validate sender, chain, nonce, fee/transaction type, and input fields. Never silently discard unsupported access lists, authorization lists, blob fields, or conflicting transaction fields. Return actionable errors for unsupported forms.
- Preserve Foundry transaction semantics and support its ordinary broadcast/resume workflow. No private keys are passed to Forge, Hardhat, or a WASM guest.
- Reject unsupported signing/unlocking/admin RPC methods. Do not forward a broad arbitrary RPC surface or expose upstream credentials.
- Scope access with an ephemeral credential, enforce local binding/Host/Origin rules, bound payload sizes/concurrency, and clean up the listener on shutdown. Localhost alone is not authorization. The endpoint grants submission access, not approval.
- Approval can outlast an RPC timeout. Keep a durable staged identity and an actionable approval/status path; retrying a submission must not produce a second deployment. Never return an invented transaction hash before actual signing.

### 3. Useful review and recovery

- Use existing exact transaction approval controls initially; avoid introducing broad “anything this script requests” permission.
- Bind approvals to the actual immutable signing payload and consistently persisted review facts. Owner/authority claims that cannot be inferred from arbitrary bytecode must be labeled advisory, not verified.
- Keep results addressable after the client exits: stage ID, transaction hash when available, receipt, created address, failure/approval reason, and recovery instructions.
- Resolve ambiguous broadcast from persisted signed bytes/hash before sending another transaction. A client disconnect must not release a nonce that may already have been broadcast.
- Do not allocate a nonce already used by an in-flight transaction; reject conflicting explicit nonce reuse. Replacement preserves creation/call semantics. Cancellation remains a separately authorized operation.
- Distinguish broadcast acceptance from successful mining and verification. Do not automatically rerun a failed multi-transaction script from the beginning.

### 4. Compatibility and onboarding

- Test one pinned Foundry version end-to-end as a release gate, and record the supported version.
- Test a pinned Hardhat provider-backed script AND a small Ignition module against the same adapter. Keep the same transport implementation; document any unsupported workflow rather than advertising all Hardhat plugins.
- Provide a working multi-contract example with initialization and a second run/resume exercise, clear wallet selection, and instructions for removing private-key reads from existing scripts.
- Document that explorer source verification is a separate optional step; do not require an explorer API key to deploy.

## Architecture and current gaps

The Petal owns application-facing artifact/job/plan/result surfaces. A small host-side compatibility adapter owns the local HTTP listener and uses existing authenticated Machine services. Broker remains the authority boundary and Signer remains the only wallet-key custodian. Build tools remain external to the WASM runtime.

At the inspected baseline, `RawIntentBody::Raw`, `TxIntent`, `StagedTx`, and the Petal outbox record require recipient strings, and `TxEngine::build_unsigned_evm_tx` always constructs `TxKind::Call`. Creation is therefore a platform change, not just a new Petal route. Audit every recipient consumer, especially simulations, policy destination checks, sent-entry scanners, and replacement paths.

Related work must be reconciled rather than duplicated:

- #217 / #214: staging nonce can collide with a broadcast-but-unmined prior transaction; multi-step deployment release gate.
- #206 / #214: owner ceremony must display meaningful transaction facts, not only opaque digests; review release gate.
- #215: Petal-origin authorization must survive staging into the native outbox; no laundering a Petal request into a different approval subject.
- #209: version Petal WIT interface changes properly.
- #216, #218, #213: ceremony expiry/terminal signing/cancellation recovery must not strand jobs or unexpectedly broadcast.
- #171 and #84: composed/batch approvals are future improvements, not a prerequisite to the exact-approval first release.
- #156 / #170 / #169: future Solana/HD-wallet integration context.

## Acceptance criteria

- [x] Direct CREATE stages, simulates, reviews, signs and mines through Bloom for legacy and EIP-1559 encoding; constructor argument and endowment are correct.
- [x] Zero-address calls remain calls; old persisted entries retain behavior; creation cannot accidentally pass a destination allowlist as a transfer to an invented address.
- [x] Foundry deploys two dependent contracts and runs an initialization call with no wallet credential in its environment/files/arguments.
- [x] A pinned Hardhat script and Ignition module deploy through the same adapter with remote accounts.
- [x] Chain/sender mismatch, malformed inputs, conflicting calldata, unknown transaction fields/types, unapproved operations, and unauthorized endpoint access fail closed.
- [x] Exact approval cannot be reused after initcode/value/nonce/fee/payload mutation.
- [x] A nonce regression test holds a prior transaction unmined and proves the next deployment/call uses the correct nonce.
- [ ] Timeout, lost response, expired ceremony, process restart and Foundry resume tests demonstrate no duplicate deployment or hidden broadcast.
- [x] Receipt outcome and created address are persisted and exposed; reverted constructor and failed follow-up call are distinguishable.
- [x] Existing native transfer/call/Petal paths still pass their relevant tests; no unversioned WIT break.
- [ ] A real developer can complete onboarding and a multi-step deployment using documented commands without manually editing Bloom's internal state.

## Explicitly deferred

Solana implementation; a new deployment language; compilers inside the Petal; universal script-plan capture; unattended arbitrary-byte signing; broad reusable deployment permissions; automatic source verification; production multisig governance; account abstraction and chain-specific transaction formats not already supported. CREATE2 through ordinary existing factory calls can work within supported transaction semantics, but custom factory discovery and deterministic deployment orchestration are deferred.

## Delivery checkpoints

1. Implement and verify native contract creation as a reviewable foundational change.
2. Add the scoped adapter and durable RPC submission behavior, integrating existing approval and nonce fixes.
3. Add the deployment Petal surfaces/examples, Foundry and Hardhat compatibility tests, and recovery gates; then validate repeat use before expanding to Solana.

This issue tracks the useful first release. A foundation PR alone must not close it.

## Sources

- Foundry `--unlocked`, resume and slow behavior: https://github.com/foundry-rs/foundry/blob/master/crates/script/src/lib.rs
- Foundry broadcast overloads: https://getfoundry.sh/reference/cheatcodes/start-broadcast
- Hardhat remote accounts: https://hardhat.org/docs/reference/configuration
- Hardhat keystore: https://hardhat.org/docs/guides/configuration-variables
- Ignition execution: https://hardhat.org/ignition/docs/explanations/execution
- Ignition reconciliation: https://hardhat.org/ignition/docs/explanations/reconciliation
- Anchor artifacts/build: https://www.anchor-lang.com/docs/quickstart/local
- Solana loader/deployment: https://solana.com/docs/core/programs/program-deployment
- Solana signed RPC: https://solana.com/docs/rpc/http/sendtransaction
- Agave signer sources: https://github.com/anza-xyz/agave/blob/master/clap-utils/src/keypair.rs
- Pinned Agave v4 implementation: https://github.com/anza-xyz/agave/blob/v3.0.0/cli/src/program_v4.rs

Research performed 2026-09-05. Upstream branch links are mutable; integration tests must pin actual tool versions.

## Tracked work (PM board)

- bloom-directory/bloom#222 — loopback deployment endpoint (closes #221, #206).
- bloom-directory/bloom-broker#36 — semantic review of exact EVM creation/call payloads (required by #222).

0xdewy:
Deployment tools want a local signer, but handing them a wallet key gives the project and the coding agent unlimited signing power. Bloom now offers a private local RPC endpoint that turns each transaction from Foundry, Hardhat or Ignition into a durable, owner-approved outbox operation, so a developer can run an existing deployment script with a Bloom wallet and never put the key in the project, an environment variable or the agent.

Depends on bloom-broker#36, which decodes each transaction for the owner.

## What changed
- `bloom deploy --wallet <w> --chain <c> rpc` prints an ephemeral token URL. Scripts submit unsigned transactions to it; each becomes a `deploy-…` outbox entry with a plan, an approval and, later, a receipt. After the owner approves, `bloom deploy … resume <id>` continues that exact submission. `list` and `status` recover after a timeout or restart.
- Contract creation is a first-class intent through parsing, simulation, policy, signing, replacement and receipts. Legacy and EIP-1559 are supported. Retries return the existing entry; a changed request on a reserved nonce fails closed.
- The endpoint binds to loopback only, rejects browser origins and wrong hosts, requires an explicit nonce, allows a fixed set of read methods, and never offers raw or hash signing.
- Creation on a chain requires the wallet-policy entry `{"chain":"evm-<id>","destination":"exact"}`. On that chain the entry replaces Machine's advisory recipient allowlist with exact owner review for every native transaction.
- Runnable Foundry, Hardhat and Ignition examples with a recovery guide, and a CI job that deploys them on Anvil.

## Not in this PR
Solana deployment, one approval for a whole script, source verification, and custom transaction types. Every EVM path still sends no system claim; the amount and destination shown to the owner come from Broker's decoding in bloom-broker#36, which is where #206 is addressed.

## How to verify
The deployment CI job builds the CLI and deploys the examples through the real HTTP adapter, Machine IPC and a test Broker on Anvil, including Foundry resume and an Ignition rerun that send no duplicate. Unit tests cover parsing, creation encoding, fee bumps that keep the CREATE kind, and nonce recovery. Not run: a live passkey ceremony, a production service restart, or any mainnet transaction. The test Broker simulates the owner.

## Risk and rollout
Two-way wire break with bloom-broker#36: this Machine cannot prepare an EVM approval against an older Broker, and the new Broker refuses an older Machine. Ship both in one release. Every Broker pin (workspace dependency, CI and conformance workflows, release compatibility file and its test fixture) points at that PR's head.

## Reviewer notes
Read the transaction engine's deployment staging, the daemon's deployment RPC and the CLI bridge first. The npm lockfile under the example is the bulk of the diff.

## Checklist
- [x] Architecture docs updated (`docs/architecture/Wallet.md`)
- [x] Sealed Approval invariants respected: exact payload selectors, owner approval per transaction, unchanged Signer custody, recovery from persisted signed bytes
- [x] Agent Documentation updated; no separate Petal README because the native outbox owns this flow

Closes #221


0xdewy:
Merged `master` (`4e37a635`) and advanced the stack pins to keep the recorded combination coherent:

**Pins** (consistent across `Cargo.toml`, `Cargo.lock`, `compatibility-v1.toml`, and the conformance workflow refs/defaults/fallbacks):
- broker: `dab45df` → `af06930e` — current head of bloom-broker#36. Required: master's machine-client/daemon code (BIP-39 custody fields, plural `derivation_requests`, `root_key_ref` optionality merged in from #170) does not compile against the old pin, and #36's branch has since merged broker master forward, so the new pin is a superset of what this PR was verified on.
- signer: `0126ba45` → `941dd376` (signer master, contains the BIP-39/HD custody stack the merged machine code expects); petal: master's landed `73c5b06a`.
- `bin/bloom-worktree pins` on the merge head reports only the expected "broker #36 pin has not landed" blockers.

**Code resolutions**: all master-side for the API-evolution hunks (custody field renames, derived-account selection, projection invariants, CI nextest filter). The PR's EVM-deploy feature code is preserved, including the `native_evm_review` wiring of `evm_review_payloads` (transaction confirm/replace/cancel classes only; Petal approvals send none). Duplicate struct fields produced by the auto-merge were dropped in favor of master's forms; two master tests adapted to the PR's `Option<String>` staged `to`; one master fixture initializer gained `derived: Vec::new()`. `docs/examples.md`: master's post-#220 structure with the Deploying-an-EVM-contract section inserted and paths adapted to `\$BLOOM`.

**Verification on merge head `8c1665ce`** (against the pinned revisions): `cargo fmt` and `cargo clippy --workspace --all-targets -D warnings` clean; `cargo test -p bloom-machine-client -p bloom-tx -p bloom-proto` 52+171+156 passed; `cargo test -p bloom-vfs` 402+14 passed; `cargo test -p bloom-daemon` 99+3 passed; `cargo test -p bloom-it` all Linux-runnable tests pass — the single failure (`macos_upgrade_rollback_handles_the_system_etc_symlink`) fails identically on plain master (macOS-only tooling). The Linux release build and both macOS conformance workflows still need to run on this head before landing, with the explicit Broker/Signer refs recorded above.

0xdewy (commented):
Reviewed the merge head `8c1665ce` (post master-merge, pins advanced to broker #36 head `af06930e` / signer `941dd376` / petal `73c5b06a`):

- **No secret surface.** The deployment RPC adapter answers only public facts (`from` address, chain id, `bloom_deploymentInfo`); foundry/hardhat run `--unlocked` against it, so no private key ever reaches the project toolchain, and `DeploymentTransaction::parse` stages through the ordinary outbox. Ceremony URLs print only to the operator's own terminal, matching the existing confirm flow.
- **CREATE is modeled, not special-cased.** Deploy staging returns `to: None` with nonempty-initcode enforcement; the staged record's optional `to` keeps creation distinct from a zero-address call, and `plan.md` identifies CREATE with the initcode hash and predicted address.
- **Owner review stays exact.** `native_evm_review` forwards signing preimages only for `transaction.confirm`/`replace`/`cancel` provenance, so Broker's decoded EVM review (and its CREATE numeric-chain opt-in enforcement) applies to transaction flows while EIP-712/Petal approvals are untouched.
- **Pins are internally consistent.** Broker build and release pins, the compatibility manifest, and all workflow defaults/fallbacks agree on `af06930e`/`941dd376`; `pins` reports only the expected "broker #36 not landed" blockers.
- Local gates on the head: fmt/clippy clean; machine-client/tx/proto 379, vfs 416, daemon 102, and all Linux-runnable bloom-it tests pass (the one failure is macOS-gated and fails identically on plain master).

Outstanding before landing: the Linux release build and both macOS conformance workflows with the explicit refs above must run on this head. No blocking findings. (Filed as a comment review — GitHub forbids the PR-opening account from approving.)

Follow-up out of scope for this PR: bloom#224 targets this branch and will need its base advanced and its broker pin reconciled after this merge.

0xdewy:
Follow-up to the remaster head `8c1665ce`: its CI's "local integration tests" job failed `bloom-it::deployment_rpc::deployment_tools_and_recovery` deterministically (bare ENOENT spawning `target/debug/bloom`). That archive-only lane never builds the bloom CLI, and the pre-remaster head excluded `binary(deployment_rpc)` from it for exactly that reason; the master-remaster merge dropped the exclusion. The dedicated "Foundry, Hardhat and Ignition deployment" job already builds the CLI and runs this suite, so `aef83f2a` only removes a duplicate, unrunnable copy. CI is green on `aef83f2a` (the earlier "transient Foundry HTTP 500" note in the handoff was this same ENOENT, not transient).

0xdewy:
Merged master (`9dc4039d`, v0.3.0) to clear the conflicts. New head: `19318089`.

**Pins.** All seven conflicts were pinned revisions (three workflows, `Cargo.toml`, `Cargo.lock`, `triad_release.rs`, `compatibility-v1.toml`).
- Broker: `7ee52d531639709b6bfb38af6c5f678c86123101`, the current head of bloom-broker#36. It contains master's Broker pin `dd2add2b`. #36 is still open, so `bloom-worktree pins` reports that this pin has not landed on Broker's default branch. That is expected until #36 merges.
- Signer and service-runtime: master's pins (`ccc9adb3…`, `5db670e1…`), which are also what broker#36 pins.
- `Cargo.lock` is master's lockfile re-resolved for the new Broker pin. No `af06930e`, `941dd376`, `bc88cc67` or `dd2add2b` reference is left.

**One code change.** v0.3.0 removed `ChainSpec::allow_broadcast` (broadcast is always enabled; old config values are ignored). `crates/bloom-it/tests/deployment_rpc.rs` still set it, which merged without a textual conflict but did not compile. The line is deleted, the same way v0.3.0 fixed its own tests.

**Checks on the merged head**
- `cargo metadata --locked`, `cargo check --locked --workspace --all-targets`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo fmt --check`: pass.
- `cargo test --locked -p bloom-it --test triad_release`: 50 passed, 9 failed. The nine are installer-staging tests, not pin checks. They fail identically on the #291 branch, whose hosted CI is green, because the machine I ran them on has a live Bloom listening on `127.0.0.1:20000`, the first port the installer probes. Hosted CI is the reference for those.

0xdewy (commented):
Verified this head (19318089) locally from a detached checkout: fmt clean, strict clippy clean on all six touched crates, unit suites green (machine-client 52, tx 172, proto 156+1i, daemon 99+3, VFS 411+14), and the ignored deployment_rpc live test passes twice — the standard Foundry/Hardhat/Ignition flow plus with BLOOM_TEST_FOUNDRY_PROJECT pointed at multicall-scripting (CREATE2 factory path, idempotent rerun). Also deployed a fresh Forge-built Counter through the real bridge binary + CLI resume: staged, approved, mined, runtime code and constructor arg confirmed on Anvil.

One nit (non-blocking): docs/architecture/Wallet.md:179, crates/bloom-vfs/src/docs/examples.md:87, and examples/evm-deploy/README.md:13 say Machine/Broker protocol 1.5, but this head's compatibility manifest is 1.6/1.6. Suggest a 3-line touch-up.

Noted as still required to land: broker#36 must land first (pins point at its unmerged head), then repin + independent approval.

0xdewy:
Pushed `91139b1`: latest master merged, conflicts resolved, and every recorded
dependency reference advanced. The branch was `DIRTY`; it is now `BLOCKED` on
required review only.

**What landed here.** The eight integration and recovery commits that were
sitting unpublished in a local worktree are now on the PR, plus a merge of
master — which carries the ceremony-port work from #319.

**Signer** moves to the landed `97b9ac7e47ee682e59804d04f6288419f3598a58`,
master's own pin, in all 17 places.

**Broker** is set to `717ece2d2b5ec6caeb62df1a7e8fd07fe27b4716`, the published
head of bloom-broker#36, rather than master's `3352190`. This branch calls the
clear-signing API, which master does not carry until #36 merges. All 16 Broker
references agree: manifest, lockfile, release compatibility manifest, both
macOS conformance workflows, CI and the release test literal.

**Why the reference matters beyond the API.** It was tempting to leave the
Broker at an older clear-signing commit on the grounds that `bloom-broker-api`
was byte-identical, so the seam under test was unchanged. That reasoning is
wrong for the release and conformance references: they select which Broker
gets *built and packaged*, and the approval page assets are compiled into that
binary. A reference naming a Broker without the token artwork would package
approval pages without it, however well the API matched. The named revision is
confirmed to contain `ceremony_assets/tokens/` with both logos, the route that
serves them, and #78's `ceremony_port`.

**Conflicts** were the pin blocks in six files: `Cargo.toml`,
`compatibility-v1.toml`, `triad_release.rs`, `ci.yml` and both macOS
conformance workflows. Master's side was taken throughout, then the Broker
revision set to the candidate above, so nothing master changed was dropped in
order to keep an old pin.

## Verification

| Suite | Result |
| --- | --- |
| `bloom-tx` | 166 passed |
| `bloom-daemon` | 99 passed |
| `bloom-vfs` | 411 passed |
| `cargo fmt --all -- --check` | clean |
| `bloom-it triad_release` | 52 passed, 9 failed |

The nine are the macOS and Linux installer tests. They fail identically on a
plain master-based branch carrying master's own pins — the same nine names,
established by diffing the two failure sets, not by assuming. They are
host-only and unrelated to this change.

Separately, against a running Triad built at the referenced revisions: every
asset the live Broker serves is byte-identical to a detached checkout of
`717ece2`, including both token logos, and the approval page makes zero
off-origin requests.

No owner approval was performed and nothing was signed or broadcast.

0xdewy:
**Status — supersedes all prior progress comments (heads through `91139b1`). Current head: `ef2494b9`.**

**Pins.** Signer is the landed `97b9ac7e47ee682e59804d04f6288419f3598a58` (master's own pin, all 17 places). Broker is `717ece2d2b5ec6caeb62df1a7e8fd07fe27b4716` (bloom-broker#36 head, all 16 references agree) — required not only for the clear-signing API but for the packaged approval assets: the approval page artwork is compiled into the Broker binary, and the named revision is confirmed to contain `ceremony_assets/tokens/` with both logos plus the serving route. Merge conflicts were pin blocks throughout; master's side was taken, then the Broker revision set. **Broker#36 must land first, then repin here + independent approval** (GitHub forbids the opening account from approving its own PR).

**What this PR does (unchanged).** Authenticated loopback RPC stages Foundry/Hardhat/Ignition unsigned transactions in the wallet outbox for owner approval; contract creation via `to: null` with nonempty-initcode enforcement; exact payload review; no secret surface (public facts only, tools run `--unlocked`, ceremony URLs print to the operator's terminal only).

**Head `ef2494b9` (review fixes).** Failed deployment rows restage under current policy instead of replaying as staged; unknown `review_mode` rejected at stage time before nonce/outbox side effects; status reads no longer persist `receipt.json` and never erase a reconciler-decoded revert reason; Wallet.md notes the no-downgrade window for `ContractCreation` rows; 10 new unit tests (mock-RPC staging, RPC validation, bridge guards).

**Verification carried forward.** `bloom-tx` 166, `bloom-daemon` 99, `bloom-vfs` 411 passed; `cargo fmt --check` clean; the 9 `triad_release` installer failures fail identically on plain master (host-only). Against a Triad at the referenced revisions, every served Broker asset is byte-identical to a detached checkout of `717ece2`, with zero off-origin requests. No owner approval performed; nothing signed or broadcast. Prior design assessments stand: CREATE is modeled not special-cased, owner review stays exact, Petal approvals untouched.

**Current CI: red — landing blocker.** `bloom-it::replace_keeps_nonce_and_bumps_fees` fails: first replacement returns `NonceConsumed` where the test expects its own exact approval (`run 35915103404`). Root cause is this PR's own nonce-consumed guard reading the pending-block nonce, so an entry's still-unmined broadcast counts as consumed. Under triage; CI re-run on the current head in progress.

**Remaining gates.** Green CI (blocker above) → broker#36 lands → repin → Linux release build + both macOS conformance workflows at the compatibility revisions → independent approval.

PR #232, opened 2026-09-07
fix(vfs): report a missing path as not found, not as a fault

### LINKED ISSUE #230: Mounted VFS returns EIO and EINVAL where it should return ENOENT, and a node that lies about being a directory
Found while running the wallet benchmark's fixture directly — no model involved, reproducible in a minute on a local Anvil. Three separate things, all in the mounted VFS, all cheap to check.

## 1. A path that does not exist returns "Input/output error"

```
$ cat wallets/w00/chains/anvil/outbox/pending/NOPE/confirm
cat: wallets/w00/chains/anvil/outbox/pending/NOPE/confirm: Input/output error
```

It should be `No such file or directory` (ENOENT). The difference matters more than it looks. "No such file" is a **fact**: the thing is not there, act accordingly. "Input/output error" is a **fault**: something went wrong, try again or try another way.

This is exactly what an agent meets right after a successful discard, because discarding moves the transfer out of `pending/` and the path it just wrote to stops existing. In the benchmark, two agents that had already discarded the correct 7 transfers went on to discard all 12 — they could not confirm the first attempt and widened. One of them spent 374,722 tokens and 31 turns.

## 2. A shallower missing path returns "Invalid argument"

```
$ cat wallets/NOPE
cat: wallets/NOPE: Invalid argument
```

So there are at least three different answers for "this is not here" depending on how deep the path is, and none of them is the right one.

## 3. `wallets/<w>/capabilities` says it is a directory, then refuses to be one

```
$ stat -c %F wallets/w00/capabilities
directory
$ ls wallets/w00/capabilities
ls: reading directory 'wallets/w00/capabilities': Not a directory
```

`stat` reports mode `drwxr-xr-x`, `IO Block: 4096`, type directory. `ls` disagrees.

The practical cost: every `find` over the wallet tree emits one error per wallet.

```
find: 'wallets/w00/capabilities': Not a directory
find: 'wallets/w01/capabilities': Not a directory
... one per wallet
```

`find` is a normal way to explore a filesystem, and this makes it look broken to anyone using it.

## Why these are worth fixing before the next measurement

The whole premise of a filesystem interface is that ordinary tools behave the way people and agents expect. These three are places where it does not, and all three are on the path an agent walks after a write.

The specific worry: an agent that gets `ENOENT` where it currently gets `Input/output error` may well behave differently, because the correct action differs. Any experiment about write acknowledgement measured before this is fixed is partly measuring this instead.

## Reproducing

`scripts/probe_write_flow.py` in `bloom-directory/vfs-research`. Needs a local Anvil and the `bench_machine` binaries; calls no model and costs nothing.

```
anvil --host 127.0.0.1 --port 8545 --silent &
python3 scripts/probe_write_flow.py
```

0xdewy:
## Summary

Closes #230.

Four places in the wallet VFS answered "this is not here" with an error that mounts render as `EIO`, `EINVAL` or `ENOTDIR`. A client reads `EIO` as a server fault worth retrying; `ENOENT` is a fact it can act on.

**Why it matters.** In the 2026-09-06 wallet benchmark, agents asked to discard every policy-denied transfer **got it right on their first command** — the correct 7 of 12, verified against outbox state. Discarding moves a transfer out of `pending/`, so every way of checking afterwards returned `EIO`. Two of them concluded nothing had happened, reran the loop without its filter, and discarded all 12. One spent 374,722 tokens doing it.

| path | was | now |
|---|---|---|
| unknown outbox action id | `Backend("not found: <id>")` → EIO | `NotFound` → ENOENT |
| id in the wrong state | `Backend` → EIO | `NotFound` (the path named does not exist; the entry lives elsewhere) |
| unregistered wallet | `Backend("BACKEND_INVALID_REQUEST: unknown wallet projection")` → EIO | `NotFound` |
| `cat` a bare wallet path | `NotAFile` without checking existence → EISDIR | `NotFound` when absent |
| `wallets/<w>/capabilities` | dir to `lookup`, no `list` arm → ENOTDIR | lists `active.json`, `active.md` |

The first is the striking one: the message already said "not found" and only the variant was wrong. `err_be` flattened every `OutboxError` into `Backend`, directly beneath a comment saying that case "should NotFound".

The `capabilities` mismatch is why every `find` over the wallet tree emitted one error per wallet into whatever was reading the output.

**Verification.** Four tests, one per case, each confirmed to fail against the old mapping before the fix. `cargo test -p bloom-vfs`: 353 passed. `cargo test --workspace`: the 7 failures in `bloom-it triad_release` are macOS installer and release-bundle tests that fail identically on the then-current base — checked on the base rather than assumed. `cargo fmt` and `cargo clippy -p bloom-vfs --all-targets` clean.

**Reproducer:** `scripts/probe_write_flow.py` in `bloom-directory/vfs-research`. Needs a local Anvil and the bench binaries; calls no model and costs nothing.

## Checklist

- [x] Tests added or updated for behavior changes — four tests in `crates/bloom-vfs/src/handlers/wallets.rs`, one per case
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A. No architecture doc states VFS error semantics; the agent-facing guidance is where a caller reads them, and that is updated below.
- [x] Sealed Approval invariants respected — N/A. No signing, grant, PRF or ceremony path is touched; `PermissionDenied` and `OperationNotPermitted` mappings are unchanged.
- [x] Agent Documentation updated — `crates/bloom-vfs/src/docs/agent-guidance.md` gains a "What an error means" section, including that a discarded or confirmed transfer leaves `pending/` so its old path is *supposed* to report not-found, and that re-issuing a broadened write you could not confirm is how a correct action becomes an incorrect one.


0xdewy:
Addressed the actionable review findings in 003cee6.

The wallet projection mapping now requires BackendInvalidRequest plus one of the two exact absence messages emitted for the requested wallet. This keeps unrelated broker faults such as “key not found” as EIO and maps cached tombstones to ENOENT. Added regression coverage for both cases and for reading a registered wallet root.

I did not remove the projection check from a bare-wallet read. When the edge is down and there is no cache, existence is not locally knowable; returning EISDIR would assert that an unknown path exists and restore the bug this PR fixes.

Local verification: cargo test -p bloom-vfs passed 356 unit tests and 8 integration tests; cargo fmt -p bloom-vfs --check, cargo clippy -p bloom-vfs --all-targets, and git diff --check passed.

0xdewy:

GavinPacini (approved):

PR #234, opened 2026-09-07
Carry native SOL approval across refresh and budget Petal sessions

### LINKED ISSUE #231: Native Solana transfers cannot be approved: an exact approval pins a message whose blockhash expires during the ceremony
## Summary

A native Solana transfer can never be approved and broadcast reliably, because the
owner's passkey ceremony and the blockhash lifetime are in a race that the
ceremony usually loses.

`sign_exact_payload` builds `ApprovalSelector::Exact { ordered_payload_digests, .. }`,
which pins the approval to one exact message. A Solana message embeds a recent
blockhash valid for ~150 slots (~60s). The sequence is:

1. `outbox/new.tx` stages a transfer and builds message `M1` with blockhash `B1`.
2. `outbox/pending/<id>/confirm` asks Broker to sign; Broker has no approval and
   returns `ApprovalRequired` with a ceremony URL.
3. The owner completes the passkey ceremony. On this machine that means a
   provider prompt, then a phone, then a QR scan — tens of seconds.
4. `B1` has now expired. The Machine rebuilds with `B2`, producing `M2`.
5. The approval covers `M1`, not `M2`.

So the owner must scan a QR inside a ~60 second window or the operation is dead.

## Observed failure sequence

Live triad, `bloom-broker` at `47189f8`, canary-capable Machine, mainnet-beta.
Each retry fails differently because the previous attempt leaves lineage behind:

```
blockhash_expired                                             (staged entry swept)
OPERATION_ID_CONFLICT: ceremony operation ID was reused with different stable input
QUOTA_EXCEEDED: wallet already has a live ceremony
CLAIM_INVALID: approval is not active
```

The `QUOTA_EXCEEDED` case is bloom-directory/bloom-broker#39 reproduced: a
`confirm` created a `sealed_approval` ceremony, the write errored before
returning its URL, and the wallet was left locked behind a ceremony with no
reachable URL. It had to be recovered by reading `ceremony.state_persisted` out
of the Broker log and calling `ceremony status` on the operation id.

## Root cause

An exact approval binds bytes that are guaranteed to go stale on a timer shorter
than a human interaction. No amount of retrying fixes this; the approval and the
broadcast must both fall inside one blockhash window.

## The machinery already exists

`bloom-daemon/src/lib.rs:252-268` already computes a blockhash-normalized Solana
message template digest for the Pump.fun canary: it parses the message header,
derives the blockhash offset from the account count, zeroes those 32 bytes, and
hashes with domain separator `bloom-solana-message-template/v1`. It is exposed as
`message_template_sha256` and enforced in `bloom-proto/src/pumpfun_canary.rs`.

The Pump.fun Petal already relies on the same property, documented in its README:
an approval may be reused "only when the independently parsed transaction
template is unchanged and the recent blockhash is the sole changed message
field. Any other change is rejected."

The native CLI transfer path does not use any of it.

## Proposed fix

Let a Solana exact approval authorize the blockhash-normalized template rather
than one frozen message, so a restage with a fresh blockhash reuses the approval
that the owner already granted.

- Carry the template digest in the approval terms for Solana exact approvals.
- In Broker authority, accept a payload whose template digest matches the
  approved one and whose only differing message bytes are the blockhash span.
  Every other field — fee payer, destination, lamports, program ids, account
  keys — must still match exactly.
- On approval activation, restage with a fresh blockhash and broadcast without a
  second human step (this is root cause 7 of the Pump.fun handoff).
- Reset stale approval/signing lineage on terminal Broker states so a dead
  operation stops poisoning the next attempt (#218 covers this half).

## Acceptance

- An owner can take minutes over the passkey ceremony and the transfer still
  broadcasts.
- A message differing from the approved template in any byte outside the
  blockhash span is refused.
- A second transfer with the same stable inputs but a different payload does not
  collide on the ceremony operation id.
- A `confirm` that creates a ceremony always returns its URL, or the ceremony is
  not created.
- Tests cover: expiry during the ceremony, restage after approval, a tampered
  non-blockhash byte, and repeated staging of identical intents.

## Scope

Broker (approval terms and authority), Machine (`exact_signing`, the Solana
outbox confirm path). This widens what one approval covers, so it wants its own
review rather than riding along inside a feature branch.

0xdewy:
A native SOL transfer now carries one owner approval across a blockhash refresh, while preserving single-use authorization and the exact reviewed amount, destination, account, chain, and fee. This also retains the independent Petal session-budget and key-setup fixes that had accumulated on the branch.

## Why

A Solana message embeds a recent blockhash whose validity can expire while an owner completes a browser passkey ceremony. Master can restage expired bytes, but the replacement needs another Exact approval; it cannot meet the “approve once, take minutes, settle” flow tracked in #231.

The old branch also contained independent Petal fixes that must not disappear with the selector decision: reusable approval budgets, address-before-approval setup ordering, developer fee provenance, scope-expiry reconciliation, and safer failure classification.

## What changed

- Rebuilds an approved native transfer with a fresh blockhash and last-valid height, then carries the same single-use approval to the replacement only when its stable reviewed intent is unchanged. Broker #55 independently verifies the exact rebuilt payload.
- Keeps block-height freshness separate from authorization: signing and broadcast still refuse a stale replacement.
- Migrates approval lineage and the replacement-specific challenge before retiring the predecessor. Operation locks make refresh, signing, broadcast, cancellation, and repeated restage race-safe.
- Records the exact signed transaction before network submission. Ambiguous signing keeps the approval; ambiguous broadcast remains owned by signature reconciliation and cannot trigger a second payment.
- Preserves ordinary Exact behavior, including master’s claim-derived value limits. The refresh selector is used only for the verified native SOL transfer.
- Persists explicit reusable Petal approval budgets and rejects retries that change them. Derivation first publishes the scoped key/address as `key_derived`; only a later identical request prepares the approval against the owner’s updated policy.
- Retains derived-key scope expiry, expired-ceremony restaging, authoritative key/address reconciliation, and precise terminal-versus-retryable denial handling.
- Declares native developer-fee provenance for the fixed seven `pumpfun.*` operation classes in the development harness.
- Restores the wallet-export custody match arm lost during master reconciliation.
- Fixes macOS rollback extraction so GNU tar does not accidentally exclude canonical `private/etc/...` entries.
- Pins Broker #55 (`0c29e057e3e13186b704beb3324fe9e3f371e9ce`) and Signer #28 (`c1cf78270815f27f39a81e913e2cebf7828832f2`) consistently in manifests, lockfile, release compatibility data, tests, and workflow refs.

## Not in this PR

- Durable nonce accounts or a general Solana transaction normalizer.
- SPL token, Pump.fun, Jupiter, multi-instruction, or priority-fee support under the refresh selector.
- Live Pump.fun execution. The session-budget/provenance changes cover the authority path, not venue behavior.

## How to verify

- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo test -p bloom-solana-tx --all-features --locked`: 20 unit, 26 lifecycle, 16 outbox, 10 reconciliation, and 4 signing tests pass; the local-validator test remains ignored without Agave.
- [CI run 34894418394](https://github.com/bloom-directory/bloom/actions/runs/34894418394) passes formatting, Clippy, doctests, unit tests, local integration and triad packaging, external E2E, Linux W0 isolation, and the locked release build.
- [Agave run 34894417961](https://github.com/bloom-directory/bloom/actions/runs/34894417961) passes the transaction lifecycle, daemon/VFS, and multi-account refresh E2Es against pinned Agave 3.0.0. The tests follow a refreshed successor, compare every approval-bound fact, and verify the exact persisted signing bytes.
- `bash tests/packaging/macos-upgrade-rollback.sh` and its `macos_launchagent` regression pass.
- `bin/bloom-worktree pins` confirms all ten dependency groups agree and reports only the expected pre-landing status for the unmerged Broker/Signer commits.
- [Exact-ref macOS run 34894456895](https://github.com/bloom-directory/bloom/actions/runs/34894456895) builds and uploads the signed candidate, then fails its disposable-host gate when `mount_nfs` times out mounting the already-listening loopback NFSv4.1 server. The identical failure predates this candidate in run 34639530054; it is recorded as an unresolved release-conformance obstruction, not treated as passed.
- Not run here: a fresh real-owner mainnet ceremony and settlement. The local canonical ceremony port is currently owned by another session and must not be taken over.

## Risk and rollout

This changes the Machine–Broker–Signer authorization combination and persisted Petal key-request state. Land Signer #28, then Broker #55, then repin this PR to both landed squash commits and rerun the release gates. Downgrading to v0.2.1 requires removing and recreating Petal key-request state; that version cannot decode the new state.

Do not deploy Broker #55 with unchanged master Machine: Broker intentionally rejects a native system claim under Exact.

## Reviewer notes

Review the Solana engine/outbox lifecycle and Machine client first. The daemon’s Petal budget/setup state is a separate substantive unit. Dependency, compatibility, workflow, and fixture changes are integration mechanics.

## Checklist

- [x] Architecture docs updated for the Petal state/lifecycle behavior.
- [x] Sealed Approval invariants preserved: only freshness facts change, one verified native transfer remains one operation and one signature, and execution uses sealed bytes.
- [x] Agent documentation updated for restage replacement discovery and challenge review.

Closes #231


0xdewy:

0xdewy:
Regenerated the lockfile after the manifest pin moved. Cargo.lock was the only changed file, with 3 insertions and 3 deletions. The full workspace all-targets cargo check with the locked dependency graph passed.

0xdewy:
Two follow-ups from the ceremony-stack review are pushed: `3780dc58` and `6b789ada`.

**`3780dc58`: approved refresh.**
- `sign()` now takes the entry's operation lock.
- On the approved-refresh path, `restage()` leaves an already-signed entry alone. The lock alone would have left a window between `sign()` and `broadcast()` in which a second confirm could retire the entry and delete its only signature.
- A repeated restage of an already-retired id no longer overwrites the replacement's newer attempt record.
- The replacement's approval challenge is rebuilt with its own ids instead of copied. The copy named the retired `action_id`, so it failed the check that agent guidance asks for before opening the ceremony.
- The runbook path for `restage_advice.json` is `failed/<id>/`, not `pending/<id>/`.

**`6b789ada`: budget refusals and missing tests.**
- The five `LIMIT_EXCEEDED_*` / `SIGNER_RATE_BACKSTOP_DENIED` codes (Broker contract `Never` + `ReservationReleased`) are now final in both the Petal host classification and the native engine. Before, Petals saw them as backend faults and Pump.fun replayed them indefinitely.
- New tests:
  - a live approval inside its one-hour window cannot sign past the blockhash height;
  - the boundary at the last valid height;
  - no simulate or attempt marker on a refused broadcast;
  - a changed stored Petal address is denied.

Every new test fails when its fix is removed. The four Solana validator E2Es pass on a fresh ledger. Details are in the updated description under How to verify.

M3 is not acted on: both revisions are unreleased, so under the release-compat rule it is a transition. Pins are unchanged until broker#55 and signer#28 land.

GavinPacini:
I think we should close in favour of solving via: https://github.com/bloom-directory/bloom/issues/259

PR #54, opened 2026-07-02
[codex] Harden Bloom status VFS for sandbox exposure

josh-richardson:
## Summary

- remove host home and raw audit tail exposure from the mounted status VFS
- redact URL userinfo and broader secret-like query parameters in status output
- add regression tests for sandbox-safe status exposure and redaction

## Validation

- `cargo test --package bloom-vfs --manifest-path Cargo.toml redacts -- --nocapture`
- `cargo test --package bloom-vfs --manifest-path Cargo.toml status_does_not_expose_home_or_raw_audit_tail -- --nocapture`
- deployed on iris as part of the Bloom Kubernetes filesystem integration and verified from Centaur sandboxes

- [x] Ready

PR #238, opened 2026-09-08
Provide both loopback ceremony listeners the Broker now requires

0xdewy:
Chromium resolves `localhost` to `::1` before `127.0.0.1`, but the ceremony listener — the loopback listener serving the browser approval page on port 18734 — was published on one loopback family only, so browser approvals failed about half the time on Linux. The Broker now requires two inherited listeners and refuses to bind either itself (bloom-directory/bloom-broker#54); this repo's units and developer launcher still provided a single IPv4 listener, which a release built on that Broker would reject at startup.

## Why

The Broker mints a ceremony URL pointing at `localhost`. Only `127.0.0.1` had a listener, so when the browser picked `::1` nothing answered — it presented as a dead approval page rather than a bind error. The fix lives in the Broker (bloom-directory/bloom-broker#54, open; supersedes bloom-directory/bloom-broker#38): it consumes two inherited listeners named `broker-ceremony-ipv4` and `broker-ceremony-ipv6` and exits if either is missing. This repository owns the units that hand the Broker its listeners, and they still named a single IPv4 descriptor `broker-ceremony`.

## What changed

- Linux production packaging publishes both loopback families on port 18734, each through its own socket unit: `bloom-broker-ceremony@.socket` binds `127.0.0.1` as `broker-ceremony-ipv4`, and the new `bloom-broker-ceremony-ipv6@.socket` binds `[::1]` as `broker-ceremony-ipv6`. One unit per family is a systemd constraint, not a preference: `FileDescriptorName=` names every fd a unit passes, so a unit with two `ListenStream=` lines hands the service two fds under one name and the Broker's activation library rejects the duplicate before either listener can be used. The Broker service unit now requires, orders after, and takes sockets from both units, and passes both activation names.
- The developer launcher writes one single-listener socket unit per family, so a dev launch exercises the same per-unit activation as production.
- Machine's Broker startup diagnostic accepts the acquisition failure the new Broker reports. `read_broker_startup_failure` now pairs each incident with the exact address it may carry: `ceremony_listeners_unavailable` at `localhost:18734` alongside the two existing `127.0.0.1:18734` incidents. The address is validated against the incident rather than against one hardcoded literal, so neither `[::1]:18734` nor an attacker-chosen host is accepted for any of them. On macOS the Broker binds both families itself; the isolation spec (section 7) now states the two exact binds and their socket options, and the conformance scripts and packaging guide were updated to the new incident: a failure to acquire a listener is no longer reported as a foreign or unverifiable process, because an unavailable IPv6 stack is not evidence that another process owns the port.
- Linux upgrades restore the previous release's units when they roll back. Binaries are versioned in a digest-named directory behind a `current` symlink, but the unit templates are installed at fixed paths, and rollback only moved the symlink back — so an upgrade that failed after the unit writes restarted the pre-#54 Broker under the `broker-ceremony-ipv4`/`-ipv6` units, which it cannot start with. The upgrade transaction (`bloom.linux-upgrade-transaction.2`) now snapshots every Bloom-owned unit template it overwrites or removes, with its mode and explicit absence, before the first write. Rollback restores them — removing the IPv6 unit a pre-#54 release never had — and starts nothing unless stopping, restoring, reselecting the release, and rewriting its metadata all succeed; otherwise the transaction is kept for a retry. An interrupted upgrade is always rolled back rather than resumed, because the unit writes are not atomic with the binary switch and resuming could start one release's binaries under another's units.
- A schema-1 transaction, the only interrupted state an already-deployed installer can leave, carries no snapshot. It is rolled back the way that installer would have — binary selection and metadata only — and the retried installation then rewrites every unit. Refusing it strands the host: the interruption has already marked the enrollment records `activating`, so every later run is rejected by `validate_linux_release_set` before anything can rewrite them. A schema-2 transaction missing its snapshot still fails closed.
- Operator docs: the Linux packaging guide names both socket units and states the IPv6-loopback prerequisite (`net.ipv6.conf.lo.disable_ipv6=1` is unsupported); the local-mainnet prerequisites require port 18734 free on both families and IPv6 loopback enabled.

## Not in this PR

- The Broker-side change itself: bloom-directory/bloom-broker#54 — now merged (squash `735218ce93869b294770927ccf7f1f3170678325`).
- Repin to that landed squash: prepared on branch `pr238-pin-broker-54` (commit `1b39395a`, based on this head) — every build and release pin location advanced (`Cargo.toml`, `Cargo.lock`, `compatibility-v1.toml`, the `triad_release` pin literal, `broker_ref` in `ci.yml`, and the defaults/fallbacks in both macOS conformance workflows), plus the `CustodyPrepareRequest::derivation_request` → Vec-valued `derivation_requests` call-site adaptation that revision requires. At those pins, `cargo test -p bloom-it --test triad_release` passes 50/50 on Linux (same 5 macOS-only failures) and `packaging/triad/release.sh build linux` passes end to end (Machine `1b39395a`, Broker `735218ce`, Signer `9abe7917`). Integrate before merge per issue #241 step 7.

## How to verify

- `cargo test -p bloom-it --test linux_packaging` — 7 tests passed, including a new guard that fails any `*.socket` in the packaging tree carrying more than one `ListenStream=` or `FileDescriptorName=`; that guard is the assertion that would have caught this bug. Each ceremony unit is asserted to publish exactly its own family and descriptor, and the service unit to wire both units through `Requires=`, `After=`, and `Sockets=`.
- `cargo test -p bloom-it --test triad_release` — 50 tests passed; the 5 `macos_*` failures require macOS and are skipped in CI. Launcher assertions now check one single-listener writer called twice and both unit variables; installer assertions cover the new unit at every stop/manifest/install/disable/uninstall site.
- Upgrade rollback and recovery, in `triad_release`: `linux_upgrade_failure_restores_the_previous_unit_contract_before_restart` stages the real pre-#54 layout and forces a failure after the unit writes; `linux_interrupted_upgrade_recovery_restores_before_any_restart` interrupts at each unit-write boundary through a harness that sources the installer and fingerprints the units every start attempt observes; `linux_upgrade_recovery_failure_preserves_the_transaction_and_retries` corrupts the snapshot six ways and fails the health gate, and each must fail closed without starting anything, keep the transaction, and succeed on retry once repaired; `linux_upgrade_recovery_shares_templates_and_rejects_foreign_transactions` covers two active logins plus retained custody, unknown schemas, and a schema-2 transaction without its snapshot; `linux_legacy_interrupted_upgrade_recovers_without_operator_intervention` stages a schema-1 interruption with `activating` records and asserts it completes unattended. Refusing that transaction and following the removal instruction the installer printed failed the next run with "installed Linux enrollment is not active".
- `cargo clippy -p bloom-it --all-targets -- -D warnings` and `cargo fmt -p bloom-it -- --check` — clean; `bash -n packaging/triad/release/install-linux.sh` — clean.
- Why CI never flagged the gap: the old assertion `FileDescriptorName=broker-ceremony` is a prefix of `broker-ceremony-ipv4`, so it kept passing after the Broker renamed its descriptors; a substring check cannot catch that regression.
- Live check on systemd 261: with two units, a socket-activated process receives `LISTEN_FDS=2` and `LISTEN_FDNAMES=broker-ceremony-ipv4:broker-ceremony-ipv6`; with the previous single two-listener unit, systemd hands out `broker-ceremony-ipv6:broker-ceremony-ipv6`, which the activation library rejects ("activation socket name ... is duplicated").
- Not run: an end-to-end launcher run to an HTTP response from both families. The developer root on the test host carries enrollment state rendered for an older binary trio; the Broker exits in an early startup stage that reports no detail. The fd-name capture above is the evidence for the activation mechanism itself.
- Live systemd, in disposable containers with their own PID 1 and network namespace: production binaries from `release.sh build linux` at the compatibility pins, release A installed from `4ce4dc88` (one IPv4 listener), then an upgrade to two-listener units whose pinned pre-#54 Broker cannot start, so it fails at the health gate after every unit write. With this head's installer the rollback restored `broker-ceremony`, removed the IPv6 unit, and cleared the transaction; the Broker came back healthy and answered on `127.0.0.1:18734`, and again after a reboot. With the installer from `87c70a10` the same upgrade left the two-listener units in place, the Broker crash-looped, nothing answered on the ceremony port, and it stayed that way after a reboot.
- Not run: a successful upgrade to a candidate carrying bloom-broker#54. #54 pins Signer API `104cf218`, which is only on the unlanded Signer stack, so no landed candidate triple exists to install yet.
- Not run: the macOS conformance jobs (W0), which are skipped without a Mac. The updated `run-disposable.sh` and `run-two-login.sh` assertions have not been executed.
- Disposable Linux systemd VM (qemu, ubuntu 24.04, systemd 255, no KVM): real signed payloads through the live `ROOT=/` installer, old release = bloom `59eacb11` packaging + Broker `cd1e31e9` (pre-#54) + Signer `9abe7917`, candidate = bloom `2c36e054` (this fix) + Broker `7d6a3bf6` (#54 head; the landed squash `735218ce` carries the same tree). Authenticated `bloom serve triad-health-check` exit=0 at every stable state: fresh old install (v4 answers, v6 refused, no IPv6 unit on disk), candidate upgrade (both `127.0.0.1:18734` and `[::1]:18734` answer), forced failure after the unit writes (rollback restored the old units, removed the IPv6 unit, restarted healthy, transaction cleared), and interrupted-transaction recovery (schema-2 transaction with partial unit writes recovered by restoring the coherent old installation first, then the retried install landed the candidate). Evidence: `shared/2026-09-10-pr238-unit-restore-handoff.md`.
- macOS conformance (released-combination gate): neither dispatched run passes yet, both for reasons outside this change. `macos-unix-conformance` on this branch ([run 34558989551](https://github.com/bloom-directory/bloom/actions/runs/34558989551)) fails at "W0 payload has the wrong platform claim" because this base predates #244's W0-claim build step; on a verification ref carrying that CI fix ([run 34560836152](https://github.com/bloom-directory/bloom/actions/runs/34560836152), `restore-linux-upgrade-units-w0` @ `ebd0e426`) it fails at full activation with the Machine daemon's IPC socket not responding — the same terminal failure, same step, as an unrelated branch earlier that day; the workflow has no recorded green run and needs its own investigation. `macos-two-login-conformance` is pending until the disposable runner's owner supplies the two GUI login UID/name pairs. A skipped, queued, or failed conformance run is not a pass; these must land green before release.

## Risk and rollout

- This and broker#54 must land as one coordinated change. Both mismatched pairs fail Broker startup on Linux, closed:
  - New units + old Broker: the Broker at the current release compatibility pin reads `BLOOM_BROKER_CEREMONY_ACTIVATION_NAME`, defaults to `broker-ceremony`, and exits when no listener with that name was handed to it. A release pairing this packaging with that pin would not start.
  - Old units + new Broker: broker#54 requires both named listeners and refuses to bind either itself.
- Merge gate: this lands together with broker#54 per issue #241 step 7, which also advances the broker pin to that change's landing commit in every location the pin table lists. Do not merge either side alone.
- Rollback: revert this PR and move the broker pin back in the same change.

## Reviewer notes

- Substantive: the two socket units, the service unit's `Requires=`/`After=`/`Sockets=` lines, the launcher's single-listener writer called twice, the incident/address pairing in `read_broker_startup_failure`, and the installer's upgrade transaction — `snapshot_linux_upgrade_units`, `restore_linux_upgrade_units`, the gated `rollback_linux_upgrade`, and `recover_interrupted_linux_upgrade`. The installer diff mirrors the new unit at every site that names the IPv4 unit (9 paired references each side). Tests assert unit, launcher, and installer content as strings and move with the change; the remaining files are documentation.

## Checklist

- [x] Tests added or updated for behavior changes — `linux_packaging` now asserts per-unit publication and guards every `*.socket` against a second `ListenStream=` or `FileDescriptorName=`; `triad_release` covers upgrade rollback and interrupted-upgrade recovery against the pre-#54 unit layout
- [x] Architecture docs — N/A, packaging only
- [x] Sealed Approval invariants respected — N/A, no approval path touched
- [x] Agent Documentation updated — N/A, no agent-visible surface

Landing order: bloom-directory/bloom#241


0xdewy:
Cleanup pushed on top of this branch: the dead `write_linux_socket_unit` helper is deleted (its only caller was replaced by the loopback writer), and the `BLOOM_TRIAD_DEV_CEREMONY_PORT` knob is removed from the launcher and from the release test — nothing on this stack reads it. The canonical port is hard-coded again, as on the base launcher.

Verification on this head:
- `bash -n scripts/triad-dev-launch.sh` — clean
- `cargo test -p bloom-it --test triad_release` — 44 passed, 5 failed (the known `macos_*` tests that cannot run on Linux; skipped in CI)
- `cargo test -p bloom-it --test linux_packaging` — 6 passed, 0 failed

GavinPacini:
Reviewed `87c70a1083803b35587931f45715c0d68b929758` against `feat/solana-native`. One P2 finding; no additional P1 findings.

**[P2] Roll back the socket/service unit contract together with the Broker binary.** [install-linux.sh:1182–1189](https://github.com/bloom-directory/bloom/blob/87c70a1083803b35587931f45715c0d68b929758/packaging/triad/release/install-linux.sh#L1182) replaces the existing IPv4 descriptor name and installs the IPv6 socket; the service template is replaced at lines 1211–1216. However, `rollback_linux_upgrade` (lines 750–760) restores only the binary symlink and release metadata/config digests before restarting services. Neither the transaction nor the versioned release directory saves these unit files.

If an upgrade from the pre-#54 Broker fails after those writes (for example, a later configuration validation or health check fails), rollback starts the old Broker with the new `broker-ceremony-ipv4` / `broker-ceremony-ipv6` units. That Broker asks for `broker-ceremony`, so it cannot start. Coordinating the new release with Broker#54 does not fix this failed-upgrade path.

I reproduced the file-state mismatch in a staged harness using the actual `begin_linux_upgrade`, `atomic_install`, `switch_linux_release`, and `rollback_linux_upgrade` functions: the current release returned to the old binary directory, but `FileDescriptorName=broker-ceremony-ipv4` and the two-socket `Sockets=` line remained. No live systemd services were started. On a live installation, the rollback health check would consequently fail rather than restore the working release.

Please snapshot/version and restore the matching unit files, including removing the IPv6 unit when rolling back to the old layout, before daemon-reload/restart. Add a regression that begins with the single-listener installation and forces failure after the new units are written; cover interrupted-upgrade recovery too.

Product context: this PR supplies the OS-managed listeners needed for wallet approval pages to work when localhost connects over either IP family. Its startup-status support also gives users an actionable failure instead of an unavailable approval flow. The rollback issue matters because an unsuccessful update must leave users able to access and approve wallet operations on the previously working release.

Other review notes: the listener names and service wiring match Broker#54; each listener is bound to its exact loopback address, and startup-status parsing retains exact incident/address/message validation. The one-unit-per-name arrangement matches [systemd's descriptor naming contract](https://raw.githubusercontent.com/systemd/systemd/main/man/systemd.socket.xml). The documented Broker pin update remains a release prerequisite. The PR description's PF-rule claim is stale: the current diff has no PF rule change after the base retired that mechanism.

Validation: shell syntax checks pass; 7 Linux packaging tests and 21 macOS packaging tests pass locally. The triad-release suite passes 48 tests; 3 Linux installer tests cannot complete on this macOS host because BSD `mv` lacks `-T`. GitHub's Linux checks pass, while privileged macOS W0 is skipped. No live systemd or privileged multi-login conformance run was performed. GitHub reports a clean merge into the feature-branch target, but I recommend fixing the rollback finding before release.

0xdewy:
Thanks — agreed, and fixed in `2c36e054` + `275d07c8`. The upgrade transaction (now `bloom.linux-upgrade-transaction.2`) snapshots every Bloom-owned unit template it overwrites or removes — bytes, modes, and explicit absence — before the first write. Rollback restores them, including removing the IPv6 unit a pre-#54 install never had, and starts nothing unless stop, restore, release reselection, and metadata rewrite all succeed; otherwise the transaction is kept for retry. Interrupted upgrades are now always rolled back rather than resumed, since the unit writes aren't atomic with the binary switch. A schema-1 transaction from an already-deployed installer has no snapshot and is rolled back the way that installer would have; refusing it strands the host, because the interruption has already marked enrollments `activating`.

Verified live under systemd in disposable containers with production binaries at the compatibility pins: install the single-listener release, then upgrade to the two-listener units with the pinned pre-#54 Broker, which fails at the health gate after every unit write. With this head the rollback comes back healthy on `127.0.0.1:18734`, including after a reboot; with the installer from `87c70a10` the Broker crash-loops and stays down after a reboot — your scenario reproduced. Not yet exercised: a successful upgrade to a #54 candidate, which needs #54's Signer pin to land first. PF note also fixed — the description no longer claims a PF change.

Issue #60, opened 2026-07-04
[refactor]: tear out chain related code and tests until they're needed


GavinPacini:
We should remove features and code related to a bloom blockchain / bloom chain for now. Focus on core walletFS product, with Petals serving as wallet extensions only, not generic / onchain programs. 

We will remove a lot of currently unused code and long-running tests if we do this.

AC: 
- Core walletFS features continue to work: EVM Petal, Hyperliquid Petal, Polymarket Petal, Paid HTTP requests Petal, VFS, serve / mounting VFS, keystore, CLI, daemon
- Everything in `docs/architecture` is still in the codebase, these documents are live and current and focused on walletFS
- Release binary is smaller
- Fewer LOC across the repo 
- No `bloom-chain*` crates
- No bloom programable transaction block (PTB) types or codepaths
- No "smart-contract" like / onchain petals e.g. the DEX example
- No bloom chain related tests or onchain petal related tests
- Any mention of a bloom blockchain or chain is gone

PR #39, opened 2026-06-09
ci: remove obsolete acceptance workflow

violet-agent:
## Summary
- Remove the standalone `acceptance` GitHub Actions workflow.
- This workflow currently only fires on `push` to `master` plus manual `workflow_dispatch`, so it does not protect PRs before merge.
- Its two jobs have been failing on `master` and are not configured as required status checks.

## Why
Recent `master` run `27179826569` failed in both jobs:
- `cargo test -- --ignored`: invokes `cargo nextest run --workspace --ignored --no-fail-fast`; current nextest rejects `--ignored` and wants `--run-ignored <which>`.
- `docker petal DEX acceptance`: the host runner lacks the NFS mount helper/privileges needed by `bloom serve --mount` (`mount.nfs4: command not found`, then `must be superuser to use mount`).

The maintained `CI` workflow already runs on PRs and `master`, including the normal workspace tests, chain subprocess test, docker petal VFS acceptance, and real-wasm invariant gates.

## Test Plan
- Inspected recent GitHub Actions runs/logs for `acceptance` on `master`.
- Verified branch protection does not require acceptance checks (`GET /branches/master/protection` returned 404 / no protection).
- Verified this PR only deletes `.github/workflows/acceptance.yml`.

Issue #33, opened 2026-06-03
Track remaining test runtime optimizations


josh-richardson:
## Context

A test-runtime review found several high-impact optimizations that are separate from the initial nextest / WASM-cache / nested-build cleanup.

## Remaining work

- Add a fast default workspace scope, likely via `default-members`, so root-level local runs do not compile every E2E/example target by default. Keep an explicit full workspace lane for nightly or pre-merge coverage.
- Gate default-running subprocess and live-network tests. Known candidates include `crates/bloom-it/tests/chain_testnet_provision.rs`, the Anvil-backed simulate tests in `crates/bloom-vfs/src/handlers/simulate.rs`, and live mempool provider tests.
- Split and speed up Docker acceptance flows. Add build-stamp or `--no-build` support, avoid rebuilding the mempool mock for non-mempool modes, prebuild/reuse `mount_demo`, and collect deep diagnostics only on failure.
- Reduce fixed sleeps and serial polling in integration tests. Prefer receipt/event polling, configurable test intervals, concurrent validator waits, and test hooks where production cadences are intentionally long.
- Collapse overlapping DEX coverage where pure math/router/pool behavior is asserted repeatedly across unit, real-WASM, and Docker acceptance tests.

## Notes

The first follow-up PR handles:

- removing the nested `cargo build` from `chain_smoke`,
- caching/resolving petal DEX WASM artifacts in the test harness,
- switching CI test execution to `cargo nextest` while retaining doctests.

GavinPacini:
Closing because we're not working on chain rn.

Issue #32, opened 2026-06-03
Proposal builder should not silently fall back to empty blocks after execution failure


violet-agent:
## Problem

The proposal builder masks selected-transaction execution failures by falling back to an empty block.

Current behavior:

```rust
// crates/bloom-chain-node/src/node.rs:1449-1458
match finalize_proposal_block(template, base_state, executor, block_emission, accepted) {
    Ok(block) => block,
    Err(e) => {
        warn!(
            err = %e,
            "proposal.builder fell back to empty block after filtered tx set failed"
        );
        finalize_proposal_block(template, base_state, executor, block_emission, Vec::new())
            .expect("empty proposal block must execute")
    }
}
```

## Why it matters

Failure at this point may indicate a deterministic executor bug, bad gas/fuel accounting, incorrect filtering, or state/index corruption. Silently proposing empty blocks can hide the root cause and can degrade into a chain that keeps producing empty blocks while valid-looking transactions remain in the mempool.

## Suggested fix

Classify the failure before falling back.

- If it is specific user-transaction invalidity, quarantine/drop the offending transaction(s) with structured reason codes and metrics.
- For executor/system/consensus failures, fail proposal construction loudly instead of masking with an empty block.

## Evidence

Confirmed on current `origin/master` at `crates/bloom-chain-node/src/node.rs:1449-1458`.

GavinPacini:
Closing because we're not working on chain rn.

PR #248, opened 2026-09-09
Mount numbered HD accounts: reads, spends, creation, and account-scoped Petal dispatch

0xdewy:
## Summary

Implement the [HD accounts design (revision 4)](https://gist.github.com/0xdewy/1d875335e19bcfd5ed12193966419aae): mount numbered accounts at `wallets/<wallet>/<n>/`, stage and confirm through them, create accounts through one Signer-numbered ceremony that always creates both families, and dispatch installed Petals under `wallets/<w>/<n>/petals/` with host-trusted account context that the daemon validates against fresh Broker membership.

### Numbered accounts (reads and spends)

- `wallets/<w>/<n>/account.json` renders both families (path, address, fingerprint, lifecycle, freshness) from the wallet projection's **cached authenticated account inventory**; listings, stats and reads carry no authority side effects, and a stale projection is marked as such.
- `wallets/<w>/<n>/chains/<chain>/...` reads and the full outbox surface (`new.tx`, `confirm`, `confirm.override`, `cancel`, `replace`, `restage`) work through the account's key for the chain's family. Pending controls are fenced to the staging key, so account 1 never sees or confirms account 0's entry.
- Signing resolves the key from the staged identity: EVM signs with the active child whose address is the staged `from`, Solana pins the account's fingerprint at stage and re-resolves that exact child at confirm/restage. The wallet-level `wallets/<w>/chains/<chain>/outbox/` is account 0's outbox, fenced exactly like a numbered account (see "Wallet-level outbox fence" below).
- `accounts.json` entries carry `number` (paths outside the default mapping render `number: null`). Legacy/imported wallets are account 0 from their projection.
- A wallet Broker cannot characterise — no root key and no active derived key (every child retired), or legacy BIP-32 custody — still mounts. Its numbered tree is empty and `accounts.json` carries the refusal as `accounts_unavailable`, so one such wallet no longer fails the whole projection refresh and takes every other wallet's `/wallets` tree down with it. Spends and sender selection against it fail closed naming that reason; every other Broker error still fails the refresh, so nothing is served on a broken edge.

### Account creation

- Writing `{"request_id":"<id>"}` to `wallets/<w>/new` starts one account-allocation ceremony that always requests the EVM and Solana keys under one Signer-chosen number; extra fields are rejected (`families` is gone). Machine derives the custody operation id from wallet + request id, so a retry returns the same ceremony or, after success, the same account. Reading `new` reports each request's pending/created/failed/expired/cancelled state, ceremony URL and — only after the authenticated receipt — the shared number and both keys. Failed, expired, and cancelled outcomes are durable: reusing the ID returns that result, and a new ID starts a new ceremony.

### Account-scoped Petal dispatch

- `wallets/<w>/<n>/petals/` mounts the installed Petals for that account. Accounts other than 0 run only Petals that declare `[account] aware = true`; an unaware Petal is not found and the message names the missing declaration. Account 0 and the flat `/petals/` mount list everything.
- The runner injects host-owned `bloom.wallet`, `bloom.account`, `bloom.owner_key_fingerprint` next to `bloom.route_id`; any caller-supplied context entry starting with `bloom.` is rejected before injection. The owner fingerprint is chosen **per route** (the family its key-derive suites name when they name exactly one, else the account's only family) — never per Petal and never by list order.
- The daemon resolves the owner from the numbered path `(wallet, number, suite)` against fresh Broker membership before any approval, custody ceremony, or signature. A Petal-supplied key reference must equal that owner or the call is denied; a body wallet that differs from the mounted account is denied; mixed-family key requests are invalid; EVM staging selects the chain projection for the requested chain.

### Pins (all landed on their default branches)

- `bloom-broker-api` and release Broker → `bd85885bb4fc2c05f20af1f4410683ffbebca6a9` (broker master). It descends from `81b30c07`, the landed #58 squash, which carries #60's list-only account-allocation terms (#60 merged into #58's branch before #58 merged).
- `bloom-signer-derive` and release Signer → `941dd376568b52ccd269ecf495bcff7dcd9af504` (signer master). It descends from `104cf218`, the landed Signer #40, which is required for Exact recovery after Petal scope expiry.
- Both pins came in with #238 when `b1d68d4f` merged `feat/solana-native` forward. Each is a fast-forward of this branch's earlier pin. All build, lockfile, compatibility, workflow, and test references agree.
- `bloom-petal-contract` → `9ad7a5c6` (landed petal #33 squash).
- Stacked on `feat/solana-native` (#170, review-only). Retarget to master when #170 lands.

## Verification

- #238 pin merge (`b1d68d4f`, Broker `bd85885b` / Signer `941dd376`): `cargo check --workspace --all-targets --locked` passed. `cargo test --locked --lib` passed for bloom-machine-client, bloom-daemon, bloom-vfs, bloom and bloom-petals: 715 tests, 0 failed. `cargo fmt --all --check` passed. The items below were recorded at the earlier pins.

- Wallet-level outbox fence (`503b468c`, merged forward with `feat/solana-native` in `a3660808`): `cargo test -p bloom-vfs --lib --locked` (390 passed), `cargo clippy -p bloom-vfs -p bloom-it --all-targets --locked -- -D warnings`, `cargo fmt --all --check`, and `cargo test -p bloom-it --test solana_multi_account -- --ignored` against a fresh agave v3.0.0 ledger (account 1 stages, confirms, and settles through `wallets/alice/1/`; the wallet-level stage naming account 1 is refused). The controls test routes confirms through the Machine→Broker signing path and asserts no Broker request for another account's entry, while account 0's own confirm does reach it; with the fence removed the test fails.

- Review fixes: `cargo test -p bloom-daemon --lib --locked` (99 passed), `cargo test -p bloom-vfs --lib --locked` (373 passed), and `cargo clippy -p bloom-daemon -p bloom-vfs --all-targets --locked -- -D warnings` passed. New coverage exercises account-bound retry state, numbered EVM fingerprint rejection, pending-session IPC install/uninstall guards, fixed expiry, and terminal account-creation retries.

- `cargo test -p bloom-it --test solana_multi_account -- --ignored` against a pinned agave v3.0.0 validator: wallet-level staging defaults to account 0; the numbered account-1 path stages, confirms, and settles on the validator (only account 1 is funded, so a wrong key cannot settle); account 0's pending does not list account 1's entry and cross-account confirm is not found; a body fingerprint that disagrees with the path is refused.
- `cargo fmt --all --check` and `git diff --check` passed for the review fixes. Full workspace Clippy and other current-head checks are tracked by CI.
- Broker-side mechanisms (list-only terms, two-child receipt adoption, by-key revocation, Exact-after-scope) are proven in broker #58/#60 tests; Signer-side mechanisms in signer #39/#42 tests. The production Broker–Signer `w5_ceremony::policy_service_requires_completion_then_commits_and_replays_over_authenticated_rpc` test passed at Broker `81b30c0725200c97db0d3ab67299c617f31f30ad` / Signer `104cf2183b3b33224ef79caab091bcc80875dd9d`, including Exact recovery after scope expiry. Mocked-Broker tests here do not claim to prove Signer rejection or chain settlement.

### Sessions, stop, and the install guard

- `wallets/<w>/<n>/sessions/<petal>/<slot>/session.json` renders every key a Petal derived through that account from the local key-state files: delegating owner, delegated key, scope, recorded approvals, truthful `signing_authority` (`pending`, `active`, `stopped`, `expired`, `package_replaced`), freshness, and — after a stop or expiry — the routes still reachable through a payload-specific Exact approval. Listing, stat, and reads never call Broker.
- Writing the sibling `stop` journals `sealed_approval.revoke_for_key` under a deterministic operation id (idempotent retries), succeeds only once every reported approval is terminal, and works after the Petal is uninstalled; a partial revocation persists with `complete: false`, leaves the authority unchanged, and names the stragglers.
- Replacing **or removing** an installed package that still has active or unresolved pending sessions is refused naming their mounted paths; `bloom petal install --force` and `bloom petal uninstall --force` override it and the sessions read `package_replaced`. Reinstalling the scoped hash restores them. Both guards resolve the outgoing package hash under the mutation lock that mutates it, so neither can race a concurrent install, and a forced removal stays recoverable because the session keeps rendering and its `stop` still revokes.
- Petal key states record their mount, request time, last observed success, and the expiry of the approval terms accepted by Broker. Polling never extends expiry; older records with unknown expiry report null and remain guarded until stopped or forcibly removed (schema v3, v2 accepted on read; the v2→v3 file migration only moves records the wallet owns).
- A Petal-named signing key may be the mounted account's owner or one of its delegated session keys — never a key of another account or wallet.

Durable Exact and reusable batch requests include the full selected KeyRef in their identity; persisted retries reject a changed key before contacting Broker.

### Wallet-level outbox fence

A live two-account validation showed the wallet-level outbox was not account 0's view: it listed and read account 1's pending entries, and a wallet-level `confirm` on account 1's entry prepared the approval that later settled it (the returned `permission denied` was `ApprovalRequired`). Now:

- The wallet-level outbox lists, looks up, reads, and acts only on entries account 0's key staged, for EVM and Solana, including `latest` and every pending control; another account's entry is not found. Account 0's identity comes from the cached projection (no live Broker call on reads).
- Wallet-level staging always spends from account 0. The wallet-level Solana `account_fingerprint` override is removed: a body fingerprint naming another account is refused; stage from account `n` at `wallets/<w>/<n>/`. Wallet-level Solana staging no longer falls back to a lone active child (account 1) when account 0's key is retired.
- One scoped outbox implementation per family serves both surfaces, and the write surfaces take a required sender, so an unfenced outbox write is not expressible. Numbered paths gain `latest` and the `restage` sink on expired `failed` Solana entries.
- `nonce_conflicts.json` compares only account 0's outbox nonces against account 0's mempool view.
- A root-key wallet keeps its root as account 0 even when the Broker reports its inventory unavailable (legacy BIP-32 custody). Only a rootless wallet with an unavailable inventory falls back to unfiltered wallet-level reads, with writes refused naming the reason.

### Native-path gaps closed

- A root-key wallet renders in its root key's own family (an Ed25519 root is Solana, not EVM).
- Wallet-level Solana reads resolve through the cached authenticated inventory and re-observe live only when it is stale-marked; the error says so.
- The numbered EVM outbox rejects a body fingerprint naming another account, symmetric with Solana.
- `bloom wallet accounts` prints each entry's `number` (null for off-mapping paths).

The SDK surface (`account_ctx`) landed in petal #32; the Hyperliquid (#20) and Pump.fun (#2) Petals adopt it in their own repositories. Review fixes are published in `67e8205f8408b53a41d3151545c558c444ce0830`; the wallet-level outbox fence is `b55621fe` + `503b468c`, and `a3660808` merges the base forward (#232's not-found mapping now applies in the scoped outbox). `b1d68d4f` merges the base forward again to take #238's Broker and Signer pins. The 12 PR checks passed at `67e8205f`.

The Linux triad release build, reproducible bundle verification, and installer install/uninstall smoke passed at Machine `67e8205f8408b53a41d3151545c558c444ce0830` / Broker `81b30c0725200c97db0d3ab67299c617f31f30ad` / Signer `104cf2183b3b33224ef79caab091bcc80875dd9d`. Local triad installer tests passed 45 cases; five macOS simulations failed on this Linux host, and the identical five failures reproduce on the original `37951326` head (missing plutil/enrollment parsing). These results predate the #238 pin bump. The Linux release build and macOS conformance have not been re-run on this branch at Broker `bd85885b` / Signer `941dd376`.

[Explicit-ref macOS Unix conformance](https://github.com/bloom-directory/bloom/actions/runs/34553540061) is still in progress. Two-login conformance has not run: the required self-hosted runner and real GUI identities are unavailable. Full macOS release conformance remains outstanding. Merge still requires independent approval and landing/retargeting from parent #170.

## Checklist

      behavior changed (`crates/bloom-vfs/src/docs/agent-guidance.md` is
      updated in-tree; the Wallet.md / Petal succession / Sealed Approvals
      paragraphs for the sessions, stop, and install-guard slices land with
      those slices in this PR)


violet-agent:
### Blocking: active sessions can still be stranded by uninstall

The new active-session guard runs only in `IpcOperationContext::commit_petal_package` (the install/replacement path). `IpcServer::do_petals_uninstall` directly calls `runner.uninstall` under the mutation lock, without resolving the outgoing package or consulting `active_session_slots`.

That means `bloom petal uninstall <name-or-hash>` can strand an active delegated session without the explicit `--force` override the new package-replacement contract promises. Please apply the same guard to uninstall (resolve its full outgoing package hash while holding the mutation lock), or introduce an explicit, audited force option with the same `package_replaced` behavior. Please add IPC-level coverage; the existing direct-runner uninstall exercise does not prove this boundary.

0xdewy:
Confirmed and fixed in `37951326`.

The finding is accurate: the guard ran only in `commit_petal_package`, and `do_petals_uninstall` called `runner.uninstall` under the mutation lock without resolving the outgoing package or consulting `active_session_slots`. The branch already recorded the consequence without acting on it — `the_install_guard_names_live_sessions_and_replacement_restores_them` comments "Uninstalling the package strands a live session" and asserts the slot list goes empty afterwards.

`petals.uninstall` now resolves the target to its full package hash inside the `spawn_blocking` body that holds the mutation lock, so the guard names the exact outgoing package and cannot race a concurrent install, and refuses when that hash has active session slots. The refusal names them and points at the same `stop` path the replacement refusal does.

On the choice of an override rather than a hard refusal: `stop_session` requires the Broker edge, so a categorical refusal would make a package unremovable for as long as the Broker is unreachable. `bloom petal uninstall --force` therefore mirrors the install contract — the sessions read `package_replaced`, and their `stop` still revokes, so a forced removal stays recoverable rather than stranding authority. The request shape is now strict like the install side, so a misspelled `force` is rejected instead of silently ignored.

IPC-level coverage as requested: `ipc::tests::petals_uninstall_refuses_to_strand_active_sessions_until_forced` drives `do_petals_uninstall` itself. It asserts refusal both by petal name and by full hash, that nothing is removed on either refusal, that a package with no active session is unaffected, that `force` proceeds, that a second forced removal reports `removed: false`, and that an unknown field is rejected. The older test is annotated to say it calls the runner beneath this guard and to name the new coverage.

Docs updated: `agent-guidance.md` and the PR body now say replacing *or removing* is guarded, and that a forced removal leaves `stop` working.

CI is green on `37951326` (12 checks, 2 skipped). Locally: bloom-daemon 97, bloom-petals 178, bloom-vfs 372, bloom CLI 64 pass, `cargo clippy --workspace --all-targets -- -D warnings` clean.

GavinPacini:
Reviewed head `379513262f7d8ca6c9d6c3d16ca24cdfa5504d1c`. These are source-review findings; I inspected CI results but did not rerun the suites locally.

1. **[P1] Repin Signer to the recovery-compatible revision.** `packaging/triad/release/compatibility-v1.toml:5` still pins `18e4d041`, which predates Signer #40 (`104cf2183b3b33224ef79caab091bcc80875dd9d`). Broker accepts the Exact-after-expiry recovery contract, but this packaged Signer rejects it, potentially preventing a funded expired session from sweeping. Advancing Signer is required for the advertised behavior, rather than the optional follow-up described in the PR body. Update every manifest, lockfile, CI/conformance ref, release pin, and test literal together. The repin in Bloom #250 (`aa1e56ee`) provides the matching tuple, but #248 needs it independently if it lands first.

2. **[P2] Include the selected account/key in durable Petal approval identity.** In `crates/bloom-daemon/src/lib.rs:960-1024`, `petal_reusable_signing_paths` hashes package, route, wallet, operation class, and signature count, but no account or selected key. Two accounts invoking the same route therefore share an approval state file. Exact signing has the same collision when payload and claim match. The persisted-state checks in `crates/bloom-vfs/src/exact_signing.rs:605-643` also omit the selected key, while the outgoing request uses the newly selected key. This can reuse another account's approval/operation IDs and cause authorization conflicts; this is not a claim of a Signer bypass. Include the full selected key in durable identity and retry validation, and test alternating accounts on one route.

3. **[P2] Fix the unreachable numbered-EVM fingerprint mismatch check.** In `crates/bloom-vfs/src/handlers/wallets/accounts.rs:550-565`, a normal `/wallets/w/1/chains/base/outbox/new.tx` request has `rest = [chains, base, outbox, new.tx]`, but the guard matches the two-element `[outbox, new.tx]` slice. It therefore never executes. The downstream JSON intent parser ignores the extra fingerprint field, so a body naming account 0 is staged through account 1 instead of being rejected. Match `chain_rest == [new.tx]` and add a negative test through the mounted EVM path.

4. **[P2] Make the package guard account for approved sessions that have not been polled.** In `crates/bloom-daemon/src/lib.rs:3311-3340`, `session_authority` returns `pending` until the local record becomes `succeeded`, and the replacement/uninstall guard includes only `active` sessions. The transition to `succeeded` happens when the Petal retries after approval. Approve a session and replace its package before that retry: the guard can allow removal even though Broker already holds an active approval. The package mutation lock does not close this observation gap. Treat unresolved sessions conservatively or reconcile their authority before allowing removal; cover approval followed by uninstall/replacement without an intervening Petal poll.

5. **[P2] Derive session expiry from authenticated authority, not polling time.** In `crates/bloom-daemon/src/lib.rs:1418-1427`, every successful retry overwrites `succeeded_at_ms` with the current time. `session_authority` and `session_entry` calculate expiry as that timestamp plus the scope lifetime. Repeated polling moves the apparent expiry forward, although custody authority has not changed; delayed polling also starts the displayed clock late. Broker/Signer still enforce expiry, but `session.json` and the install guard become misleading. Persist authenticated expiry separately from the last-observed timestamp, and test delayed/repeated polls around expiry.

6. **[P2] Render terminal account-creation failures as terminal states.** In `crates/bloom-vfs/src/handlers/wallets/accounts.rs:1197-1205`, every ceremony state except `Succeeded` renders the unchanged pending record, including terminal failures. Retrying the same request ID only refreshes that record; it neither recovers nor explains that a new request is necessary. Add terminal states and an explicit retry contract, with failed/expired ceremony tests. This is particularly relevant when #250 removes the alternative creation route.

0xdewy:
Addressed all six findings in `67e8205f`: updated the Signer pins, bound signing retries to the selected key, fixed the EVM fingerprint guard, guarded unpolled sessions, corrected expiry tracking, and added terminal account-creation states.

Regression tests pass, all 12 PR checks passed, and Linux release packaging passed. Verification details and remaining macOS conformance limitations are recorded in the PR description. Ready for re-review.

GavinPacini:
**[P2] Preserve the numbered account in Solana approval challenge paths**

Reviewed head `a36608081329b8a4eec27b935725d6a054e90e78`. The new wallet-level account-0 fence introduces a broken approval-resume flow for accounts above zero:

1. Stage and confirm a transfer through `wallets/alice/1/chains/<solana-chain>/outbox/` that requires owner approval.
2. Read the generated `approval_challenge.json` and complete the ceremony.
3. Follow its `retry_path`: it points to `wallets/alice/chains/<solana-chain>/outbox/pending/<id>/confirm`, which now returns not found because that path is fenced to account 0. The generated `plan_path` is inaccessible for the same reason.

The challenge still constructs wallet-level paths in [`crates/bloom-vfs/src/handlers/wallets.rs:4084–4085`](https://github.com/bloom-directory/bloom/blob/a36608081329b8a4eec27b935725d6a054e90e78/crates/bloom-vfs/src/handlers/wallets.rs#L4084-L4085), while the wallet-level write now pins account 0 at lines 3091–3101. Agent guidance explicitly tells callers to use `retry_path` verbatim. Retrying through the original numbered account path works, so this is a broken recovery pointer, not an account-isolation bypass.

Please carry the mounted account/path into challenge generation and emit reachable `plan_path` and `retry_path` values. Add coverage that stages through account 1, obtains a challenge, reads its advertised plan, and resumes through its advertised retry path after approval. The current multi-account integration test retries its original numbered `confirm` variable, so it does not exercise the challenge's pointer.

This is a source-review finding; I did not rerun CI.

0xdewy:
Fixed in `e6d9cb71`. `write_solana_outbox` now takes the mounted account number, and the challenge builds `plan_path`/`retry_path` from it: `wallets/<w>/<n>/chains/<c>/outbox/…` for a numbered confirm, and the wallet-level path for account 0's wallet-level outbox. Wallet-level keeps its own paths rather than switching to `…/0/…`, because a root-key wallet whose account inventory is unavailable has no numbered tree. `agent-guidance.md` now says which outbox the pointers name.

Coverage: `solana_multi_account` step 5e now stages through account 1, reads the refused confirm's `approval_challenge.json`, asserts both pointers name account 1's outbox, reads the advertised `plan_path`, and resumes through `retry_path` verbatim after approval. The transfer settles on the validator. It passed against a fresh agave 3.0.0 ledger. bloom-vfs lib tests (390) and clippy are clean.

PR #235, opened 2026-09-07
Grade an agent's native SOL transfer through harbor

0xdewy:
Teach Harbor how to grade an agent doing a native SOL transfer through Bloom: the agent writes a transfer intent into Bloom's mounted wallet directory, an owner approves it in a signing ceremony, the Signer signs, and the transaction settles on a disposable local validator. A verifier then checks the chain shows exactly one correct payment.

## Why

There is currently no working evaluation of Bloom's Solana path with an agent in the loop. This adds one, so regressions in staging, approval, signing, or settlement get caught by a graded task instead of by users.

## What changed

- **The task.** An agent instruction, a grading script (`verify_result.py`), and a setup guide that takes an operator from a fresh local validator to a funded wallet with an approved destination allow-list.
- **The grader.** A harness that stages the transfer, approves the owner's signing ceremony only when every field matches exactly (destination, amount, fee payer, fee ceiling, signing account — a missing identity field is refused, never passed), and confirms the chain identity by asking the endpoint for its genesis hash rather than trusting a label. Configured amounts above 0.02 SOL are rejected; wallet balances are never checked.
- **Restage handling.** If a staged transfer expires, the follow-up entry is approved only when the failed entry's own advice file names it as the replacement, so a duplicate staging can't farm a second approval.
- **Runners.** Two scripts: a generic Harbor entry point and a local-lane wrapper that refuses to start unless the ceremony listener is up on the canonical port 18734, plus a smoke mode that drives the expiry → restage → single-payment path without an LLM.
- **Tests.** 267 harness unit tests and verifier tests covering the clean pass plus eight tampered chain views (wrong destination, source, amount, fee, failed tx, missing slot, inner instructions, extra payment).
- **Mount-free regression gate.** `BLOOM_EVAL_SOLANA_TRANSPORT=vfs` runs `--smoke-only` through `bloom vfs cat/ls/write` over the Machine IPC socket — the same VFS handlers the kernel mount serves — so the deterministic smoke needs no NFS mount, no sudo rule, and no interference with a host's canonical ceremony port. Agent trials keep the mounted transport. The documented 24-turn default is now implemented (`EvalDefinition.default_max_turns`).

## Not in this PR

- No product code changes: `evals/` and `scripts/` only. Nothing to migrate, flag, or roll back.
[...]

PR #42, opened 2026-06-11
Feat/polymarket

0xdewy:
# Polymarket v2 integration + agent-signing hardening

This PR adds Bloom's first Polymarket integration. A Bloom wallet can now
onboard to Polymarket, fund its deposit wallet, place and exit trades, and clean
up approvals.

It also tightens the signing boundary used by agent workflows. The important
point is not "agents can trade now." They cannot, unless a policy path explicitly
allows the value-moving action.

## The Mental Model

Bloom separates three things:

1. **Planning**
   - Agents can read markets, inspect balances, build plans, and stage drafts.
   - This does not move funds.

2. **Policy**
   - Wallet policy decides what is allowed, warned, or denied.
   - Deny means stop. A human review does not bypass deny-level policy.

3. **Authorization**
   - If the action is value-moving, Bloom needs an approval path.
   - Today that is either:
     - a fresh reviewed signature; or
     - for EVM outbox broadcasts only, `agent_autonomy = "under_policy"` after
       every required check passes.

Polymarket orders still use CLI confirmation. They are not approved by the EVM
`under_policy` path.

### Polymarket

Bloom can:

- onboard a wallet to Polymarket's deposit-wallet flow;
- fund the deposit wallet;
- create durable order drafts;
- confirm and post orders;
- sell, cancel, inspect obligations, redeem resolved positions, withdraw pUSD,
  and revoke approvals.

The integration uses Polymarket's deposit-wallet mode:

- `signatureType = 3` / POLY-1271;
- maker/funder is the deterministic deposit wallet;
- Bloom signs with the owner key;
- builder/relayer credentials are submission credentials, not builder fee codes.

Trading is disabled by default. A wallet must opt in through `policy.toml`.

The Polymarket policy checks:

- per-order USD cap;
- trailing 24-hour daily USD cap from receipts;
- max price;
- allowed/denied slugs and condition IDs;
- binary-market requirement;
- negative-risk restrictions.

Confirming an order re-reads live market/book state and current policy under the
wallet lock. Draft values are not trusted at confirm time.

### Funding

Funding supports:

- direct pUSD transfer when possible;
- Enso swap funding when needed.

Funding transactions go through the EVM outbox:

1. Build the route/funding plan.
2. Evaluate policy.
3. Stage a reviewable transaction.
4. Approve via fresh review or `under_policy`.
5. Broadcast only if chain broadcast gates allow it.

Same-chain Polymarket funding has a receiver consistency check: Bloom only marks
[...]

PR #250, opened 2026-09-09
Remove the standalone account allocation command

0xdewy:
## Summary

BIP-39 registration and import now return canonical EVM and Solana accounts together, so the standalone Solana allocation ceremony is obsolete. This removes `bloom wallet account-allocate`, its Machine IPC variant, the follow-up error hint, and the corresponding documentation and import-script step.

The release combination now pins bloom-broker `91836edb3191b37a358a298859d92b0e3dd7557c` (broker #60's head, unlanded — repin at squash time) and bloom-signer `18e4d04181f29e752480adb3934326328539568c` (signer master, the squash that carries #42) in manifests, lockfiles, compatibility metadata, workflow defaults, and test literals. The BIP-39 transfer script verifies both account-zero paths and a Base58 Solana address immediately after import.

Validation:

- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p bloom -p bloom-daemon -p bloom-machine-client -p bloom-vfs --quiet`
- `bash -n scripts/test-bip39-import-transfer.sh`

A follow-up commit `3838af770ee526a391dd15cd526a341d233b027b` rewords the `wallet address` hint for wallets without derived accounts; `cargo fmt --check`, `cargo clippy -p bloom --all-targets -- -D warnings`, and `cargo test -p bloom` were re-run at that head.

## Checklist

- [x] Architecture docs updated — N/A; this removes a CLI/API compatibility path and does not change an architecture contract
- [x] Sealed Approval invariants respected — the live transfer still fails closed until its approval ceremony completes
- [x] Agent Documentation updated — N/A in this base; the VFS `wallets/<w>/new` guidance is updated in #248

Live dev Triad verification passed with Machine `75af9615028abf32fb47970fcd5f94a940ca0e4d`, Broker `91836edb3191b37a358a298859d92b0e3dd7557c`, and Signer `c278cccc93cf87df71173da8755ce77b5f5e9fca`, all built with `triad-dev-harness`. The import acceptance completed import, two-family projection, policy update, sealed approval, EVM signing, Anvil broadcast, and reconciliation. A separate `bloom wallet new bip39-new-e2e` ceremony in the same run returned exactly these two active paths with no allocation ceremony:

```text
m/44'/60'/0'/0/0
m/44'/501'/0'/0'
```

`bloom wallet address bip39-new-e2e --profile solana` returned `2JncJaZYiCb6UHq3F8UMkSMxYRHrPGXGeTBAdQPUF6Hp`. The full account projection and run log are retained at `shared/verification/solana-at-registration-20260909/bip39-import-transfer-final.log`; exact revisions and harness limits are beside it in `revisions.txt`.

macOS Unix-principal conformance was dispatched twice for this branch with the exact Broker and Signer refs, in runs [34396728108](https://github.com/bloom-directory/bloom/actions/runs/34396728108) and [34397863628](https://github.com/bloom-directory/bloom/actions/runs/34397863628). Both failed at the `Run destructive W0 isolation checks` step with `W0 payload has the wrong platform claim`. Cause: bug 1 of #244, merged to master as `dea8fbe01139316e05bda9ff98f3540fb9377cd5`, re-bundles the candidate under the `macos-unix-principals-w0` claim; neither `feat/solana-native` nor this branch contains that commit, so the workflow cannot pass on this base regardless of this PR's change. It was not re-dispatched; the re-run belongs after #170 merges master forward and brings the #244 fix with it. The two-login workflow additionally requires the active GUI users and UIDs of its disposable self-hosted runner; no such runner inputs are recorded in the repository or prior workflow history, so it has not been dispatched with invented identities.

The same documentation and CLI context also appears on open bloom PRs #235 (`evals/solana-harbor-recovery`), #200 (`docs/triad-development-workflow`), and #234 (`agent/root/pumpfun`); their owners will need to resolve those lines when merging `feat/solana-native` forward. #248 (`agent/claude/account-tree`) sits on the same base and still carries `bloom wallet account-allocate`, its IPC variant, and the README line (adapted to the list-only allocation API), so whichever of this PR and #248 lands second merges forward through those conflicts as well.

Base: bloom #170 (`feat/solana-native`) at `4ce4dc88ae3624b75e3c27b407d3cab274bc09a3`, merged into this branch.

### DIFF
--- a/crates/bloom-daemon/src/ipc.rs
+++ b/crates/bloom-daemon/src/ipc.rs
@@ -649,10 +649,6 @@ pub enum MachineCommand {
     WalletAccounts {
         name: String,
     },
-    WalletAccountAllocate {
-        name: String,
-        profile: String,
-    },
     WalletAccountRetire {
         name: String,
         fingerprint: String,
--- a/crates/bloom-daemon/src/ipc.rs
+++ b/crates/bloom-daemon/src/ipc.rs
@@ -19,7 +19,7 @@
 //! | `petals.install` | remote source or package transport | package metadata          |
 //! | `petals.build` | `{ "package_dir", "out"? }`           | package metadata       |
 //! | `petals.list` | `null`                              | `[ package, ... ]`        |
-//! | `petals.uninstall` | `{ "hash" }`                   | `{ "removed" }`          |
+//! | `petals.uninstall` | `{ "hash", "force"? }`         | `{ "removed" }`          |
 //! | `machine.execute` | tagged [`MachineCommand`]        | [`MachineCommandOutput`] |
 //! | `shutdown` | `null`                                | `null`                    |
 //!
@@ -470,11 +470,20 @@ pub struct IpcOutputEvent {
     pub bytes: Vec<u8>,
 }
 
+/// Sessions an install must not strand, keyed by the installed package
+/// hash they scope to.
+pub type ActiveSessionSlots = Arc<dyn Fn(&str) -> Vec<String> + Send + Sync>;
+
 #[derive(Clone)]
 pub struct IpcOperationContext {
     output: Option<tokio::sync::mpsc::Sender<IpcOutputEvent>>,
     cancelled: Arc<AtomicBool>,
     petal_mutation: Option<Arc<tokio::sync::Mutex<()>>>,
+    /// Sessions keyed by installed package hash that an install must not
+    /// strand, supplied by the daemon from its Petal key-state files.
+    petal_active_sessions: Option<ActiveSessionSlots>,
+    /// Set by an install that explicitly overrides the session guard.
+    petal_install_force: bool,
 }
 
 impl IpcOperationContext {
@@ -483,6 +492,8 @@ impl IpcOperationContext {
             output: Some(output),
             cancelled: Arc::new(AtomicBool::new(false)),
             petal_mutation: None,
+            petal_active_sessions: None,
+            petal_install_force: false,
         }
     }
 
@@ -491,6 +502,8 @@ impl IpcOperationContext {
             output: None,
             cancelled: Arc::new(AtomicBool::new(false)),
             petal_mutation: None,
+            petal_active_sessions: None,
+            petal_install_force: false,
         }
     }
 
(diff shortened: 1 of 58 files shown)


packaging/triad/release/compatibility-v1.toml
```diff
@@ -1,8 +1,8 @@
 schema = "bloom.triad-compatibility/1"
 
 [revisions]
-broker_commit = "cd1e31e9f33c74355bf3e7fa244af12159716064"
-signer_commit = "9abe7917b8380d8b6d31af0b376a79c26e816764"
+broker_commit = "91836edb3191b37a358a298859d92b0e3dd7557c"
+signer_commit = "18e4d04181f29e752480adb3934326328539568c"
```
josh-richardson: [P1] Pin the Signer revision that implements the same recovery contract as this Broker. Broker 91836ed includes #58’s rule allowing a fresh Exact approval after a Petal key scope expires, but Signer 18e4d041 contains #39/#42 and predates Signer #40 (104cf218). With this recorded release tuple, Broker admits the owner-recovery path and Signer rejects it at ceremony/sign authorization, leaving funds on an expired delegated key unsweepable. Before merge, advance every Signer ref/lockfile/test literal to 104cf218 or its landed successor, together with the final #58/#63 Broker revision.

crates/bloom-daemon/src/ipc.rs
```diff
@@ -649,10 +649,6 @@ pub enum MachineCommand {
     WalletAccounts {
         name: String,
     },
-    WalletAccountAllocate {
```
josh-richardson, line 652: [P2] Gate removal of this creation route on the replacement in #248. This head keeps WalletAccountRetire, but after an account-zero child is retired there is no CLI or daemon IPC path left that can call account_allocate; automatic creation during registration only solves initial provisioning. #248 adds wallets/<w>/new, but it is a sibling rather than a dependency of this PR, so merging/releasing #250 first leaves a BIP-39 wallet unable to create a replacement active account. Either stack/land #248 first (and state that gate) or retain this route until the replacement is present. This is lifecycle completeness for the new unreleased feature, not a backward-compatibility claim about v0.2.1.

josh-richardson (changes requested):
I found one P1 cross-service pin mismatch and one P2 landing-order gap, both inline, so I am requesting changes. This is bloom-directory/bloom#250—the only Bloom-directory PR numbered 250; there is no bloom-signer#250.

The removal and list-only API edits are otherwise coherent, and the removed command itself was not part of the published v0.2.1 surface. All current GitHub CI jobs are green. Local formatting passed; the relevant Bloom CLI tests passed before the broader package run stalled in unrelated long-running source-install tests, so I stopped that redundant run.

packaging/triad/release/compatibility-v1.toml
GavinPacini: Resolved in aa1e56eecb290f0c71e49e759d54198cf8dbf8b9. Signer is now pinned to 104cf2183b3b33224ef79caab091bcc80875dd9d, and Broker to 81b30c0725200c97db0d3ab67299c617f31f30ad (merged #63, following #58). Updated all Cargo manifest and lockfile pins, CI/conformance refs, release compatibility metadata, and test literals together. No old pins remain. `cargo check --workspace --locked` passed; packaging tests had 75 passes and three Linux installer failures caused by macOS mv lacking -T.

Issue #242, opened 2026-09-08
Any Petal claim that debits value or declares a fee is unconditionally rejected — `value_limits` is never populated and `fee_asset` is always `None`


jer117:
## Summary

Two gaps in the sealed-approval claim pipeline combine to make it impossible for a Petal to actually move value or charge a fee today, on any wallet:

1. `ApprovalLimits.value_limits` is never populated with anything by Machine — every construction site hardcodes it empty.
2. Every operation class's `fee_asset` in the provenance catalog is hardcoded `None`, regardless of whether the operation actually charges a fee.

Either one alone would block real usage; together, no Petal action with a non-empty `declared_debits` or a non-`none` `declared_fee` can ever be approved.

## Observed

While funding a Hyperliquid testnet eval wallet via `usd_send` (a plain internal transfer, no fee), the write was denied post-signature:

```
petal exact signing failed code=ClaimInvalid declared debit or fee asset is absent from approval limits
reason=CLAIM_INVALID: declared debit or fee asset is absent from approval limits
```

This is not a policy misconfiguration — no wallet policy anywhere can produce a non-empty `value_limits`, because nothing ever writes to it.

Separately, [`bloom-petal-hyperliquid`'s `builder_order_fee`](https://github.com/bloom-directory/bloom-petal-hyperliquid/blob/main/route/src/workflow.rs#L62-L73) already declares a real `DeclaredFee::Fee` for orders that carry a builder fee — this is the mechanism by which a builder collects revenue on a trade. Any such order is rejected before it even reaches the `value_limits` check, because the `Order` operation class's `fee_asset` is `None`.

### 1. `value_limits` is never populated

`account_claim_values` (bloom-broker, [`crates/bloom-broker/src/authority.rs:2906-2979`](https://github.com/bloom-directory/bloom-broker/blob/main/crates/bloom-broker/src/authority.rs#L2906-L2979)) builds the set of assets a claim touches from `claim.declared_debits` plus, if applicable, `claim.declared_fee`, then checks every one is present in `terms.limits.value_limits`:

```rust
let allowed: BTreeSet<_> = terms
    .limits
    .value_limits
    .iter()
    .map(|limit| asset_id(limit.asset.chain.as_str(), &limit.asset.asset))
    .collect();
if values.keys().any(|asset| !allowed.contains(asset)) {
    return Err(denied(
        "VALUE_ASSET_NOT_ALLOWED",
        "declared debit or fee asset is absent from approval limits",
    ));
}
[...]

PR #251, opened 2026-09-09
fix: bind approval value limits, unblock Hyperliquid actions, and harden the builder-fee eval (#242)

jer117:
## Summary

When a wallet owner approves an agent's signing request with their passkey, Bloom's Broker checks each signature against the spending limits recorded in that approval. Those limits were never being filled in, so any request that declared money moving — a debit or a fee — was refused, even after the owner had approved it.

This PR makes approvals carry the amounts they actually declare, fixes a related enrollment bug that was blocking most everyday Hyperliquid actions, and adds and hardens the Harbor eval that tests the Hyperliquid builder-fee flow end to end.

## Short version

- **Approvals now carry the value they declare.** Requests that declared a debit or fee were refused even after the owner approved them, because the approval's spending limits were empty. All three signing paths now fill them in.
- **Everyday Hyperliquid actions are no longer blocked.** Cancels, leverage changes and plain orders were refused because their shared category was wrongly enrolled as fee-bearing.
- **The builder-fee eval now proves what it claims, and cleans up safely.** The agent does the real work, the eval checks the right wallet, and nothing is left live afterwards. Passkey counters are never reused — not by two evals sharing one passkey, and not by an eval and the operator's own runs — and each run reserves every counter its cleanup will need before it starts. It also refuses a wallet that already approves the builder, rather than erasing that approval during cleanup.
- **The Harbor test suite roughly doubled,** from 72 tests to 141.

### 1. Approvals now carry the value they declare

**The problem.** All three signing paths in `bloom-machine-client` built approvals with an empty `value_limits` list. The Broker rejects any request whose asset is missing from that list (`VALUE_ASSET_NOT_ALLOWED`), so a request declaring a debit or fee could be approved by the owner and still never be signed.

**The fix.** Each path now works out its limits from the request it is signing. A fee paid in the same asset as a debit is added to it, and an amount too large to represent is rejected up front instead of being silently mishandled.

### 2. Everyday Hyperliquid actions are no longer blocked

**The problem.** Developer enrollment marked Hyperliquid's shared `hyperliquid.agent_action` category as fee-bearing. The Broker reads that as "every request in this category must declare a fee" — but cancels, leverage changes and plain orders all share that category and declare no fee, so most of them were refused (`FEE_REQUIRED`).

**The fix.** Shared categories are now enrolled without a fee. The reasoning is written up in `docs/architecture/Sealed Approvals.md`, because adding a fee to a shared category breaks every other action that uses it.

### 3. The builder-fee Harbor eval now proves what it claims

This adds the `hyperliquid-approve-builder-fee` eval and closes several ways it could report a pass without really proving anything, or leave the wallet worse off than it found it:

- **The agent does the real work.** The harness prepares the owner's approval but leaves the actual submission to the agent. Previously the agent's write could be a silent no-op, with the result reflecting work the harness had already done.
- **It checks the right wallet.** It confirms the wallet it writes to actually owns the address it reads results from, so it can't approve on one wallet and grade another.
- **Nothing is left live afterwards.** Cleanup uses up an approval the agent never submitted, instead of leaving it active behind a later revoke.
- **It never erases an existing approval.** Cleanup resets the builder fee to zero rather than to whatever it was before, so a run on a wallet that already approved the builder — even for a smaller fee — would have wiped that approval out. The eval now refuses any wallet with an existing approval for the builder, before it changes anything.
- **Passkey counters can't be reused.** Every passkey ceremony uses up a counter, and the Broker rejects a reused one as a replay. Counters are saved before each ceremony and survive a restart, and a run refuses to start if that record can't be written. The record belongs to the passkey rather than to an eval or a checkout, and lives in one place per user (`~/.bloom/eval-counters/`). So everything that signs with that passkey — evals run directly, and the operator's policy updates and order-cancel runs — draws from a single locked record and never signs with the same counter, even across separate clones or worktrees and even when they run at the same moment.
- **Cleanup always has a counter left.** Passkey counters are 32-bit, so they can run out. Checking that enough remain isn't sufficient on its own: two evals on one passkey could both pass the check, and one could then use up the counter the other's cleanup needed after it had already granted. So just before a run can change anything, it reserves every counter it may need — cleanup included — in one locked step. Those counters are then its alone, and a run that can't reserve them all is refused before it starts.

### Also

- Installing from source no longer picks a pre-release (such as `v2.0.0-rc.1`) over a stable release.
- Hand-rolled code was replaced with established libraries: `alloy` for EIP-712 hashing and `governor` for Etherscan rate limiting (now with tests).
- The dev triad launcher no longer loses Broker logs on macOS.
- The clock helper's docs now say honestly which callers still fall back to `0`.

## How this was verified

- **Automated tests:** the Harbor harness suite grew from 72 tests on `master` to 141 — 69 new tests, the last 6 of them covering the operator and direct eval runs sharing one passkey counter. All 141 pass locally, as do 400 Rust tests, and the Harbor CI script passes end to end. (Three installer tests fail only on macOS, because of a Linux-only `mv` flag, and pass on Linux CI.) Every new test was checked to fail when the fix it covers is removed, including the two cleanup races above, tested in both orders.
- **Live, on the final commit:** against a running triad with a freshly re-imported wallet, preauthorization, wallet binding and the full policy check all passed on real data. A passkey-approved `approveBuilderFee` then reached Hyperliquid mainnet and came back with the exchange's own answer — the test builder address holds too little balance to be approved, an exchange rule that confirms the request really reached the venue.

This also addresses the review feedback on earlier revisions of this branch.

## Found during testing (separate follow-ups, not fixed here)

Live testing turned up two problems outside this PR's scope. Both are tracked as their own follow-up tasks:

- **Passkey approvals fail at random in Chrome.** The Signer refuses any passkey response that contains a field it doesn't recognise. Chrome sometimes adds an extra field to that response on purpose, so strict parsers like this one break. When it does, the ceremony fails with "WebAuthn clientDataJSON is malformed", even though the passkey is fine. Safari doesn't add the field, which is why the live test above was completed in Safari. The fix is for the Signer to ignore unknown fields while still checking the ones it needs.
- **Hyperliquid Petal builds aren't reproducible.** `petal build` picks the newest compatible version of every dependency each time it runs, instead of using a locked set. So rebuilding the same source can produce a package with a different hash. The wallet's policy approves one exact package hash, so a rebuild can silently stop signing working until the owner approves a new policy. The fix is to build from a committed lockfile.

## Checklist

      behavior changed


violet-agent (changes requested):
Blocking finding:

`HyperliquidApproveBuilderFeeEval._grant_or_revoke()` advances only its local `counter`. It never persists the returned value back to `self.sign_count`, so cleanup starts again from the original `BLOOM_EVAL_AUTHENTICATOR_SIGN_COUNT` after a successful grant. The revoke ceremony will reuse the WebAuthn sign counter already consumed by the grant, which should be rejected as a replay and leaves the granted builder-fee approval live. Update the instance's next counter after each `_drive_action()` call (including on the error path as appropriate) and add a grant-then-cleanup test proving strictly increasing driver counters.

violet-agent (commented):
Additional blocking findings from the completed parallel review:

1. **Harbor task can pass without the agent doing the task** (`evals/harbor/harness/hyperliquid_approve_builder_fee.py:602`, `.../verify_result.py:132`): `provision()` submits and confirms the exact approval before Harbor starts. The verifier subsequently accepts the still-present venue state plus agent-authored fields, but has no trusted proof of a post-container write. An agent can emit a synthetic report and pass without touching `/bloom`. Bind grading to a trusted receipt/request record produced by the agent's mounted write, not the host-created precondition.

2. **Ambiguous grant failures skip cleanup** (`hyperliquid_approve_builder_fee.py:602-603,641`): `granted` becomes true only after `_grant_or_revoke()` returns. If the venue accepted the side effect but a later response/poll read fails, cleanup returns early and leaves a live approval. Set a conservative cleanup-needed flag before the side-effecting write and reconcile/revoke on every ambiguous outcome; test that path.

3. **Panic on aggregating valid individual U256 claim amounts** (`crates/bloom-machine-client/src/lib.rs:2140-2144`): two same-asset debit/fee amounts can each parse as `DecimalU256` while their sum exceeds U256. The `expect` then panics before normal protocol rejection. Return a protocol error on overflow and add a max-value duplicate regression test.

4. **Implicit source installs can select prereleases** (`crates/bloom/src/github_source.rs:1618`): the no-`--ref` selector now considers `v2.0.0-rc.1` newer than a stable `v1.9.9`, unlike the prior parser. Filter `version.pre.is_empty()` for automatic/latest selection; retain prerelease support when explicitly requested; add the stable-vs-RC regression test.

violet-agent (commented):
**High — `crates/bloom/src/triad_enrollment.rs:478-485`: `hyperliquid.agent_action` is too broad to be marked fee-bearing.**

The Broker's `account_claim_values()` treats a class with `fee_asset: Some(...)` as requiring `DeclaredFee::Fee` and rejects `DeclaredFee::None` with `FEE_REQUIRED`. But the shipped Hyperliquid Petal signs *all* delegated session actions (including cancel, update-leverage, and order) as `hyperliquid.agent_action` with `ClaimEffects::none()` / `{"kind":"none"}` (`bloom-petal-hyperliquid/route/src/workflow.rs:1359-1367`; its protocol currently explicitly rejects builder fees because they cannot be represented exactly). Enrolling this catalog therefore makes every such existing agent action fail Broker authorization, not merely fee-bearing orders.

Do not attach a fee asset to this shared class. Introduce/use a distinct fee-bearing operation class only when the Petal emits an exact declared fee for that class, and add a cross-repo contract test covering a `DeclaredFee::None` agent action plus the intended fee-bearing path.

violet-agent (commented):
**High — `evals/harbor/harness/hyperliquid_approve_builder_fee.py:786-799`: cleanup does not invalidate an unconsumed staged grant.**

`provision()` intentionally leaves the grant request in `approved_retry_required`, which is a durable owner-approved exact write. If the agent never submits it (or exits before the write), `cleanup()` signs and submits a *different* revoke request using `nonce + 1`; it never cancels or consumes the old grant request. That old nonce/body remains an executable approval. A later writer that can replay the original request can raise the builder fee again after cleanup has observed zero, so cleanup's claimed postcondition is not durable.

Please explicitly cancel/invalidate the staged grant when it was not consumed, or bind revocation to invalidate prior outstanding approvals for this builder. Add a test that stages a grant, skips the agent write, runs cleanup, then asserts replaying the original staged request is rejected and cannot restore the venue fee.

violet-agent (changes requested):
## High — builder-fee evaluator does not durably advance the WebAuthn counter

`evals/harbor/harness/__main__.py:43-45` constructs `HyperliquidApproveBuilderFeeEval` without a `counter_committed` callback. In `_drive_action()` (`hyperliquid_approve_builder_fee.py:609-617`), the advanced counter is only persisted if that callback exists; otherwise it is retained only on the current Python object.

After a process interruption or a later fresh run, construction reads the unchanged `BLOOM_EVAL_AUTHENTICATOR_SIGN_COUNT` and may reuse a counter already presented to Broker by the prior grant or cleanup ceremony. Broker treats that as a replay. The durable `StateStore.update_counter` wiring exists for `HyperliquidOrderCancelEval` in `operator.py:1015-1019`, but this evaluator is not connected to an equivalent durable state path.

The new counter regression test proves only in-process monotonicity; it does not persist the next counter and instantiate a fresh evaluator.

**Required fix:** run the builder-fee evaluator through a durable state store (or require an atomic counter-file callback), pass the store's counter-update callback at construction, and add a restart regression that verifies the next evaluator begins with the persisted counter.

violet-agent (changes requested):
## High — the builder-fee evaluator still has no durable counter sidecar

The new shared `EvalDefinition.reserve_counter()` only persists when `counter_committed` is configured (`evals/harbor/harness/core.py:122-134`). But the normal builder-fee entry point still constructs `HyperliquidApproveBuilderFeeEval(repo_root)` with no callback (`evals/harbor/harness/__main__.py:43-45`). Its `preflight()` therefore calls a no-op `require_counter_durability()`, and each ceremony advances only the in-memory `sign_count` (`hyperliquid_approve_builder_fee.py:610-616`).

The operator wiring at `operator.py:1027-1032` attaches `StateStore.update_counter` only to `HyperliquidOrderCancelEval`; it does not run the builder-fee evaluator. So a direct builder-fee run that consumes a grant/revoke counter and then exits leaves `BLOOM_EVAL_AUTHENTICATOR_SIGN_COUNT` unchanged. A new process can reuse that counter, which Broker rejects as replay.

The new generic helper and tests cover the callback-present path, but do not make the builder-fee production path provide one.

**Required:** either add a durable counter-file/state-store implementation for `HyperliquidApproveBuilderFeeEval`'s normal `__main__` path (with atomic pre-driver reservation and restart coverage), or route builder-fee through an operator state lifecycle that supplies `StateStore.update_counter`. The current change does neither.

violet-agent (commented):
Follow-up to my changes-requested review — two additional confirmed blockers at `c5273733`:

## High — reusable Petal approvals still omit `value_limits`

`crates/bloom-machine-client/src/lib.rs:1127-1133` builds the only `ApprovalSelector::Petal` path with `value_limits: Vec::new()`. This selector deliberately permits refreshed payloads after ceremony, so it is the path where the declared debit/fee cap matters. Broker denies claim assets absent from the approval limits; a reusable value-carrying Petal approval can therefore be ceremonially approved but cannot execute. Derive limits from the required claim here, as the exact-payload paths already do, and add a prepared-terms regression test.

## High — builder-fee preflight does not bind wallet ID to owner address or validate policy projection

`evals/harbor/harness/hyperliquid_approve_builder_fee.py:472-484` reads only `wallets/<wallet_id>/policy.json`. It neither verifies that this wallet ID owns `BLOOM_EVAL_WALLET` nor requires `broker_verified`, `fresh`, and matching policy digest metadata from `addresses.json`. But it signs via `exchange/<wallet_id>` while its venue observation is under `users/<BLOOM_EVAL_WALLET>`.

A mismatched configuration can thus authorize one wallet while grading another. Mirror the existing owner, freshness, broker-verification, and digest checks in `hyperliquid_order_cancel.py:627-654`, with mismatched-ID/address regression coverage.

violet-agent (changes requested):
## High — direct-run WebAuthn sidecar is partitioned by eval, not credential

`harness/__main__.py:52-60` defaults the sidecar path to `evals/harbor/{definition.name}.counter.json`. Both Hyperliquid evals consume the same configured authenticator seed and `BLOOM_EVAL_AUTHENTICATOR_SIGN_COUNT`, but direct runs of `hyperliquid-order-cancel` and `hyperliquid-approve-builder-fee` therefore read/write separate sidecars and take separate locks. Each can reserve the same next count. A later ceremony then replays a counter already accepted by Broker; in builder-fee that can prevent the required revoke after a grant was staged.

Key and serialize the direct-run counter store by stable authenticator identity (for example a canonical credential/seed identifier), not eval name. Add a cross-eval regression proving one direct eval's reservation becomes the starting floor for the other.

## High — builder-fee accepts the final u32 counter although its mandatory cleanup needs another

`hyperliquid_approve_builder_fee.py:143-156` accepts `BLOOM_EVAL_AUTHENTICATOR_SIGN_COUNT=4294967295`. Grant staging can consume it and updates `self.sign_count` to `4294967296` (`:621-628`). `cleanup()` always performs a fresh-nonce revoke (`:847-870`), which then sends that out-of-range value to the driver; sidecar restart parsing also rejects it as out of range.

So a valid boundary configuration can stage/grant authority but cannot complete its required revoke, leaving cleanup incomplete. Fail preflight before staging unless enough counter capacity exists for both the grant and mandatory cleanup ceremony (or otherwise reserve a bounded counter budget), with a max-boundary regression test.

violet-agent (changes requested):
## High — cleanup capacity is checked but not reserved atomically across the shared credential

The new `CEREMONY_BUDGET` preflight in `harness/core.py:284-301` reads the shared counter but does not reserve the whole budget. Individual counter values are committed later, one at a time. Each eval's lifecycle lock is still definition-specific, so a builder-fee run and an order-cancel run sharing an authenticator can both pass preflight.

At a shared counter of `4294967291`, builder-fee passes its two-counter budget and order-cancel passes its four-counter budget. Builder consumes `4294967291`; order-cancel then consumes `4294967292..4294967295`; builder's mandatory revoke has no usable counter. Its grant can therefore be live while the cleanup ceremony cannot execute.

Reserve the full cleanup-inclusive budget atomically under the credential-sidecar lock, or serialize all evaluator lifecycles by the same credential lock for the duration of a run. Add the competing-evals boundary regression.

## High — cleanup can destroy a pre-existing lower builder-fee approval

`HyperliquidApproveBuilderFeeEval.provision()` only rejects an existing fee at or above the eval target (`hyperliquid_approve_builder_fee.py:775-784`). A lower nonzero baseline is accepted, overwritten by the eval grant, and cleanup unconditionally submits a zero-fee revoke (`:875-878`). That leaves the venue at zero rather than its prior lower nonzero value.

Refuse **any** nonzero baseline before staging (simplest / least authority), or durably capture and restore the exact old value with the corresponding additional ceremony and counter budget. Add coverage for a lower nonzero baseline.

violet-agent (approved):
Final review of 4c99f3e2: approved. The prior blockers are addressed: full per-run counter blocks are atomically reserved under the credential sidecar lock before authority creation, and builder-fee now refuses every nonzero baseline so cleanup cannot erase existing venue authorization. Focused local verification and all hosted required checks pass.

violet-agent (changes requested):
Correction to my approval: a final independent pass found a confirmed mixed-topology counter collision, so I am changing this back to **changes requested**.

## High — operator order-cancel and direct builder-fee still do not share counter state

The direct runner attaches `CounterSidecar` (`harness/__main__.py:75-80`), but the operator constructs order-cancel with only `StateStore.update_counter` (`operator.py:1027-1032`). It does not attach the credential-keyed sidecar. Consequently `reserve_run_counters()` returns early for the operator path (`core.py:354-360`); it reserves no shared budget and observes only its independent state file.

A concurrent direct builder-fee run and operator-driven order-cancel run using the same authenticator can therefore allocate overlapping counter values/ranges. Their locks differ too. This permits Broker counter replay and can again leave a builder-fee grant without a usable cleanup counter.

Unify the counter allocation and lock across *both* direct and operator paths—e.g. make the operator use the same credential-keyed sidecar/block reservation as direct runs, with operator state updated from that allocation—or prohibit/technically prevent mixed topology for one authenticator. Add an operator-vs-direct concurrency regression.

violet-agent (approved):
Re-review approved at 3c43ff02. The operator/direct topology now shares one credential-keyed durable counter allocator for policy ceremonies and evaluator runs, atomically reserves full run budgets, and mirrors the restart floor into operator state. Local Harbor regression coverage passed; hosted CI is still completing for this head.

0xdewy:
Cross-linking the shared-defect landing order from the read-only comparison posted on #265 ([comment](https://github.com/bloom-directory/bloom/pull/265#issuecomment-5644088784)), so it's tracked here too. Posting for coordination — this PR's owner decides how to proceed.

#265 fixes the `value_limits` half of #242. It was opened from live normal-install verification of bloom-petal-hyperliquid#22, where every debit-declaring Petal operation failed post-ceremony with `CLAIM_INVALID: declared debit or fee asset is absent from approval limits`. It's a one-file fix (derive per-asset lifetime limits from the claim's `declared_debits` + declared fee at the three `bloom-machine-client` prepare sites), live-verified end to end against Hyperliquid mainnet, with automated end-to-end acceptance tests added at head `39213ff8` (CI green; awaiting required non-author review; one open review request there to aggregate through the full U256 range instead of `u128`).

**Recommended landing order for the shared defect:**
1. #265 lands first (squash), with the U256-range review request folded in or otherwise addressed.
2. This PR rebases onto it and reconciles.

**What reconciliation means here, per the comparison:**
- Drop the duplicate `value_limits_from_claim` + `add_decimal_strings` in `crates/bloom-machine-client/src/lib.rs` (same three call sites, guaranteed textual conflicts, near-identical semantics). The one arithmetic divergence — this PR bounds sums at U256 via string addition while #265 currently rejects beyond `u128` — should be settled by adopting the landed helper once #265 takes the U256-range change.
- Keep the `triad_enrollment.rs` change: enrolling shared operation classes (e.g. `hyperliquid.agent_action`) without a fee asset is the disjoint `fee_asset` half of #242 (`FEE_REQUIRED` denials for cancels, leverage changes, and plain orders), which #265 deliberately does not touch.
- Eval hardening and the unrelated extras (installer pre-release pinning, alloy/governor adoption, dev launcher) ride along after the rebase.

Full split tracking is on #242.

violet-agent (approved):
Re-review approved after the master merges at 4027b839. The merge deltas do not alter Harbor; the current implementation retains atomic shared credential counter-block reservation, fail-closed builder-fee baseline handling, and cleanup retirement. Local Harbor regression suite and all hosted checks for this exact head pass.

crates/bloom/src/github_source.rs
```diff
@@ -1716,8 +1705,34 @@ mod tests {
 
     #[test]
     fn selects_latest_semver_like_tag() {
-        let latest = latest_semver_tag(["v0.1.0", "v0.10.0", "junk", "0.2.1"].into_iter()).unwrap();
-        assert_eq!(latest.tag, "v0.10.0");
+        let (tag, _version) =
+            latest_stable_semver_tag(["v0.1.0", "v0.10.0", "junk", "0.2.1"].into_iter()).unwrap();
+        assert_eq!(tag, "v0.10.0");
```
jer117, line 1748: We could add something like renovate to help keep things up to date.

jer117:
@0xdewy Should I split the changes out a bit? I know some refactoring was done to remove hand written that is outside of the scope of the actual issue. 

"What are the largest opportunities to refactor this codebase where portions of functionality are handrolled that could be imported from large, well used crates? We want to handroll less of our own crypto / functionality that other people have done before in a well audited manner?"

I gave claude this prompt but I think the removal of some the hand-written EIP-712 digest with alloy's eip712_signing_hash is a good thing to add in.

PR #256, opened 2026-09-10
release: enforce matching binary versions and add IPC versions and changelog

GavinPacini:
Releases currently permit different Machine, Broker, and Signer binary versions and publish only static installation notes. This change requires all three compatibility-matrix versions to equal the release version before either platform builds; the existing candidate checks verify the actual binaries against those declarations.

Release bodies now include the common binary version, both Machine–Broker and Broker–Signer IPC ranges, and every Bloom commit since the preceding published stable release on the target commit's ancestry. The generator excludes drafts, prereleases, unpublished tags, and later publications on retries, and includes full history for the first release. Both release creation and retry updates use these notes.

The release guide and generated version-bump PR remind developers to bump all three repositories and update the reviewed authority pins. The current matrix (Machine 0.2.1, Broker/Signer 0.1.0) intentionally fails the new gate until that coordinated bump is done.

Validation: seven Python unit tests pass; git diff --check passes; running the gate against the current matrix rejects Broker 0.1.0 as expected. Metadata tests also run in release preparation. Full Linux/macOS dry-run candidate builds were not run because the current version mismatch blocks preparation; dispatch a dry run after completing the coordinated bump and before merging the release changes.


GavinPacini:
Superseded by #257: harmonize triad releases while allowing independently versioned Machine, Broker, and Signer components. The issue covers release-tag selection, Broker and Signer release processes, canonical version metadata with local commit overrides, and complete triad version/IPC/changelog reporting. Closing this PR because requiring equal binary versions is not the intended policy.

Issue #241, opened 2026-09-08
Landing order for the Solana custody stack


0xdewy:
Remaining work to land the Solana custody stack. Updated 2026-09-11 21:14 UTC.

## Order

**Done:** these merged into `feat/solana-native`: #238 with #261 (20:32), #248 (20:49), #266 (20:51) and #250 (21:02). The branch head is `456d0a09`.

1. **#170 → master.** Every child that is ready has merged in. All checks pass on `456d0a09`. Get a review, then squash merge. Before merging, retarget every child still open (#260, #234, #235, #200) to master or close it. Then delete `feat/solana-native` only with the web "Delete branch" button.
2. **Decisions, not blocking #170:**
   - **#260:** Gavin moved it back to draft because it "contains a bunch of pretty important product decisions", which need deciding before it merges. After step 1 it targets master.
   - **System selector vs durable nonces:** Gavin now points Signer #28, Broker #55 and Bloom #234 at #259 (durable nonce accounts) and suggests closing #234 in favour of it. He also notes that durable nonces may cover only native SOL transfers, not other protocols or account types. Signer #28 and Broker #55 stay on hold until that is settled.

#235 and #200 are optional and have no review. Both now merge cleanly into the branch head: #200 was merged forward at 22:24 UTC, resolving the #248 overlap. Merge either in if it is reviewed before step 1, otherwise retarget it to master.

The rest of this issue explains why, and gives each PR's state.

**Landed**
- Signer: #36 (`8304e556`, including #43's scope-takeover fix), #27 (`9a83d425`, ceremony window is now the Signer config key `ceremony_ttl_ms`), and #44. Signer master is `941dd376`.
- Broker: #35 (`90bc3ea5`, carrying #54's dual-loopback ceremony listeners and #64's review fixes), then #49 and #65. Broker master is `57cbc6c6`. #65 only rewords destination refusals, so the stack stays pinned to `bd85885b`, which is on Broker master.
- Bloom, into `feat/solana-native`:
  - #238 with #261 (`ef15077f`): both loopback ceremony listeners, and the repin to Broker `bd85885b` and Signer `941dd376`.
  - #248 (`63ea7cdf`): numbered HD accounts.
  - #266 (`9db6f9ad`): read-only Solana mainnet in newly generated configurations.
[...]

PR #261, opened 2026-09-11
Pin Broker master into #238 and fix Linux upgrade recovery

GavinPacini:
## Summary

Follow-up to #238, targeting its branch. It covers the review findings on #238: the Broker repin #238 needs before it can merge, the recovery P2, and lower-priority installer and conformance items. It also includes #263, which fixed three regressions the review of this PR found.

### 1. Pin Broker master (`bd85885b`) and the Signer it is built against

- **No standalone #54 commit exists.** Broker #54 was merged into #35's branch and reached Broker master inside #35's squash commit (`90bc3ea5`). `bd85885b` is that commit plus a dependency bump. Every Broker commit that contains #54 also contains #58.
- **Broker pinned in all 7 places:**
  - `Cargo.toml` and `Cargo.lock`
  - `packaging/triad/release/compatibility-v1.toml`
  - the `ci.yml` macOS W0 call
  - the defaults in `macos-unix-conformance.yml` and `macos-two-login-conformance.yml`
  - the matrix assertion in `triad_release.rs`
- **Signer advanced to `941dd376`** (Signer master) in the same places and in `crates/bloom-it/Cargo.toml`. That is the revision Broker master is built against.
  - Relative to #238's old pin (`9abe7917`) the bump is required: that Signer predates the `derivation_request` → `derivation_requests` rename, so the release pair would disagree on account-allocation requests.
  - Relative to #248/#250's pin (`104cf218`) the bump is only additive.
- **Machine migrated for Broker #58:**
  - `CustodyPrepareRequest.derivation_request` becomes `derivation_requests: Vec<_>`.
  - `AccountTerms.derivation` becomes `derivations`.
  - The change is mechanical because Machine already sends no pinned account number, which the new contract requires.
  - The Machine–Broker API (`bloom-broker-api`) is identical between #248/#250's `81b30c07` and `bd85885b`. #248 and #250 make the same rename, so whichever lands second resolves trivial textual conflicts.

### 2. P2: recovery no longer strands a host whose previous release cannot start

**Before:** recovery rolled back through `rollback_linux_upgrade`, which clears the transaction only after the *previous* release passes its health gate. A host upgrading away from a release that can't start never got past that gate, so every later run restored the old release, failed its health check, kept the transaction and exited.

**Now:**
- The stop, restore-units, select-binary and restore-metadata steps live in `restore_linux_previous_release`.
- The in-process rollback still restarts and health-checks the restored release, and keeps the transaction if that fails.
- Recovery restores, **clears the transaction**, and only then restarts the previous release without gating on it (#263). A broken previous release no longer blocks the retry, and a run that exits after recovery, or reinstalls the previous payload, doesn't leave logins stopped.
- Recovery no longer requires every login to have an active session, since it doesn't start anything under a health gate (#263).
- `rewrite_linux_release_set`, `write_linux_record`, `replace_linux_json_digest` and `atomic_install` now check their own failures. Rollback runs with errexit off, and with no health gate after recovery nothing else would catch a half-written metadata rewrite.
- Uninstall refuses while an upgrade transaction is pending (#263).

### 3. Lower-priority fixes

- **Rollback no longer reports success while a login is down.** `start_linux_release_set` and `stop_linux_release_set` used to return only the *last* login's status, because rollback runs with errexit off. They now check each step, keep going for later logins, and fail if any login failed.
- **IPv6 check before installing.** A live install stops with status 69 unless `lo` has `disable_ipv6=0` and `::1` assigned. The check runs after recovery, which restarts the restored release first, and before anything new is written.
- **Conformance for the IPv6 family.** The macOS W0 foreign-listener conflict now runs once per family (`127.0.0.1` and `[::1]`). A foreign listener on either address must stop the Broker from serving at all, with no TCP listener of its own and no durable-state change. W0 also requires the ceremony-owner marker on both addresses after activation, and no answer on either after logout.

### Reverted: Signer KMS drop-in snapshot

An earlier commit here added each login's AWS KMS Signer drop-in and credential to the upgrade snapshot. `16017e95` reverts it, returning to #238's fixed unit set and three-field manifest.
- **Low value:** it only helped when a KMS host was upgraded with a payload lacking the KMS overlay and the upgrade then failed. Rerunning with the overlay repairs that, as it does after a successful upgrade with the same payload.
- **Caused a regression:** it made restoration depend on which logins were enrolled at restore time, which is what the uninstall regression came from.
- **Copied secrets:** it copied every login's AWS credential into the transaction directory, although an install only rewrites the installing login's.

The README now says the KMS files aren't snapshotted and how to repair them.

## Verification

- **Linux, run locally:** `cargo test -p bloom-it --test triad_release --test linux_packaging --locked -- --skip macos_` in a Debian container as a non-root user gives 47 + 7 passed, 0 failed. This covers every Linux installer test, including the rollback, recovery, uninstall-guard, start/stop and IPv6 tests.
- `cargo clippy --workspace --all-targets --locked -- -D warnings` and `cargo fmt --all -- --check`: clean. `--test macos_launchagent` (21) passes on macOS.
- **CI:** the "local integration tests" job runs the Linux installer tests on every push.
- **macOS W0 ([run 34639530054](https://github.com/bloom-directory/bloom/actions/runs/34639530054)):** dispatched on `w0/pr261`, which is this branch plus a cherry-pick of #244 because #238's branch lacks it. It failed at install, before the ceremony checks, with the same activation failure every W0 run has hit since Sept 8. The captured Machine log shows the cause: the LaunchAgent's `--mount-home` NFS mount of `127.0.0.1:12049` times out after 10s on GitHub's hosted runners, Machine treats that as fatal, and launchd restarts it 14 times. This is unrelated to this PR.
- **Not run:** an end-to-end upgrade to this Broker/Signer pair on a live host.

## Landing notes

- `schema bloom.linux-upgrade-transaction.2` is unchanged from #238's head.
- #241 should be updated: the #238 repin target is now `bd85885b` (Broker #54 is inside `90bc3ea5`), the Signer moves to `941dd376`, and #248/#250 overlap on the `derivation_requests` rename.

## Checklist

      behavior changed — N/A: the Machine–Broker field rename follows the Broker API and no architecture doc names these fields. Operator behavior is documented in `packaging/triad/release/README.md` and `packaging/triad/linux/README.md`.
      bytes) — see `docs/architecture/Sealed Approvals.md` — N/A, no approval path touched
      `docs/architecture/Agent-native Documentation.md` — N/A, no agent-visible surface

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WU9eDmAqur5eXvU9SEaWZB

### DIFF
--- a/packaging/triad/release/install-linux.sh
+++ b/packaging/triad/release/install-linux.sh
@@ -188,14 +188,6 @@ numeric_file_mode() {
   fi
 }
 
-numeric_file_owner() {
-  if [[ "$(uname -s)" == Darwin ]]; then
-    stat -f '%u:%g' "$1"
-  else
-    stat -c '%u:%g' "$1"
-  fi
-}
-
 # The Broker requires both loopback ceremony listeners, so a host whose
 # kernel cannot bind [::1] cannot run this release. Check before anything is
 # mutated rather than let the health gate discover it after the unit writes.
@@ -638,17 +630,12 @@ switch_linux_release() {
   mv -fT "$replacement" "$release_base/current"
 }
 
-# Every Bloom-owned file the installer overwrites or removes while installing
-# a release. An upgrade transaction snapshots exactly these paths so a
-# rollback restores the previous binary selection together with the unit
-# contract and Signer instance configuration it ran with. The unit templates
-# are shared by every login. Each enrolled login also has an AWS KMS Signer
-# drop-in and the credential it loads: installing a payload without the KMS
-# overlay deletes both, and restoring one without the other would leave the
-# previous Signer loading a missing credential. Administrator units and
-# drop-ins outside this list are never touched by restoration.
+# Every Bloom-owned unit template the installer overwrites or removes while
+# installing a release. An upgrade transaction snapshots exactly these paths
+# so a rollback restores the previous binary selection together with the unit
+# contract it ran with. Administrator units and drop-ins outside this list are
+# never touched by restoration.
 linux_upgrade_unit_paths() {
-  local install_root="$1" record uid
   printf '%s\n' \
     usr/lib/systemd/system/bloom-broker-ceremony@.socket \
     usr/lib/systemd/system/bloom-broker-ceremony-ipv6@.socket \
@@ -661,41 +648,32 @@ linux_upgrade_unit_paths() {
     usr/lib/systemd/system/bloom-broker-control@.socket \
     usr/lib/systemd/user/bloom-session.service \
     usr/lib/systemd/user/bloom-machine.service
-  for record in "$install_root/etc/bloom/enrollments"/*.json; do
-    [[ -f "$record" && ! -L "$record" ]] || continue
-    uid="$(linux_record_number "$record" login_uid)"
-    [[ "$uid" =~ ^[1-9][0-9]*$ ]] || continue
-    printf '%s\n' \
-      "usr/lib/systemd/system/bloom-signer@$uid.service.d/50-aws-kms.conf" \
-      "etc/bloom/$uid/signer/aws-credentials"
-  done
 }
 
 linux_upgrade_unit_is_tracked() {
-  local install_root="$1" candidate="$2" tracked
+  local candidate="$1" tracked
   while IFS= read -r tracked; do
     [[ "$tracked" == "$candidate" ]] && return 0
-  done < <(linux_upgrade_unit_paths "$install_root")
+  done < <(linux_upgrade_unit_paths)
   return 1
 }
 
-# Records the actually installed contract — explicit absence included — into
-# the upgrade transaction scratch directory, one manifest line per tracked
-# path: "present MODE UID:GID PATH" or "absent - - PATH". Fails without
-# mutating the installation when a tracked file is substituted or unreadable.
+# Records the actually installed unit contract — explicit absence included —
+# into the upgrade transaction scratch directory. Fails without mutating the
+# installation when a tracked unit is missing, substituted, or unreadable.
 # Every write is checked explicitly: when this function runs inside a
 # rollback step, errexit is suspended for the whole call tree, so a skipped
 # operation would otherwise be reported as a successful snapshot.
 snapshot_linux_upgrade_units() {
   local install_root="$1" scratch="$2"
-  local relative path mode owner
+  local relative path mode
 
   mkdir -m 0700 "$scratch/units" "$scratch/units/files" || return 65
   : > "$scratch/units/manifest" || return 65
   while IFS= read -r relative; do
     path="$install_root/$relative"
     if [[ ! -e "$path" && ! -L "$path" ]]; then
-      printf 'absent - - %s\n' "$relative" >> "$scratch/units/manifest" || return 65
+      printf 'absent - %s\n' "$relative" >> "$scratch/units/manifest" || return 65
       continue
     fi
     [[ -f "$path" && ! -L "$path" ]] || {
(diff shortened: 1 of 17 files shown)


packaging/triad/release/install-linux.sh
```diff
@@ -615,32 +661,41 @@ linux_upgrade_unit_paths() {
     usr/lib/systemd/system/bloom-broker-control@.socket \
     usr/lib/systemd/user/bloom-session.service \
     usr/lib/systemd/user/bloom-machine.service
+  for record in "$install_root/etc/bloom/enrollments"/*.json; do
```
0xdewy: **Uninstalling any login while a transaction is pending strands the host.** The snapshot writes a drop-in and credential line (present or absent) for every login enrolled at snapshot time, but `restore_linux_upgrade_units` re-derives the tracked set from whatever enrollments exist at restore time.

Repro, run against this head: install 1000 and 2000, interrupt an upgrade (harness `interrupt all`), then `install-linux.sh uninstall --retain-custody ROOT 2000`. Uninstall doesn't check for a pending transaction, so it succeeds. Every recovery after that, and every `install`, fails with `Linux upgrade unit snapshot names an untracked path` (exit 65).

`begin_linux_upgrade` refuses while the transaction exists, and deleting it by hand doesn't help once the records are `activating`. The natural trigger is the preflight 69 ("requires an active session for enrolled UID 2000"), which invites the operator to remove that login.

This didn't exist in #238, where the tracked set was static. Also note the installer only ever rewrites the *installing* login's drop-in and credential (`config_root`/`dropin_root` use `$login_uid`), so snapshotting every login's is broader than what an upgrade mutates.

Minimal fix: uninstall refuses while `var/lib/bloom/upgrade-transaction` exists, and recovery drops `preflight_linux_release_set`, since it no longer starts anything under a health gate. That way the guard is never a dead end.

0xdewy (commented):
Three regressions in the new recovery path, all in `install-linux.sh`. The repin itself checks out: Broker `bd85885b` pins Signer `941dd376` in all 7 Signer crates, and #238's current Broker pin `cd1e31e9` predates Broker #54, so #238 does need it. I reproduced finding 1 against this head with the existing upgrade harness, and I'll open a fix PR against this branch.

packaging/triad/release/install-linux.sh
```diff
@@ -992,9 +1107,12 @@ recover_interrupted_linux_upgrade() {
...
+  # upgrade_rollback_required stays false: a failed restoration preserves the
+  # transaction for the next run rather than attempting a restart from the
+  # EXIT trap.
   echo "restoring the interrupted Bloom Linux release before retrying" >&2
-  if ! rollback_linux_upgrade; then
+  if ! restore_linux_previous_release; then
+    echo "Bloom Linux upgrade recovery failed; transaction preserved for retry" >&2
     return 65
   fi
   finish_linux_upgrade
```
0xdewy, line 1070: **Recovery now leaves every login stopped, but only the `release_upgrade=true` path restarts the set.** Any run that recovers and then doesn't reach `start_linux_release_set` leaves every enrolled login down until a reboot:

- **Reinstalling the previous payload:** after an interrupted A→B, rerunning with A gives `shared_release_digest == release_digest`, so `release_upgrade=false` and only `bloom-session@$login_uid.path` is enabled. Other logins' path units were stopped by recovery and nothing re-arms them.
- **An early exit after recovery:** the IPv6 check (69), a non-enrolled login requesting a different release (65 at `:1402`), or any validation failure before `begin_linux_upgrade`.

Before this PR, recovery restarted the restored release. The README's "until the next installer run" doesn't hold for the first case either, because rerunning A again changes nothing.

Suggested fix: keep the P2 fix and restart after clearing the transaction, without gating on it:
```bash
  finish_linux_upgrade
  start_linux_release_set "$install_root" ||
    echo "the restored Bloom Linux release did not start cleanly; continuing" >&2
```
A host whose previous release can't start still gets through, because the transaction is already gone. The cost is one start/stop cycle before a real retry.

packaging/triad/release/install-linux.sh
```diff
@@ -1269,6 +1387,9 @@ case "$action" in
       exit 65
     }
     recover_interrupted_linux_upgrade "$root"
+    if linux_root_is_live "$root"; then
+      require_linux_ipv6_loopback /proc
```
0xdewy, line 1345: **This check runs after recovery has stopped every login and cleared the transaction.** On a host without IPv6 loopback it exits 69 with the restored release down. That release is typically pre-#238 and runs fine on IPv4 only. Every later run fails here before starting anything.

The restart suggested on `:1118` covers this, and keeps the check where it is: recovery still has to run first, or a host that was interrupted and rebooted stays on mixed units.

0xdewy:
Fixes for the three review findings are in #263, which targets this branch.

PR #269, opened 2026-09-12
feat(mcp): proxy the VFS command surface

violet-agent:
## Summary

Adds an opt-in, stdio MCP server that proxies Bloom's canonical daemon VFS command surface.

- Adds `[mcp] enabled = true` as an explicit, disabled-by-default server gate; missing or legacy config remains disabled.
- Adds `bloom mcp status` and `bloom mcp serve`. `serve` refuses before touching stdio or the daemon socket unless the flag is enabled.
- Maps every daemon VFS method one-to-one: `list`, `read`, `lookup`/stat, `write`, and `write_with_lookup`.
- Preserves canonical daemon error codes, VFS path discovery, byte reads/writes (base64 for non-UTF-8), and stdio JSON-RPC/MCP protocol behavior.

## Verification

- `cargo fmt`
- `cargo test -p bloom-mcp` — 32 passed
- `cargo test -p bloom-proto config::tests::mcp` — 2 passed
- `cargo test -p bloom --test cli mcp` — 2 passed
- `git diff --check`

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A: this is an opt-in local CLI transport/proxy; VFS and custody contracts are unchanged.
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes) — N/A: MCP forwards only existing VFS daemon operations and introduces no signing or approval path.
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — N/A: no agent guidance or Petal surface changed; MCP tool descriptions document the new transport.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2133,6 +2133,7 @@ dependencies = [
  "bloom-daemon",
  "bloom-evm",
  "bloom-machine-client",
+ "bloom-mcp",
  "bloom-mount",
  "bloom-petals",
  "bloom-proto",
@@ -2357,6 +2358,23 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber 0.3.23",
+]
+
+[[package]]
+name = "bloom-mcp"
+version = "0.2.1"
+dependencies = [
+ "async-trait",
+ "base64",
+ "bloom-daemon",
+ "bloom-proto",
+ "bloom-vfs",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "thiserror 2.0.18",
+ "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2369,6 +2369,7 @@ dependencies = [
  "bloom-daemon",
  "bloom-proto",
  "bloom-vfs",
+ "percent-encoding",
  "serde",
  "serde_json",
  "tempfile",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ reqwest = { version = "0.12", default-features = false, features = ["json", "rus
 flate2 = "1"
 mpp = { version = "0.10.4", default-features = false, features = ["client", "tempo", "reqwest-rustls-tls"] }
 url = "2"
+percent-encoding = "2"
 humantime = "2"
 humantime-serde = "1"
 tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "crates/bloom-machine",
   "crates/bloom-daemon",
   "crates/bloom-vfs",
+  "crates/bloom-mcp",
   "crates/bloom-paid-mpp",
   "crates/bloom-paid-x402",
   "crates/bloom-paid-http",
@@ -109,6 +110,7 @@ bloom-rpc      = { path = "crates/bloom-rpc" }
 bloom-evm      = { path = "crates/bloom-evm" }
 bloom-tx       = { path = "crates/bloom-tx" }
 bloom-vfs      = { path = "crates/bloom-vfs" }
+bloom-mcp      = { path = "crates/bloom-mcp" }
 bloom-paid-mpp = { path = "crates/bloom-paid-mpp" }
 bloom-paid-x402 = { path = "crates/bloom-paid-x402" }
 bloom-paid-http = { path = "crates/bloom-paid-http" }
--- a/README.md
+++ b/README.md
@@ -112,6 +112,28 @@ For the full wallet walkthrough, read
 [`docs/AGENTIC_WALLET.md`](./docs/AGENTIC_WALLET.md) and
 [`QUICKSTART.md`](./QUICKSTART.md).
 
+## MCP server
+
+Agent clients that speak the Model Context Protocol can reach the same VFS
+commands over stdio, without a mount. It is **off by default**; enabling it is
+an explicit edit to `~/.bloom/config.toml`:
+
+```toml
+[mcp]
+enabled = true
+```
+
+```sh
+cargo run -p bloom -- mcp status   # is it on, and which socket would it use
+cargo run -p bloom -- mcp serve    # stdio; an MCP client spawns this
+```
+
+`mcp serve` adds no permissions of its own: it forwards `list`, `read`,
+`lookup`, `write`, and `write_with_lookup` to the daemon a mount would have
+used, so policy, confirmation, and audit gates apply unchanged. There is no
+network listener. See [`docs/guides/mcp.md`](./docs/guides/mcp.md) for the tool
+catalog, resource URI rules, and error codes.
+
 ## Development commands
 
 For local development, use the package-manager-native checks:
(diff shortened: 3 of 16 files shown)


violet-agent:
## Independent MCP/VFS surface review (Opus 5)

I traced the daemon VFS IPC dispatch against the new MCP surface and separately checked the cited implementation. The five command methods are covered (`lookup`, `read`, `write`, `write_with_lookup`, `list`), but the following issues remain:

1. **Medium — `vfs_stat` is advertised as destructive.**
   `crates/bloom-mcp/src/tools.rs:48-52` marks the pure `lookup` operation as `read_only: false`. `descriptor()` therefore emits `readOnlyHint: false` and `destructiveHint: true` (`tools.rs:147-160`). That contradicts the local mutation-boundary documentation at lines 24-27 and can cause clients to seek confirmation for metadata lookup. Set it to `read_only: true` and assert the advertised annotations.

2. **Medium — invalid JSON-RPC request shapes are silently dropped.**
   `crates/bloom-mcp/src/server.rs:93-102` treats any non-object JSON (including a batch array or scalar) as an ID-less notification and produces no response. A malformed request requires an `id: null`, `-32600` response; negotiated protocol `2025-03-26` also requires batch support. Either implement batches or reject them with a valid error, and add array/scalar regressions.

3. **Medium — resource reads use a misleading/unstandardized unavailable-daemon error.**
   `crates/bloom-mcp/src/backend.rs:15-18` assigns `-32002` to daemon-unreachable; `resources/read` forwards it as a JSON-RPC error (`server.rs:194-200`, `246-250`). That code is interpreted by MCP clients as resource-not-found, while an actual missing VFS path returns the daemon's `-32004`. Keep transport failures distinct from resource-not-found in the resource protocol surface, or return an MCP-compliant error with diagnostic data.

4. **Low/Medium — arbitrary VFS `read` is advertised as an MCP resource without the tool’s safety warning.**
   `crates/bloom-mcp/src/server.rs:186-216,223-232` makes `bloom:///{+path}` invoke `VfsMethod::Read`. The tool description itself notes some reads are side-effecting (`tools.rs:41-45`), but the resource template provides no comparable warning and clients commonly treat resources as inert. Consider exposing only known-inert paths as resources, or make the path-specific risk explicit and require the tool for potentially side-effecting reads.

5. **Low — resource URI paths do not round-trip reserved characters.**
   `crates/bloom-mcp/src/tools.rs:353-371` concatenates/parses raw path strings; it neither percent-encodes nor decodes. Valid VFS segments may contain `%`, spaces, `#`, or `?`, so a standards-compliant client can send a URI that reaches a different/nonexistent VFS path. Parse and serialize as URLs (or narrowly restrict the resource template) and test reserved characters.

6. **Low — the “future VFS capabilities appear for free” claim is inaccurate.**
   `crates/bloom-mcp/src/lib.rs:19` conflicts with the hardcoded per-method argument allowlist in `tools.rs:193-217`, which deliberately rejects any new daemon parameter. Current coverage is complete; correct the claim or establish a compatibility mechanism/test for added parameters.

Also noted: the new `bloom mcp` command and `[mcp]` config block have no README/quickstart/docs update.

These are review findings only; no code has been changed in this comment.

violet-agent:
## End-to-end MCP runtime smoke test

I ran the current PR head (`61af0e3f`) against an isolated `BLOOM_HOME` and a real local daemon over the configured Unix socket.

**Passed**

- Explicit `[mcp] enabled = true` was required and `bloom mcp status` reported the expected stdio transport and five tools.
- A real newline-delimited stdio JSON-RPC session completed `initialize`, `notifications/initialized`, `tools/list`, and `tools/call` → `vfs_list /`.
- The MCP server returned responses for IDs `11`, `12`, and `13`; it negotiated `2025-06-18`; and the live VFS list returned root entries including `wallets`, `chains`, `status`, `petals`, `requests`, and `AGENTS.md`.
- With the daemon stopped, `vfs_list` returned a valid MCP tool error result with `-32002`, without contaminating stdout/protocol framing.
- Focused regressions passed:
  - `cargo test -p bloom-mcp --test vfs_proxy a_stdio_session_answers_framed_requests_in_order -- --exact`
  - `cargo test -p bloom --test cli mcp_serve_proxies_vfs_commands_over_stdio_when_enabled -- --exact`

**Non-PR runtime findings**

1. A fresh isolated `bloom init` exceeded 180 seconds during network-backed preinstalled-Petal provisioning. The config was already written and MCP testing proceeded after explicitly enabling it. This did not block MCP startup or the live proxy test.
2. After SIGTERM, `bloom serve` logged `ipc.shutdown` and exited but left its Unix socket behind; the next request got `Connection refused`. This appears to be existing daemon shutdown cleanup behavior, not an MCP-proxy regression.

No repository files were modified by the smoke-test lane. The requested `gpt-6-astra` model could not be pinned for a Hermes delegated worker in this environment; the available delegated worker ran this test instead.

violet-agent:
## Follow-up: hardened MCP/VFS semantics

Follow-up commit: `60fbb9eb3b944dedd21fca0212b3834aceaaf84b` (`fix(mcp): harden protocol and resource semantics`).

Addressed review findings:

- Corrected `vfs_stat` metadata annotations and added an annotation regression.
- Added proper JSON-RPC invalid-request handling, including non-object frames and batches; all-notification batches remain silent as required.
- Separated MCP resource-not-found from daemon-unreachable/internal failures while preserving daemon diagnostics in error data.
- Preserved `vfs_read` as the full canonical VFS read tool, but resource reads now reject explicitly side-effecting paths before issuing the daemon call; inert paths remain available as resources.
- Added percent-encoded `bloom://` URI handling and reserved-character round-trip/refusal coverage.
- Corrected the forward-compatibility claim and added a schema/forwarded-argument consistency regression.
- Added `docs/guides/mcp.md` and linked it from the README.

Focused verification passed before the follow-up commit:

- `cargo fmt`
- `git diff --check`
- `cargo test -p bloom-mcp` — 39 unit tests + 11 integration tests passed.

### Important open point

`cargo test -p bloom --test cli mcp` could not complete after the follow-up because the host filesystem filled during linking and `ld` exited with `Bus error`. This is an environment/resource failure, not a reported test assertion/compiler failure; I removed only this worktree's 17 GB `target/` build artifacts to recover space, then committed/pushed. The prior head's CLI MCP tests passed, and the new behavior is covered in the `bloom-mcp` unit/integration suite, but the full CLI test needs a clean rerun in CI or on a host with sufficient linker scratch space.

violet-agent:
## Correction: wallet outbox control read semantics

My earlier statement was wrong: reading a wallet outbox `confirm`, `confirm.override`, `cancel`, or `replace` path **does not** sign, broadcast, cancel, or replace a transaction. They are virtual write sinks; only `vfs_write` with a non-empty payload performs the action.

I corrected the MCP descriptions, resource guidance, daemon-test wording, and docs in:

- `246a7d67` — corrects the proxy's action/read claims and resource message
- `422a7105` — adds a live daemon-backed regression for write-only control sinks and corrects the guide

The controls remain excluded from `resources/read`, not because their reads act, but because they are non-readable write sinks. The daemon marks them `read_side_effecting` defensively so speculative mount rendering also avoids them. The new regression proves that an MCP `vfs_read` reaches no action and fails as unreadable, while only `vfs_write` reaches the sink.

Verification:

```text
cargo fmt --check
cargo test -p bloom-mcp   # 40 unit + 12 integration passed
cargo clippy -p bloom-mcp --all-targets -- -D warnings
```

PR #275, opened 2026-09-14
Fix macOS upgrade rollback excluding private/etc archive members

0xdewy:
A failed macOS upgrade can leave newer configuration files behind instead of restoring the backup. This fixes rollback extraction so files under `private/etc` are restored, while retaining support for older backups using `etc` paths.

GNU tar's exclusion pattern matched both paths. The installer now extracts the listed ordinary entries explicitly and restores validated legacy entries separately. This is a one-function change.

Verification: the existing rollback harness failed on master and passed with this change under GNU tar on Linux; the rollback Rust gate and packet-filter retirement harness also passed. This does not establish macOS/BSD tar compatibility. The implementation ledger records the full release build at this PR's head as outstanding; macOS conformance remains blocked by the failing baseline lane and missing two-login runner inputs. Existing review and release gates still apply.

## Checklist

- [x] Tests added or updated for behavior changes — existing regression harness + rust gate (master-fail repro recorded)
- [x] Architecture docs updated — N/A (packaging fix, no contract change)
- [x] Sealed Approval invariants respected — N/A (installer rollback path)


josh-richardson (approved):
Reviewed the rollback extraction change and regression coverage. No P1/P2 findings; focused GNU tar, bsdtar, and Rust integration checks passed.

Issue #75, opened 2026-07-08
[defi]: enable by default


GavinPacini:
Currently it seems that a user / agent needs to enable the `/defi` Petal in their `config.toml`, this should be enabled by default. 

We should also ship sensible defaults in the wallet policy, namely around `allowed_routers`. Right now it seems an agent must first run a quote to get the allowed router, if our DeFi providers have known safe routers we should just populate the default config with those known routers.

AC: No configuration is needed to use defi.

0xdewy:
Related: #78 asks for the same default-enabled behavior for Hyperliquid. These can likely share one configuration/defaults pass for first-party petals, while preserving separate acceptance checks for DeFi router policy defaults.

PR #276, opened 2026-09-14
Let a Petal declare its signing fee asset in petal.toml

0xdewy:
Addresses the fee half of #242.

## Why

Some Petal signatures pay a chain fee from the signing key. Every Solana transaction does, so every Pump.fun buy, sell, close and sweep pays one. The Pump.fun Petal declares that fee in each signing claim, as the Broker requires.

The Broker checks a claim's fee against the operation class's entry in the provenance catalog:

| Catalog `fee_asset` | Claim declares no fee | Claim declares a fee |
|---|---|---|
| none | allowed | denied `FEE_NOT_ALLOWED` |
| `solana/native` | denied `FEE_REQUIRED` | allowed, and counted against the approval's value limits |

Developer enrollment (the hidden `triad-enroll-developer-petal-provenance` command, built with `triad-dev-harness`) catalogued every Petal class with no fee asset. So the Broker denied every Pump.fun claim with `FEE_NOT_ALLOWED`, and no approval could fix that.

Bloom can't hardcode which classes pay fees without knowing each Petal's class names. The Petal knows, so it now declares it in `petal.toml`:

```toml
[sign]
allowed_intents = ["pumpfun.buy", "pumpfun.sell"]
fee_asset = { chain = "solana", asset = "native" }
```

Other Petals omit the field. They sign off-chain messages whose gas someone else pays (Polymarket orders and relayer batches, Safe confirmations, x402 authorizations), or Hyperliquid actions, which pay no gas. They correctly declare no fee.

## What changed

- **Manifest.** `bloom-petals` parses `[sign].fee_asset` and records it in the route index as `sign_fee_asset`. The field is omitted when absent, so existing packages keep the same route index and hash. It is part of `petal.toml`, so the package policy hash covers it.
- **Developer enrollment** catalogs every class of the package with that fee asset. Release provenance is unchanged and still catalogs no fee asset; adding a fee-paying Petal to a release is a separate change.
- **CI.** The unit-test lane now also runs unit tests in binary targets (`kind(bin)`). The enrollment tests live in the bin-only `bloom` crate, so CI never ran them. They now compile under `cfg(test)` as well as `triad-dev-harness`.
- **Test isolation.** Enrollment rebuilds a package's artifacts in place. The enrollment test now enrolls a temporary copy of `tests/fixtures/triad-authority-petal`, so it no longer rewrites a fixture other tests read in parallel.
- **Docs.** `docs/petals/authoring-petals.md` documents the field.

The fee asset applies to the whole package, not to each intent. That matches Pump.fun, where every transaction pays the fee. A Petal that mixes fee-paying and fee-free signatures in one package can't use it; a per-class field could be added later without changing this one.

## Compatibility

Bloom v0.3.0 rejects unknown `[sign]` fields, so a Petal that sets `fee_asset` does not install on v0.3.0. Petals without the field are unaffected, including after a rollback, because their route index is unchanged.

## Not in this PR

This alone does not make Pump.fun work on master:

- #302: session budgets. Master rejects the Petal's `approval_value_limits`, and a session approval with no value limits denies any claim that declares a fee.
- #304: close and sweep, which use Exact signing with the Petal's session key.
- bloom-directory/bloom-petal-pumpfun#5: adds `fee_asset` to the Pump.fun manifest.

It also does not unblock Hyperliquid builder-fee orders from #242. Those share the `hyperliquid.agent_action` class with fee-free actions, which a package-wide field can't separate.

## Verification

- `bloom-petals` lib and integration tests and `bloom` unit tests pass, including the six enrollment tests. A new test covers parsing an absent, a valid and an invalid fee asset. The route-class test checks that every class gets the declared asset, or none when it is absent.
- Clippy (`--all-targets -D warnings`) is clean with and without `triad-dev-harness`.
- Real services (merged locally with #302 and #304; Broker `dd2add2b`, Signer bloom-signer#57): developer enrollment of Pump.fun and an Ed25519 fixture that both declare `fee_asset`; fee-declaring claims accepted, with the fee counted against sealed budgets (a debit that fits only without its fee is refused); Pump.fun close and sweep claims land on a local validator. Evidence in the workspace `shared/pumpfun-verification-2026-09-17/`.

## Dependencies and landing

- **Depends on:** nothing; based on master.
- **Order:** first of #276 → #302 → #304 (a choice, not a code dependency); merge master forward into the next after each lands.
- **Pins:** none change in this PR (Broker `dd2add2b`, Signer `ccc9adb3`, runtime `5db670e1`). The Broker and Signer pins advance in a separate Bloom PR after bloom-directory/bloom-broker#74 and bloom-directory/bloom-signer#57 land.
- **Release:** Pump.fun (bloom-directory/bloom-petal-pumpfun#5) needs a Bloom release containing #276, #302 and #304. Recovery also requires the fixes in bloom-directory/bloom-broker#76 (for #75) and bloom-directory/bloom-service-runtime#21 (for #20), consistently pinned into the release.
- Full sequence: workspace `shared/pumpfun-verification-2026-09-17/LANDING.md`.

## Integration status

[PM #66](https://github.com/bloom-directory/pm/issues/66) owns reviews, landing, consistent pins and final release checks. [PM #35](https://github.com/bloom-directory/pm/issues/35) owns mainnet acceptance, which remains unverified. Optimized, test-enabled candidate binaries passed Part A (32/0); local combinations and Bloom's runtime patch do not establish final release conformance.

## Checklist

- [x] Architecture docs updated — N/A; the Petal authoring guide documents `[sign].fee_asset`
- [x] Sealed Approval invariants respected — no signing-path change; the Broker's fee checks are unchanged


0xdewy (commented):
## Independent review (audit of pm#35)

**Verdict: clean; one follow-up to track, none blocking this PR.**

- `[sign].fee_asset` parses into `SignPolicy`, flows into `RouteIndex.sign_fee_asset`, and developer enrollment marks every operation class fee-bearing via `developer_route_operation_classes`. Validation rejects malformed chain names; the route index omits the field when unset (no serialization churn); tests cover present/absent/invalid; `authoring-petals.md` documents the field and its semantics. CI widening to `kind(bin)` is what makes the enrollment tests actually run.
- **Known gap (documented in this PR, confirmed by reading the code):** the release path — the static `PreinstalledPetal` catalog in `github_source.rs` feeding triad enrollment — hardcodes `fee_asset: None` per class, and `authority.rs` matches `class.fee_asset` against the claim's declared fee. So a fee-bearing Petal installed from a release catalog would have claims that its provenance classes do not authorize. pumpfun is not yet in the catalog, so nothing is broken today, but registering pumpfun as a preinstalled release petal requires extending the catalog entries to carry the fee asset first. This belongs on pm#35's remaining-work list rather than in this PR.

---
*Reviewed by Lotus (Hermes Agent) as part of the pm#35 audit.*

PR #277, opened 2026-09-14
Update rustls to 0.23.45 for RUSTSEC-2026-0285

0xdewy:
## Summary

Bloom's TLS library, rustls 0.23.40, is affected by RUSTSEC-2026-0285: it accepts some TLS handshake messages at the wrong encryption level. This PR updates rustls to 0.23.45, the first fixed release. No code or manifests change; only `Cargo.lock` does.

rustls 0.23.45 requires newer versions of its crypto backend, so `aws-lc-rs` moves from 1.17.0 to 1.18.1 and `aws-lc-sys` from 0.41.0 to 0.45.0, which pulls in `pkg-config` as a build dependency. `rustls-webpki` moves from 0.103.13 to 0.103.15. Nothing else in the lockfile changes. An earlier revision of this branch also moved tempfile from getrandom 0.3.4 to 0.4.2. That change wasn't needed and has been reverted.

The Signer has its own lockfile and is updated in bloom-directory/bloom-signer#53.

## Verification

- `cargo audit`: RUSTSEC-2026-0285 no longer appears. The other six findings are the same as on master (tracing-subscriber, quinn-proto, ruint, rkyv, h2, wasmtime).
- `cargo check --workspace --all-targets --locked` passes at 963be719.
- A Linux release build with Broker 57cbc6c6 and Signer 941dd376 passed at 4e199950, including bundle verification and installer smoke tests. It hasn't been rerun since the getrandom revert.
- Not run: a live TLS handshake test, and the macOS conformance lanes. The Unix-principal lane also fails on master, and the two-login lane needs self-hosted runner inputs.
- `bloom-worktree pins` flags the stale Broker ref in the macOS workflow inherited from master; #278 fixes it.

## Checklist

- [x] Tests added or updated for behavior changes — N/A, lockfile only
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A
- [x] Sealed Approval invariants respected — N/A


josh-richardson (approved):

PR #278, opened 2026-09-15
Fix stale protocol versions and Broker refs in the release manifest

0xdewy:
## Summary

The release compatibility manifest says the Broker and Signer speak authority protocol 1.4. The revisions it pins actually speak 1.6 (Machine–Broker) and 1.5 (Broker–Signer), so a bundle built from it describes itself wrongly. This PR updates the manifest and the bundle verifier to 1.6 and 1.5. It adds a test that catches the next time these numbers drift.

The protocols themselves don't change; this only makes the release metadata match the code it already ships.

## What changed

- **Protocol ranges.** `compatibility-v1.toml` and `verify-bundle.sh` now require Machine–Broker 1.6 and Broker–Signer 1.5. Signer-control and session ranges were already correct (1.0–1.1).
- **Signer backends.** The manifest listed `aws-kms`, but the release builds the Signer with default features, which include only `local`. It now lists `["local"]`.
- **Drift test.** `triad_release` used to compare the manifest against hard-coded numbers, which is how 1.4 went stale. It now compares all four ranges against the constants in the Broker, Signer and service-runtime API crates, plus the verifier's exact literals. It also checks that `Cargo.lock` resolves those crates at the revisions the manifest records, so the constants really are the shipped ones.
- **Stale Broker ref.** The macOS conformance workflows and the CI caller still defaulted to Broker bd85885b; they now use the manifest's 57cbc6c6.
- **TESTING.md.** It named two deleted authority-boundary scripts; it now points at `cargo test -p bloom-it --test triad_release`.

## Not in this PR

- Protocol bumps: none. Protocol minors change only at release builds.
- The Linux installer still accepts AWS KMS configuration (#281).

## Verification

- `release_compatibility_declares_each_edge_without_a_global_protocol_range` passes at a55907f6. Setting Broker–Signer back to 1.4 in the manifest makes it fail with `missing [protocols.broker_signer]`.
- Constants read at the pinned revisions: broker 57cbc6c6 `BROKER_API_MINOR_MIN/MAX = 6`; signer 941dd376 `SIGNER_API_MINOR_MIN/MAX = 5`, `SIGNER_CONTROL_RANGE = 1.0–1.1`; service-runtime bc88cc67 `SESSION_PROTOCOL_RANGE = 1.0–1.1`.
- `bloom-worktree pins` passes at a55907f6.
- Not run: the full Linux release build and bundle verification, and the macOS conformance lanes.

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A, release metadata only; release README updated
- [x] Sealed Approval invariants respected — N/A, no signing or approval changes
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — N/A, no agent-facing surface changed


josh-richardson (approved):

PR #279, opened 2026-09-15
List mount names that LOOKUP resolves back to the same segment

0xdewy:
## Summary

In the mounted Bloom filesystem, some files that `ls` showed couldn't be opened. Mount lookups percent-decode names (so `hello%20world` means `hello world`), but directory listings returned names unescaped. A listed `100%done` then failed to open, and a listed `a%41` opened `aA` instead. Listings now escape `%`, `/` and NUL, so every listed name opens the entry it came from. For example, `100%done` is listed as `100%25done`.

The same PR fixes going up a directory (`..`) from inside a directory whose name contains `/`. The parent path used to be rebuilt by joining segments with `/` and parsing the result, which split a directory named `a/b%` into `a` and `b%`. It now reuses the adapter's existing segment-preserving `parent_path` helper.

## Verification

- `cargo test -p bloom-mount --features mount --lib --locked`: 78 passed at 17499204.
- `listed_names_decode_to_their_segment` checks that decoding each listed name gives back the original, including `%`, `/`, NUL and non-ASCII names.
- `listed_directory_parent_preserves_encoded_segments` lists, opens and walks up and down a directory named `a/b%`. With the old parent code it fails: the parent resolves to `["nested", "a", "b%"]`.
- Not run: a real kernel mount. These are adapter tests with a fixture handler.
- `bloom-worktree pins` flags the stale Broker ref in the macOS workflow inherited from master; #278 fixes it.

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A, no contract change
- [x] Sealed Approval invariants respected — N/A, no signing or approval changes
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — N/A, no documented path changes


josh-richardson (approved):

PR #282, opened 2026-09-15
fix: normalize wallet VFS account layout

violet-agent:
## Summary

Normalize the wallet VFS around a strict ownership boundary:

- **Wallet-wide state** remains at `wallets/<wallet>/`.
- **Account/key-specific state** lives only at `wallets/<wallet>/<index>/`.
- **Installed Petals** remain root-scoped at `petals/`.

This PR removes obsolete or ambiguous namespace surfaces rather than retaining aliases.

### Wallet and account layout

- Remove account-bound identity files from wallet roots and expose numbered account directories instead.
- Keep wallet-level controls, policy/projection state, and account-zero chain aliases at the wallet root.
- Preserve account-scoped `sessions/`; remove account-scoped Petal dispatch.
- Update VFS lookup/list/read/write routing, daemon seams, cache paths, documentation, examples, and integration/conformance tooling for the indexed tree.

### Account identity cleanup

- Delete `addresses.json` completely. It was a misleading legacy aggregate: it duplicated a single display address as `owner`/`signer`, mixed wallet-level policy state into an account leaf, and carried fixed/empty `unlocked` and `roles` fields.
- Move policy consumers to the wallet-level `projection.json` surface.
- Replace generic account address aliases with explicit family leaves, emitted only when that family exists:
  - `address.evm`, `address.evm.qr.svg`, `address.evm.qr.png`
  - `address.sol`, `address.sol.qr.svg`, `address.sol.qr.png`
- Retain `account.json` as the canonical structured multi-family account record: account number, key references, fingerprints, derivation paths, lifecycle, and family state.
- Do not provide compatibility aliases for the retired paths; they have not shipped in a release.

### Legacy/imported wallet behavior

A legacy or imported single-key wallet remains a real indexed account: it is projected as account `0`, in the root key's own family. Its identity is exposed at e.g. `wallets/<wallet>/0/address.evm` for an EVM root key. No seed, private key, or legacy local authority surface is introduced.

## Validation

- [x] `cargo fmt`
- [x] `git diff --check`
- [x] Shell syntax validation for changed integration, fixture, and conformance scripts
- [x] `CARGO_TARGET_DIR=/root/.cache/bloom-target-vfs-layout cargo test -p bloom-vfs handlers::wallets --lib` — 84 passed
- [x] `CARGO_TARGET_DIR=/root/.cache/bloom-target-vfs-layout cargo test -p bloom-vfs --test triad_policy_update` — 14 passed
- [x] Prior follow-on: `cargo check -p bloom-daemon -p bloom-petals`
- [x] Hosted CI passed for this PR head before the chain-scope follow-on was split into PR #283.

## Checklist

- [x] Tests added or updated for layout, family-specific address listing/lookup/read, and absence of generic/aggregate address leaves.
- [x] Architecture, mounted agent guidance, quickstart, examples, audit guidance, and Harbor instructions updated for the revised VFS contract.
- [x] Sealed Approval invariants respected — this changes public VFS namespace/projections only; it does not alter signing, Broker authority, or grant semantics.
- [x] No backward-compatibility aliases or legacy-path documentation retained for unshipped surfaces.


josh-richardson:
Found one merge-blocking issue:

**[P1] Required integration and release scripts still use the removed wallet-root paths.** After this change, these reads will fail:

- `scripts/local-mainnet-integration.sh:336,428`
- `scripts/test-triad-projection-fidelity.sh:529` — invoked by `scripts/acceptance.sh`
- `tests/conformance/macos-unix-principals/run-packaged-machine-negative.sh:802`

The fake integration launcher also creates the obsolete paths at `scripts/test-fixtures/fake-triad-dev-launcher.sh:112,122,220`, which lets local integration tests pass while masking incompatibility with the real VFS. These consumers and fixtures should use `wallets/<wallet>/0/{address,addresses.json}` before the wallet-root leaves are removed.

Verification: `cargo test -p bloom-vfs --locked` passed at the PR head (404 unit tests and 14 integration tests). I found no additional code-level issues.

violet-agent:
Fixed in 3bd109859d960a88a133643699ae333345999936. Updated the four required consumers and the fake Triad launcher to use `wallets/<wallet>/0/{address,addresses.json}`; also corrected the Harbor eval instructions and audit round-trip example found by the same sweep. `bash -n` passes for every changed shell script and `git diff --check` is clean. I attempted `scripts/test-triad-projection-fidelity.sh`, but its developer harness correctly refuses to run as root in this environment; hosted CI will exercise it under its supported runner.

josh-richardson:
Re-review found one new merge-blocking issue:

**[P1] `address.sol` is blank when a valid Solana account has no chain projection.** In `crates/bloom-vfs/src/handlers/wallets/accounts.rs`, `read_account_key_file` reads the Solana address from `FamilyKey.address`, but that field is populated using `chain_projections.first()` or an empty string. A chain projection may legitimately be absent when no matching chain target is configured. The VFS still lists `address.sol`, but reading it then returns only `"\n"`.

A focused regression reproduced this using a valid Ed25519 SPKI account with `chain_projections: []`: expected the base58 address derived from the public key, received `[10]`. The address should be derived from the canonical Ed25519 public key—as `SolanaAccount::from_projection` already does—and any supplied projection should be validated against it.

The previous stale-path issue is fixed. Other verification at current head:

- `cargo test -p bloom-vfs --locked`: 404 unit tests and 14 integration tests passed.
- `scripts/test-local-mainnet-integration.sh` passed.
- Changed shell scripts passed `bash -n`.

GavinPacini (approved):

Issue #286, opened 2026-09-15
Numbered-account Solana pending controls return ENOENT through the mount but resolve over IPC


0xdewy:
## Summary

Under a numbered account path, a pending Solana outbox entry's `cancel` and `confirm` controls are listed but return ENOENT through the kernel mount. The same paths resolve over IPC.

## Reproduction

On master (Bloom `4e37a635`), Linux install, mainnet config. Stage from the wallet-level path, which is account 0:

```
printf '{"destination":"<account-1 address>","lamports":1000000}' > ~/bloom/wallets/main/chains/solana-mainnet/outbox/new.tx
```

Then, for the new id:

| Path | Kernel mount | `bloom vfs stat` (IPC) |
| --- | --- | --- |
| `wallets/main/chains/solana-mainnet/outbox/pending/<id>/cancel` | regular empty file | resolves |
| `wallets/main/0/chains/solana-mainnet/outbox/pending/<id>/` (dir) | directory; `ls` lists `cancel confirm intent.json plan.md restage` | resolves |
| `wallets/main/0/…/pending/<id>/intent.json` | regular file | resolves |
| `wallets/main/0/…/pending/<id>/cancel` | `No such file or directory` | resolves |
| `wallets/main/0/…/pending/<id>/confirm` | `No such file or directory` | resolves |

This reproduced on two fresh entries (`sol-5c9d5fd6…`, `sol-0082b3f5…`), with and without a prior `ls`, and after a 3 s wait. `bloom wallet cancel` works for the same entry.

## Notes

- The mount adapter's `lookup` calls the same `vfs.lookup` IPC uses, so the difference is in the mount path, not in `solana_outbox_lookup`, which resolves pending controls for `OutboxScope::Key`.
- The wallet-level spelling of Solana `cancel` is separately refused by the mount transport guard (`Operation not permitted`), so on the mount neither spelling can cancel a Solana transfer today.

0xdewy:
The last note above is out of date since #283, and it changes what a fix here has to cover.

#283 moved both transport guards from the wallet-level spelling to the numbered one. On master, `write_path_uses_wallet_signer` (`crates/bloom-daemon/src/ipc.rs`) and `mount_write_path_uses_wallet_signer` (`crates/bloom-mount/src/adapter.rs`) match `wallets/<w>/<account>/chains/<any chain>/outbox/pending/<id>/{cancel,replace}`, and #282 made the handler deny wallet-level outbox writes.

[...]

PR #290, opened 2026-09-15
Prove outbox cancel and replace stay behind Broker approval for every route spelling

0xdewy:
## Summary

Tests only; no production code changes. They show that outbox `cancel` and `replace` stay behind the handler and Broker exact approval for both spellings of account 0's route (`wallets/<w>/chains/…` and `wallets/<w>/0/chains/…`). They are the evidence needed to decide on the duplicated transport guards described below.

- **`triad_cancel_signs_only_the_exact_self_send_after_its_approval`** (engine)
  - Before an active approval: no `SigningSign`, and no raw tx, broadcast marker or hash.
  - After approval: exactly one sign request, with the staged sender's key.
  - The payload equals the same-nonce, zero-value self-send with fees raised 10%.
- **`cancel_fails_closed_without_a_reachable_broker`** (engine): no signing edge, or a Broker that refuses every call, leaves no artifacts.
- **`substituted_replacement_is_denied_before_any_broker_request`** (engine): a VFS `replace` always substitutes an intent, and is hard-denied with zero Broker requests.
- **`evm_cancel_and_replace_fail_closed_for_both_account_zero_spellings`** (handler, through `Vfs`, the entry point IPC writes and mount flushes share)
  - Runs with and without a signing edge that carries the `transaction.cancel`/`transaction.replace` provenance both shipped catalogs grant.
  - Asserts the Broker was reached, with no `SigningSign*` and no artifacts.
  - Account 1's entry is NotFound through `/0/`.
- **`solana_cancel_uses_numbered_account_zero_while_wallet_level_writes_are_denied`** (handler)
  - The `/0/` spelling cancels an unbroadcast entry; a wallet-level write is denied (`PermissionDenied`) without mutating the entry (normalized account layout, #282).
  - Another account's entry and a durably broadcast entry are refused through the numbered route.

## Evidence from a real install (mainnet, 2026-09-15)

On master (Bloom `4e37a635`, Broker `57cbc6c6`, Signer `941dd376`), installed on Linux:

- **Mounted wallet-level cancel:** writing `wallets/main/chains/base/outbox/pending/0005-72611/cancel` returns `Operation not permitted` from the mount transport guard. The same happens for Solana `…/solana-mainnet/outbox/pending/<id>/cancel`, which signs nothing.
- **CLI cancel, same operation:** `bloom wallet cancel main base 0005-72611` goes through `machine.execute` to the same handler with no guard.
  - It required an exact Broker approval ceremony.
  - After the owner approved, it broadcast `0x990cb8ff…10659`: from = to = the wallet, value 0, nonce 2, max fee 13,200,000 = the staged 12,000,000 × 1.1.
- **Numbered `/0/` Solana cancel on the mount:** the listing shows `cancel`, but lookup returns ENOENT (#286). Mount users currently cannot cancel a Solana transfer by either spelling.

## Decision requested: remove the duplicated transport guards

`write_path_uses_wallet_signer` (`crates/bloom-daemon/src/ipc.rs`, used by `write` and `write_with_lookup`) and `mount_write_path_uses_wallet_signer` (`crates/bloom-mount/src/adapter.rs`, used by flush, WRITE, CREATE and OPEN) refuse only one spelling.

**Recommendation: remove both** (the root reviewer's option A), because:

1. **They add no protection.** `wallets/<w>/0/chains/…` bypasses both, and `machine.execute WalletOutboxCancel/Replace` on the same socket and uid performs the identical handler write.
2. **Authorization lives in the handler and Broker.** EVM cancel/replace require installer provenance plus a Broker exact approval, and replace with a substituted intent is hard-denied (tests above, and the live cancel above).
3. **The `sign/*` arms are dead.** The handler refuses `wallets/<w>/sign/*` itself (`legacy_wallet_sign_surface_is_absent`).
4. **Their stated reason is gone.** The guard comment points at `write_unlocked`, which no longer exists.
5. **They block a documented, non-signing operation:** Solana cancel on the mount.

Removal grants nothing that the CLI and `/0/` don't already allow today. The deletion is 169 lines across the two guards, their call sites and the classifier tests. It is **not** in this PR, because the owner has not approved it.

Before that follow-up can be reviewed, it must include transport tests through the real `WalletsHandler` and an authenticated `IpcServer::serve` socket, not only a recording handler. The follow-up covers IPC `write`, IPC `write_with_lookup`, a mounted FILE_SYNC WRITE and UNSTABLE WRITE + COMMIT through `BloomFs`, each over both spellings:

| Case | Expected |
| --- | --- |
| EVM cancel, Broker refuses or is unreachable | error; no `SigningSign*`; no `raw_tx`, `broadcast_attempted`, `*_tx_hash`, `*_intent.json` |
| EVM cancel, no active approval | approval-required error; Broker saw a prepare, no sign |
| EVM replace with substituted intent | refused before any Broker request |
| Account 1's EVM entry via account 0's routes | NotFound |
| Solana cancel, pending and unbroadcast | both spellings cancel |
| Solana cancel, other account or `broadcast_attempted.json` present | refused |
| `wallets/<w>/sign/{message,hash,typed_data}` | -32007 / EACCES |

Two related changes, each its own decision:
- **Error mapping:** map `TxEngineError::ApprovalRequired` to `PermissionDenied` for cancel/replace, as `confirm` does. Today it maps through `err_be` (IPC -32000, mount EIO).
- **Empty Solana cancel body:** the Solana cancel handler ignores the body, so once the guard is gone a bare `touch …/cancel` (CREATE writes an empty body) would cancel a Solana transfer. Today the guard and #286 prevent that on the mount; the CLI already cancels with any body. Rejecting an empty body changes behavior and should be justified and tested separately.

## Verification

Ran after merging current master (`23e5ff4d`), in an isolated build directory:
- `cargo test --locked -p bloom-vfs --lib`
- `cargo test --locked -p bloom-tx --lib`
- `cargo test --locked -p bloom-mount --features mount --lib`
- `cargo clippy --locked -p bloom-vfs -p bloom-tx -p bloom-mount --all-targets -- -D warnings`

Results: 411, 156 and 78 tests passed, with no failures; clippy and fmt are clean.

## Checklist

      behavior changed — N/A, tests only
      bytes) — tests assert them; no production change
      and affected Petal READMEs) — N/A


0xdewy:
Closing without merging. This PR changes no production code; it exists to support removing the mount and IPC transport guards on outbox `cancel`/`replace`, and that removal is not being done here.

Part of the description is also stale: since #283 the guards match the numbered spelling (`wallets/<w>/0/chains/…`), not the wallet-level one, so the "only one spelling is refused, `/0/` bypasses both" argument no longer holds.

The part that still matters, that the guards block Solana cancel on the mount, is now recorded on #286. The branch is kept so the three engine tests in `crates/bloom-tx/src/tx_engine.rs` can be reused by whichever change removes or narrows the guards.

PR #293, opened 2026-09-15
feat(default-policy): adding a default policy to allow users to transact faster with less ceremonies

jer117:
### Default wallet policy

Petals are strict today: a new user needs a separate policy ceremony for each
Petal, and then another denial when that Petal tries to touch a contract. This
branch lets a user choose their Petals once during `bloom init`, then approve
**one** policy for their `main` wallet that allows all of them *and* the
contracts they transact with. Every transaction still needs its own approval,
like a normal wallet.

### How it works for a user

1. `bloom init` records Bloom's five canonical Petals — Polymarket,
   Hyperliquid, Enso, NEAR Intents and Tolly — as the wallet's starting set,
   with Polymarket's daily buy limit defaulting to 100 pUSD. A terminal run
   confirms each one and lets the limit be changed; anything non-interactive
   (piped input, CI, an agent) takes those defaults without asking. The
   choices are saved under `[petals.setup]`, and every canonical Petal is
   installed regardless — the answers decide only what the first policy
   proposes.
2. `bloom wallet new main` or `bloom wallet import main` runs the normal
   wallet ceremony, then opens one policy ceremony allowing every chosen
   Petal package plus the fixed contracts those Petals use.
3. After that approval the user can transact. Each order or swap still asks
   for its own approval.

If the user steps away, `bloom wallet default-policy main` picks up where they
left off: it shows the pending approval link, or reports the policy is already
applied.

## Allowing the contracts, not just the packages

Allowing a Petal without its contracts leaves it installed but unusable:
Machine refuses any outbox transaction whose destination is not in
`allowed_destinations`, and an empty list refuses everything. So the same
ceremony now carries the destinations each chosen Petal needs
(`crates/bloom-proto/src/petal_destinations.rs`):

- **Enso** — Router V2 on 13 chains. Every address checked on-chain with
  `eth_getCode`: identical bytecode wherever the address repeats, a distinct
  deployment on Linea, and a shared one across Arc/Robinhood/Tempo.
- **Polymarket** — pUSD, native USDC, USDC.e and the router on Polygon.
- **NEAR Intents** — see below.

These live in Bloom's source, never in config and never in anything a Petal
supplies at runtime, and the owner still approves the ceremony that adds them.

## A Petal that chooses its own destination

NEAR Intents mints a **fresh deposit address for every quote**, signed by the
1Click API, so no address can be listed in advance. A catalog destination may
now name a Petal instead: `petal:near-intents`, on the eight chains its
release maps. The recipients check passes only when the Petal that staged the
transaction is the one the policy names.

What that does not do: it covers only that Petal, only on those chains, and
only for transactions it staged. Another Petal, or a transaction staged by
hand, still has to match a listed address — verified live, a hand-staged
transfer to an unlisted address was denied on a chain carrying the entry.
Every deposit still needs its own approval, and removing the entry gates the
Petal again.

What backs it: the Petal's own venue policy caps input size and slippage and
pays out only to the wallet's own address unless the owner lists others, and
the deposit address comes from a quote whose signature the Petal verifies. The
trade-off is that for that Petal, on those chains, the destination check moves
from Bloom's list to the Petal's own logic. `docs/architecture/Default Wallet
Policy.md` decision 14 records this, including the ordering: the destination
check runs at staging, while the package check runs later at confirm.

## What changed

- **Setup menu** (`crates/bloom/src/default_policy.rs`) — Petal choices and
  settings asked during `bloom init`. Non-interactive runs accept suggestions.
- **Config** (`bloom-proto`) — `[petals.setup.<name>]` with chosen values.
- **Petal catalog** (`github_source.rs`) — Polymarket declares its settings
  template; Tolly's writes are enabled when chosen.
- **Destinations** (`bloom-proto/petal_destinations.rs`) — the per-Petal
  contract table, plus the `petal:<name>` class.
- **Policy engine** (`bloom-tx`) — the recipients check honours a `petal:`
  entry when the staging Petal matches.
- **Settings re-applied after updates** — updating a Petal resets its private
  settings, so Bloom rewrites the saved setup values and says so.
- **One policy for several Petals** (`bloom-vfs`, `bloom-daemon`) — the
  existing policy-update flow proposes several packages and their
  destinations in one ceremony. Limits stay inside each Petal; the Broker is
  unchanged.
- **CLI** — `wallet new` / `wallet import` wait for the default policy, and
  `bloom wallet default-policy <name>` resumes it.
- **Docs** — `docs/architecture/Default Wallet Policy.md`.
- **Test script** — `scripts/test-default-policy-e2e.sh` starts a throwaway
  triad, creates `main`, completes both ceremonies with the Broker debug
  driver, and checks the resulting policy.

## Bugs found while testing on a local triad

1. **Two waiting commands knocked each other out.** Two commands waiting on
   the same wallet's policy could leave one exiting with "wallet policy
   coordination busy". It now retries briefly so both finish.
2. **Expired approval links weren't replaced.** After an unapproved link
   expired, `bloom wallet default-policy main` returned the dead link and then
   failed with "ceremony is terminal: Expired". Bloom now replaces a
   cancelled, expired or failed request and never shows an expired link.

## Testing

`cargo fmt --check` and workspace `cargo clippy -D warnings` pass.

- `default_policy` unit tests: 23
- `bloom` bin: 89 · `bloom-vfs`: 410 + 22 (`triad_policy_update`)
- `bloom-daemon`: 100 + 3 · `bloom-proto`: 170 · `bloom-tx`: 156
- `scripts/test-default-policy-e2e.sh` passes on a local triad

**Live on mainnet, on a triad at v0.3.1**, one policy ceremony allowed 7
packages and 24 destinations, then a NEAR Intents swap ran end to end with no
further policy ceremony:

- quote → all 8 venue-policy checks passed
- staged to a per-quote deposit address, outbox verdict
  `destination chosen by petal:near-intents, which wallet policy allows`
- one Sealed Approval, one broadcast: 0.0008 ETH on Arbitrum (block
  507430398)
- 1Click reported `SUCCESS`; **2.169889 USDC arrived on Base** (block
  51600779), matching the quote

An Enso swap on Arbitrum staged cleanly in the same wallet with no extra
ceremony, and a hand-staged transfer to an unlisted address was denied.

## Known gaps, not fixed here

- **Settlement reporting for NEAR** depends on #312. A Petal that stages in
  one route and polls in another cannot inspect its own outbox entry, so the
  swap above completed while the Petal still reported `staged`. Swaps work;
  the reporting is blind until #312 merges.
- **Lost approval message**: a Petal write needing approval shows a bare
  "permission denied". Overlaps #214, which adds
  `HandlerError::ApprovalRequired`; a small `router.rs` follow-up passes the
  Petal's details through once that merges.
- **Outbox durability** (#310) — unrelated to policy, tracked separately.
- **Hyperliquid refusals** aren't recorded in `last_response.json` — that Petal's repo.

## Scope

Repos: bloom only. No Broker, Signer or Petal changes.

## Checklist

      behavior changed


crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -180,9 +183,9 @@ impl TriadPolicyUpdateProjection {
                     challenge_path: format!(
                         "/wallets/{wallet}/policy-updates/pending/{operation}/{APPROVAL_CHALLENGE_FILE}"
                     ),
-                    includes_requested_package: proposed
-                        .allowed_petal_packages
-                        .contains(package_hash),
+                    includes_requested_package: required
```
violet-agent: `pending_view` only verifies `required` package hashes, even though this caller also requires every `destinations` entry. A pending ceremony that contains the selected Petal packages but omits one of their fixed destinations is therefore returned with `includes_requested_package = true`; `advance_default_policy` labels it as the chosen default policy and hands its approval URL to the user. Once that ceremony commits, the next poll must open a second policy ceremony to add the missing destinations. That violates this PR's one-ceremony guarantee and, more importantly, presents the first ceremony as making the Petals usable when their outbox destinations are still denied. Thread the requested destinations through `pending_view`/`PendingPolicyUpdate` and require both package and destination containment before setting `proposes_chosen_petals` true (with a regression test for a pending policy containing all packages but not all destinations).

violet-agent (changes requested):
Requesting changes for the inline finding: a pending policy proposal is treated as the default policy when it includes packages but omits required destinations, causing a second ceremony and a misleading “usable” claim. Focused tests and the latest hosted CI checks otherwise passed.

jer117:
Agreed, fixed. pending_view now takes the requested destinations and requires both package and destination containment. It's judged against what the caller requested: ensure_policy_allows was shadowing `destinations` with the first-proposal extras, so I split that into `proposed_destinations`. I renamed the field to includes_requested since it now covers both. Added the regression test you asked for; I confirmed it fails without the destination clause and passes with it.

PR #284, opened 2026-09-15
Release v0.3.0

GavinPacini:
Prepare the Bloom v0.3.0 release. The first commit bumps the shared workspace version, workspace lockfile entries, and release compatibility manifest to 0.3.0.

## Release checklist

- [x] Update version to v0.3.0 (completed in the first commit).
- [x] Update all featured petals.
- [x] Add Tolly as a featured petal.
- [x] Bump all Bloom deps to latest master head.
- [x] Repin Enso petal.
- [x] Repin Near petal.

## Validation

- `cargo metadata --offline --locked --no-deps --format-version 1` passed.
- Verified all 26 workspace packages report 0.3.0 and the lockfile changes only their versions.
- `cargo test -p bloom-proto --lib --locked` passed (169 tests).
- Focused Bloom catalog, release-manifest, prebuilt-archive, and provisioning tests passed.
- `cargo clippy -p bloom -p bloom-proto --all-targets --locked -- -D warnings` passed.
- `cargo fmt --all -- --check` passed.
- `git diff --check` passed.

The dependency-head bump remains intentionally outstanding.

PR #300, opened 2026-09-16
Add missing Arc and Solana configs; remove allow_broadcast gates

GavinPacini:
## Summary

Existing Bloom configurations gain Arc and Solana mainnet defaults when those network names are absent. Existing entries retain their endpoints, chain identities, and genesis pins; names already assigned to the other chain family are preserved. The default Solana mainnet entry includes its genesis pin.

Remove `allow_broadcast` from EVM and Solana configuration and all runtime gates. Existing values, including `false`, are ignored, and generated or saved configs omit the field. Broadcasting is available on every configured chain, subject to signing, policy, confirmation, and Sealed Approval checks.

Unpinned custom Solana networks remain usable for reads. Submission requires a pinned genesis verified against every endpoint, and missing pins fail before a broadcast attempt is recorded. Ambiguous sends retain their reconciliation behavior. Solana status keeps its schema: `broadcast.enabled` is always true, while `broadcast.eligible` reflects live genesis verification.

Updates operator documentation, mounted agent guidance, and generated test configs. Targets PR #284 (`release/v0.3.0`).

## Validation

- Affected package suites passed: `bloom-proto`, `bloom-solana`, `bloom-solana-tx`, `bloom-tx`, and `bloom-vfs`, including embedded documentation tests.
- Regression coverage verifies ignored legacy values, omission from serialized configs, successful pinned Solana submission with legacy `false`, missing-genesis rejection, and preservation of custom network settings.
- `cargo clippy --workspace --all-targets --locked -- -D warnings` passed.
- Formatting, diff checks, and shell syntax checks passed.
- Full workspace testing on macOS encountered Linux installer failures because GNU `mv -T` is unavailable. A subsequent run excluding Linux-specific tests hit a listener-diagnostic failure that passed in isolation; all remaining package suites passed separately. A completely passing full workspace run remains for CI.

## Checklist

- [x] Architecture docs (`docs/architecture/`) updated — `Solana Native Integration.md`.
- [x] Sealed Approval invariants respected (no signing outside a grant or bounded capability, no PRF/grant persistence, execution from sealed bytes).
- [x] Agent Documentation updated (`crates/bloom-vfs/src/docs/agent-guidance.md` and affected Petal READMEs) — mounted guidance updated; Petal READMEs N/A, no Petal behavior changes.


josh-richardson:
**[P2] Preserve explicit broadcast opt-outs**

`Config::migrate()` forces `allow_broadcast = true` for every EVM and Solana chain on every load. Consequently, an operator can set `allow_broadcast = false`, restart Bloom, and unknowingly have broadcast re-enabled. For Solana, previously read-only custom networks without a genesis pin now prevent startup because migration enables broadcast before validation. This also contradicts the README and security audit, which promise that explicit `false` disables sends.

New defaults should apply only when the field/network is absent; explicit opt-outs should remain authoritative.

PR #302, opened 2026-09-16
Bounded Petal sessions: budgets, address before approval, policy binding, recoverable preparation

0xdewy:
Petal session keys can now carry owner-approved spending budgets. Their reusable approval is prepared only after the session address is visible, and it is re-prepared when the wallet policy changes. The approval's preparation also survives lost responses and restarts, and never outlives the key. This PR includes the setup-order and policy-binding commits from #303, which was folded in here.

## Why

A Petal session that moves value (for example, a Pump.fun trading session) needs these guarantees:

- **Spending budgets.** The Broker refuses any declared debit or fee in an asset missing from the approval's limits (`VALUE_ASSET_NOT_ALLOWED`). Master sealed `value_limits: []`, and its key-request type rejected a Petal's `approval_value_limits` outright.
- **Address before approval.** The Broker binds every approval to the wallet policy snapshot and refuses it once the policy changes (`POLICY_SNAPSHOT_MISMATCH`). Master prepared the approval in the same step as key derivation, so admitting the new session address to the destination policy made that approval unusable. Master also kept reporting the key as Ready.
- **An expiry the Broker accepts.** Master set approval expiry from the current time using nearly the whole key lifetime. The Broker caps it at the key's fixed expiry, so preparing an approval more than about a minute after derivation produced terms the Broker refuses (`PETAL_KEY_SCOPE_MISMATCH`).
- **Recoverable preparation.** A lost prepare response, or a crash between the two state writes, left Machine unable to tell which approval the key relied on. Preparing again with new timestamps reuses an operation id with different terms, which the Broker refuses. Its review manifest includes the preparation time, so even an identical replay later conflicts.

## What changed

- **Budgets.**
  - `PetalKeyGuestRequest` and `PetalKeyRequest` accept optional `approval_value_limits`.
  - Before custody starts, Machine rejects empty assets, zero lifetimes, zero rolling-window durations, and repeated assets.
  - The limits are sealed unchanged into the reusable approval.
  - A retry of the same key slot must carry the same list; a different budget needs a new slot.
- **Setup order.**
  - When custody completes, Machine records the public key and returns Pending; the owner's sessions tree shows the address.
  - The next identical request prepares the approval against the policy then in force.
- **Frozen prepare.**
  - The complete `ApprovalPrepareRequest` is written to the key's Machine-only session record (`<state>.session`) before it is sent.
  - A retry looks that approval up by id:
    - Active → Ready.
    - Waiting on the owner → Pending; the id is recorded if the response was lost.
[...]

PR #303, opened 2026-09-16
Show the Petal session address before preparing its approval, and restage on policy change

0xdewy:
Stacked on #302 (session budgets). Retarget to `master` before #302's branch is deleted.

A Petal session key's reusable approval is now prepared on the request after key derivation completes, not in the same step. It is also restaged when the wallet policy has changed since it was prepared.

## Why

A session that moves value needs its derived address in the wallet's destination policy (for example, to fund it). The Broker binds every Sealed Approval to the policy snapshot it was prepared under. It refuses the approval at sign time once the policy version or digest changes (`POLICY_SNAPSHOT_MISMATCH`).

On master, Machine prepared the reusable approval in the same call that saw the custody result. The owner therefore approved it before anyone could see the new address, and admitting the address to the policy afterwards made that approval unusable. Master also kept reporting the key as Ready from the approval's Active state, so the session silently could not sign.

## What changed

- **Custody completes → Pending.** The derived public key is recorded and the request returns Pending with no approval prepared. The owner's sessions tree already shows `delegated.addresses` for any record with a public key.
- **Next identical request → approval.** The reusable approval is prepared against the wallet policy then in force.
- **Policy snapshot recorded.** The prepared approval's policy version and digest are stored in the Machine-only session record (`<state>.session`, added in #302) beside the request state.
- **Policy changed → replacement.** On a later request, if the wallet policy no longer matches that snapshot, Machine stages a replacement approval (a new owner ceremony) instead of reporting Ready, whatever the old approval's state. Approvals prepared before this record existed have no snapshot and keep master's behaviour.
- **Operation id includes the policy.** The reusable-approval operation id now covers the policy version and digest (`…-operation/v2`), so a replacement never reuses the replaced approval's operation id.
- **Rollback.** The request state file keeps shapes v0.3.0 already reads: a derived-but-unapproved record is `awaiting_user` with a public key. A v0.3.0 binary that reads it prepares the approval immediately, which is its existing behaviour.
- **Docs.** `Sealed Approvals.md` describes the order and the restage.

This narrows when authority is prepared; it grants nothing new.

## Verification

`bloom-daemon` lib tests (99) and `bloom-vfs` lib tests (409) pass; Clippy is clean. The key-request test now covers:

- custody completion returns Pending, records the key, and prepares no approval;
- the next request prepares it;
- after Ready, a policy change returns Pending (not Ready) and prepares a second approval under a different operation id;
- the replacement then reports Ready.

The test fails if policy-change detection is removed, or if the policy is dropped from the operation id.

Not run: a real Broker boundary test, or a live Pump.fun session through the policy update.

The owner can review the order below.
- **Unchanged:** an Expired or Orphaned approval under an unchanged policy is still refused, as on master; this PR does not restage it.
- **Design decision:** after an Active approval, a policy change triggers a new ceremony for the same key rather than requiring a new session slot.

## Checklist

- [x] Architecture docs updated — `Sealed Approvals.md`
- [x] Sealed Approval invariants respected — approvals are never reported usable over a stale policy snapshot; no scope or limit change


0xdewy:
These commits are now part of #302's branch: budgets, setup order, policy binding and their durable state are one bounded-session change, reviewed together on #302. GitHub shows this PR as merged because its commits reached its base branch; nothing landed on master.

Issue #74, opened 2026-07-07
[vfs]: AGENTS.md in the vfs should detect if the vfs is mounted or not and use examples based on that


GavinPacini:
`AGENTS.md` exposed in `/bloom` should be self-aware and dynamic. e.g. if the user has mounted the VFS, it should only show relative directory usage with normal bash tools (`cat requests/pending/...` over `bloom vfs cat requests/pending...`). We should be able to detect this and act accordingly.

If the VFS is mounted, there should be no instruction to the agent to use `bloom vfs` commands.

AC: If bloom is served / mounted, there should be no hints towards `bloom vfs` in any mounted agent-guidance documentation.

GavinPacini:
Actually changing this to just surface standard bash commands rather than `bloom vfs` at all.

Issue #218, opened 2026-09-04
Recover single-payload signing after a definite terminal failure


0xdewy:
## Problem

The single-payload triad signing path can retain `sign_dispatched = true` after Broker reports a definite terminal operation state (`Denied`, `Cancelled`, or `Failed`). Later confirms re-read the same immutable dead operation. The only escape is a different on-chain action such as cancel, even though Broker has established that no signature exists.

The batch path handles this distinction in #214. The single-payload path needs the same safety property without treating an ambiguous outcome as retryable.

## Required invariant

Signing state must distinguish:

- **definitely no signature:** Broker returned a terminal state or protocol error whose contract says no durable effect occurred; the stale approval/signing lineage may be replaced;
- **a signature may exist:** response loss, quarantine, or any ambiguous durable effect; the lineage must remain locked and reconcile rather than sign again.

Use one shared classification for single and batch paths. Do not add another list of loosely related special cases.

## Acceptance

- `Denied`, `Cancelled`, and `Failed` replace the full stale approval/signing lineage and permit a fresh exact approval for the same bytes.
- `Quarantined`, response loss, and unknown durable effects never re-sign.
- Resetting a lineage clears stale ceremony URLs, approval IDs, operation IDs, expected digests, and dispatch markers atomically while retaining the audit/history record.
- Tests cover every terminal state, response loss before/after dispatch, restart, and exact retry.

Related: #213 is the distinct local-discard operation for a transaction that was never broadcast.

PR #304, opened 2026-09-16
Let exact Petal signing use the Petal's own session key within its scope

0xdewy:
A Petal can now sign an Exact (per-payload, owner-approved) request with its own session key. This is how it recovers funds held by a session key: closing token accounts and sweeping SOL back, including after the session is stopped or expired.

## Why

Session keys hold funds, for example a Pump.fun trading session. When a session ends, the Petal needs to sign a recovery transaction from that key. Two things blocked this:
- **Machine:** denied every Exact request that named a key (`exact Petal signing supplied a key reference`).
- **Machine client:** the Exact path accepted only the wallet root or an active account key.

The Broker already supports this case. It checks the key's scope against the approval (wallet, route, suites, active lineage), and lets Exact approvals outlive the session's expiry and stop, while Reusable ones do not. `Sealed Approvals.md` already describes this recovery path; no route could use it.

## What changed

- **Machine (`bloom-daemon`).** An Exact request may name a key only when exactly one recorded Petal key state for the wallet has all of these properties:
  - it holds that public key;
  - it belongs to the executing package's active lineage;
  - it covers the executing route, operation class, and suite.

  `account_owner` already requires the key to be delegated from the mounted account's owner. Naming a wallet key (root or account) under Exact is still refused. Refusals happen before any approval is prepared. The signing request id binds the named key.
- **Class check.** Machine is where the operation class binds: for Exact approvals the Broker passes no classes to its lineage-scope check.
- **Machine client.** `ExactPayloadSignRequest.delegated_key_ref` selects the key. It is valid only with Petal provenance and without `account_key_ref`. It is verified against the Broker's public key metadata, using the same checks the Reusable path already applies (extracted into `verified_delegated_key`).
- **VFS signer.** `BrokerExactPayloadSigner::with_delegated_key` passes the key through and binds it into the persisted operation identity. It uses the existing `account_key_ref` state field, so the state file keeps the shape v0.3.0 reads.
- **Approval window.** An Exact approval for a session key asks for at most the key's maximum lifetime. Signer caps every approval for that key at it, so the fixed five-minute window made keys shorter than five minutes unrecoverable.
- **Retry before the owner approves.** When a stored Exact approval still waits on its ceremony, a retry returns the same pending approval instead of signing (single and batch paths). Signing with it failed with `CLAIM_INVALID`, which the Petal treated as a refusal and dropped the approval. This predates this PR on master.
- **Refusal versus unknown outcome.** Machine reported every Broker signing error to the Petal as a refusal, including a lost response after the Signer had signed. A Petal treats a refusal as proof that no signature exists and may rebuild, so this could produce a second transaction while a signature for the first existed. A signing failure is now a refusal only when Broker's error contract says so (never retryable, and no effect or a released reservation) and no earlier attempt was reported finalized; otherwise the host returns a fixed "signing outcome unknown" backend error. Exact (single and batch) and Reusable paths. This predates this PR on master.
- **Scope of support.** Batches still do not accept a named key.
- **Docs.** `Sealed Approvals.md` documents the rule.

This widens what a Petal can sign, so it is its own PR. Every such payload still needs its own owner ceremony.

## Verification

- `bloom-daemon` (101), `bloom-machine-client` (53), `bloom-vfs` (409), `bloom-solana-tx` (20) and `bloom-tx` (153) lib tests pass. Clippy on the workspace is clean.
- A new daemon test refuses, with no approval prepared:
  - an unrecorded key;
  - the wallet root key;
  - a class outside scope;
  - a key recorded under another lineage;
  - a route outside scope;
  - a suite outside scope;
  - another wallet's session key.
- In the same test, the matching key (on a stopped session) gets its own approval and signs; the prepared terms and the signing request both carry the session key.
- A second test runs the same boundary on a BIP-39 wallet with Ed25519 keys. A session key delegated from Solana account 1 is refused on the account 0 mount and signs on the account 1 mount (64-byte signature, Ed25519 suite).
- The test fails if the scope check always passes, or if the lineage condition is dropped.
- A machine-client test covers the request validation.
- `a_retry_before_the_owner_approves_stays_pending`, `signing_failures_are_refusals_only_under_brokers_contract`, and the session-key tests' approval-window assertion each fail when their fix is removed.
- **Real services** (Signer bloom-signer#57, Broker `dd2add2b`, merged locally with #302 and #276; evidence in the workspace `shared/pumpfun-verification-2026-09-17/`), with real Ed25519 session keys: another account's owner key and another Petal's session key are refused before any approval; a stopped key and an expired 60 s key sign an Exact payload after its own ceremony (signatures verified); two payloads need two approvals. On a local validator with test-only loopback builds, Pump.fun close and sweep land the exact approved bytes, including a sweep retried while pending, after a stop, after an RPC outage, and after a blockhash expiry. The fixes above came from this run. A lost signing response (a test proxy delivers `signing.sign` and withholds Broker's response) now leaves the Pump.fun operation possibly signed instead of rebuildable; recovering the withheld signature also needs bloom-directory/bloom-broker#74, with which the same message landed once with that exact signature.
- Not run: mainnet close/sweep or final landed/pinned Linux and macOS release conformance. The optimized development-candidate check is recorded below.

## Dependencies and landing

- **Depends on:** nothing in code; lands after #302 with master merged forward.
- **Pairs with:** bloom-directory/bloom-broker#74. This PR makes a lost signing response leave the operation possibly signed; #74 lets the retry recover the signature. Claim ambiguous-signature recovery only for a release pinning both.
- **Pins:** none change in this PR (Broker `dd2add2b`, Signer `ccc9adb3`, runtime `5db670e1`). The Broker and Signer pins advance in a separate Bloom PR after bloom-directory/bloom-broker#74 and bloom-directory/bloom-signer#57 land.
- **Release:** Pump.fun (bloom-directory/bloom-petal-pumpfun#5) needs a Bloom release containing #276, #302 and #304. Recovery also requires the fixes in bloom-directory/bloom-broker#76 (for #75) and bloom-directory/bloom-service-runtime#21 (for #20), consistently pinned into the release.
- Full sequence: workspace `shared/pumpfun-verification-2026-09-17/LANDING.md`.

## Integration status

[PM #66](https://github.com/bloom-directory/pm/issues/66) owns reviews, landing, consistent pins and final release checks. [PM #35](https://github.com/bloom-directory/pm/issues/35) owns mainnet acceptance, which remains unverified. Optimized, test-enabled candidate binaries passed Part A (32/0); local combinations and Bloom's runtime patch do not establish final release conformance.

## Checklist

- [x] Architecture docs updated — `Sealed Approvals.md`
- [x] Sealed Approval invariants respected — one owner approval per payload; key scope checked by Machine, Broker, and Signer


0xdewy:
Closing this mixed PR unmerged at the workspace owner's request as part of retiring the funded-session Pump.fun scope in https://github.com/bloom-directory/pm/issues/66. Its session-key Exact recovery extension is no longer required for direct account trading. The original branch and evidence are retained.

The independent fixes are **preserved, not discarded**: pending Exact approvals, definitive-refusal versus unknown-signing classification, and waiting on unconfirmed ceremonies have been manually extracted/adapted onto master at local branch `exact-approval-pending`, commit `904bff5e`, in `worktrees/pfdirect/bloom/exact-approval-pending`. No rebase of this PR was needed. That extraction is unpublished and unmerged; publication/review/landing remains explicitly tracked in https://github.com/bloom-directory/pm/issues/35.

The local diagnostic combination uses the extraction and Broker #74 for real-service local-validator recovery evidence (run d7 in `shared/pumpfun-direct-live-2026-09-22/`). This does not claim release completion or resolve all of issue #218. No branches are deleted by this closeout.

PR #306, opened 2026-09-17
Integrate remote ceremonies and privileged hosted setup

GavinPacini:
Adds Machine and installer support for hosted remote ceremonies. Wallet creation/import accept explicit localhost selection; `bloom ceremony surfaces` exposes Signer-approved desired state and Broker-confirmed readiness; `bloom wallet add-passkey NAME --to local|remote` launches paired enrollment. CLI and VFS preserve opaque Broker URLs and public receipt projections.

The packaged `bloom-ceremonies` wrapper invokes privileged Signer administration through fixed platform paths. Installers provision private administrator state, retain process/UID boundaries, and report incomplete hosted setup explicitly. Broker gets the networking required for relay/ACME/DNS while Signer remains network-isolated. Compatibility moves to Machine–Broker 1.7, Broker–Signer 1.6, and state/downgrade floor 2.

Reuses existing Broker RPC, Machine projection, installer and WebAuthn/HPKE browser flows. Custom code handles Bloom-specific selection/status and fixed-path privilege handoff. User-provided architecture and planning edits are preserved.

Validation and operational gaps are recorded in `docs/reviews/2026-09-17-remote-ceremonies-implementation.md`. Final locked workspace: 1,669 passed / one ignored across 64 suites; strict all-target Clippy and exact Git-pin compatibility gate passed. Integration and staged packaging evidence is recorded in the review document. Final repository CI, Solana validator CI, and Linux/macOS release dry-run passed at the frozen refs; the repository owner resolved the initial dependency access failure by making relay public. No deployment or release publication is included.

Land after runtime PR19, relay PR1, Signer PR58 and Broker PR73, repinning each dependent repository to merged revisions in that order.

## Checklist

- [x] Tests added or updated for behavior changes — CLI/VFS selection/status, pairing projection, installer permissions/lifecycle and deterministic listener diagnostic.
- [x] Architecture docs updated if contracts or behavior changed — supplied remote architecture preserved; normative Triad sections 12–13 updated.
- [x] Sealed Approval invariants respected — no Machine signing authority, no PRF/grant persistence, activation remains separate from execution.
- [x] Agent Documentation updated — embedded VFS guidance explains opaque URLs, surfaces and privilege boundaries; affected Petal README changes are N/A.

Two-login macOS conformance is deferred at the user’s direction. Frozen-ref runs: [release dry-run](https://github.com/bloom-directory/bloom/actions/runs/35231172657), [single-login macOS conformance](https://github.com/bloom-directory/bloom/actions/runs/35231176192), and [CI](https://github.com/bloom-directory/bloom/actions/runs/35231016045). No production signing or release publication is enabled.

The macOS candidate job runs the entire Broker workspace before installation. Installed acceptance repeats all tests except the exclusive canonical-listener test, and only when the workflow explicitly records that prior coverage. This avoids a conflict with the running installed Broker without dropping the listener gate.

Final sibling pins: Broker `f51fc152a35425f09efd2ecbd60eb92b5b243054`, Signer `7dddac2985bf00810589f02b4081062ef576a63b`, runtime `a046fd6075853e8591f295983b572c6e1d65d12d`. Broker’s final privacy regression proves that prior wallet activity does not alter public recovery admission or authenticated projection shape.

Final hosted evidence: CI and both release dry-run candidates passed. Single-login macOS W0 failed initial installed Machine activation on two disposable runner attempts. The first attempt captured repeated `mount_nfs` children reaching Bloom’s 10-second timeout after the embedded NFS server began listening; the retry again failed Machine socket health during installation. The mount implementation is unchanged by this PR. The evidence does not distinguish NFS negotiation from runner permissions/environment, and W0 is not claimed as passed. A real-mount diagnostic on a disposable macOS host is required before claiming installed conformance. Two-login conformance remains explicitly deferred by the repository owner.


crates/bloom-vfs/src/handlers/wallets.rs
```diff
@@ -2738,6 +3272,35 @@ impl WalletsHandler {
         if segs.len() == 1 && segs[0] == "new" {
             return Ok(Entry::writable_file("new"));
         }
+        if segs.len() == 1 && segs[0] == "recover" {
```
josh-richardson, line 3275: P2: `recover` and `recoveries` were accepted wallet names before this change. These new top-level path handlers take precedence over wallet lookup, so an existing wallet with either name is now shadowed in the mounted VFS: `recover` becomes a file, while paths under `recoveries` are routed to recovery records. The wallet can no longer be accessed through its prior `/wallets/<name>/...` paths. Please provide a migration or a collision-free route for existing wallets.

GavinPacini: On this one, I think we should now switch to wallet IDs and wallet names being separated i.e start using UUIDs for new wallets, with a mutable "display name" stored alongside it. I think in general it would just be good practice to do so, but it also means we can support renaming wallets and finally, in this specific case, we could use `recover` and `recoveries` here and rename (the display name of) any existing wallets with those names to avoid them being shadowed.

GavinPacini: Agreed: these names were previously valid, and the new controls shadow their existing VFS paths. We are retaining `/wallets/recover` and `/wallets/recoveries` and have explicitly deferred the identity/rename work to https://github.com/bloom-directory/bloom/issues/323.

The planned solution separates a full immutable wallet UUID from a mutable CLI/VFS name, plus a stable short identifier shared with passkey labels. Existing wallets retain their cryptographic custody IDs behind a durable Signer-owned UUID binding, so renaming does not require rewriting ciphertext or re-enrolling credentials. The issue covers full-UUID access/rename for shadowed wallets, reserved-name enforcement across all creation/import paths, existing-wallet recovery, and migration/backup/restore tests.

This is tracking the remaining fix, not claiming the collision is resolved in this PR. CLI projection/account reads are not a complete workaround for VFS-backed operations; no automatic installer rename is planned.

PR #307, opened 2026-09-18
Let a developer harness add a Petal HTTP root certificate

0xdewy:
## The problem

The daemon's Petal HTTP client (`DaemonPetalHost`) verifies TLS against rustls' compiled-in webpki roots. Nothing reaches it from the outside: `SSL_CERT_FILE` and `/etc/ssl/certs/ca-certificates.crt` are both ignored.

Petal net policy (`bloom-petals/src/policy.rs`) requires HTTPS on port 443 against the exact host declared in `petal.toml`, and an `[petals.runtime.<name>].endpoints` override must itself be an HTTPS origin on port 443. So there is no plaintext path either.

Together those mean a Petal's real HTTP routes cannot be driven by any external harness. They can only run against the live public venue.

Measured, not inferred. A harness serving `gamma-api.polymarket.com` from a local HTTPS server with its own CA, inside a network namespace, with the CA installed both as `SSL_CERT_FILE` and as the system bundle:

- the daemon resolves and **connects**: `hyper_util ... connected to 127.0.0.1:443`
- the server then logs `SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca`
- the Petal route fails with `backend: http_fetch send: error sending request for url (https://gamma-api.polymarket.com/markets/slug/...)` and the server never sees an HTTP request

## The change

`BLOOM_PETAL_HTTP_EXTRA_ROOT_CA` names a PEM bundle whose certificates are added as trust anchors on that client.

It is compiled out unless `unsigned-audit-test-seam` is enabled — the existing nonproduction constructor seam, which release packaging must not enable. Default builds are unchanged; the function is a no-op that returns the builder.

It widens **trust anchors only**. The request path, the `[[net.allow]]` policy check, the audit intent/result journal, the redirect bound and the per-redirect policy re-check are all untouched, so a harness using this still exercises the production HTTP path instead of bypassing it.

An unreadable file, a non-PEM file, or a PEM containing no certificates fails loudly rather than being silently ignored — silently ignoring it would reproduce the original opaque TLS symptom.

## Tests

`petal_http_trusts_an_extra_root_certificate_bundle` covers the accepted bundle (two concatenated certificates) and all three rejections. Run with `cargo test -p bloom-daemon --features unsigned-audit-test-seam`.

## Why it is wanted

The interface benchmark in `vfs-research` compares a real Bloom FUSE mount against `bloom vfs` on the same tasks. Extending it to Petal-backed workflows needs the actual Polymarket Petal executing its real routes against controlled venue responses, offline. That is impossible today.

## Checklist

- [x] Tests added or updated for behavior changes — `petal_http_trusts_an_extra_root_certificate_bundle`
- [x] Architecture docs (`docs/architecture/`) updated if contracts or behavior changed — N/A: no contract or behavior change. The seam is compiled out of default builds, and with it enabled the request path, net policy, audit journal and redirect handling are untouched; only TLS trust anchors widen.
- [x] Sealed Approval invariants respected — N/A: no signing, grant, PRF or sealed-bytes path is touched.
- [x] Agent Documentation updated — N/A: nothing on the agent-facing VFS surface or any Petal README changes; the seam is a developer-harness build option with no runtime surface.


0xdewy:
Review fixes in e50e9b17:

- The checklist-complete failure was stale: it ran against the opening body revision before the checklist section existed. Re-pushed, now green.
- Real gap found in review: no CI job enables unsigned-audit-test-seam, so the new test never ran anywhere and the gated code was uncompiled. Added an audit seam tests job (daemon lib tests plus clippy under the seam).
- PETAL_HTTP_EXTRA_ROOT_CA_ENV is now private; nothing outside the module uses it.

Verified locally: seam test passes, clippy -D warnings clean, fmt clean. Full CI green on this head. Still needs a reviewer (all checks pass except review-required).

0xdewy:
Closing: not necessary. The benchmark runs without it, and the Petal-workflow extension it enables isn't committed work yet — it can also go narrower or use the live venue. Branch kept; cheap to revive if that changes.

0xdewy:
Closing. This is not needed — Bloom already ships a supported way to do it, and the assessment that preceded this PR ruled it out by asking the wrong question.

## The mechanism that already exists

`[petals.runtime.<mount>].endpoints` is wired end to end:

- `crates/bloom-petals/src/router.rs:86` turns each entry into both a runtime setting `endpoint.<binding>` and a policy binding.
- `crates/bloom-petals/src/policy.rs:39` (`with_endpoint_bindings`) rewrites the allowed host for the matching `[[net.allow]]` rule.
- `bloom-petal-polymarket/route/src/runtime_config.rs:10` shows the consumer side: the Petal builds every venue URL from `runtime_setting("endpoint.<binding>")`, falling back to the host in its manifest.

`endpoint_origin_host` accepts any HTTPS origin on port 443 with no path, query or fragment. There is no rule tying an override to the declared host's domain — `package.rs:6534` already tests overriding `clob` to `https://clob.internal.example`.

## What that means for an offline harness

Register a domain you control. Obtain an ordinary publicly trusted certificate for the subdomains standing in for the venue, via a DNS-01 challenge, which needs no public reachability. Resolve those names to 127.0.0.1. Set the endpoint overrides in the Machine's config. The local fixture then presents a certificate Bloom's compiled-in roots already trust.

No Bloom change, no custom trust anchor, no forked Petal, and the package hash under test is unchanged. The `[[net.allow]]` check, the audit intent and result, the redirect bound and the per-redirect re-check all still run, which was this PR's stated advantage over the alternatives.

It is also more truthful at the operator surface. The consent summary reports `effective_origin=https://clob.<your-domain>`, so it is visible that the Petal is not talking to the real venue. Under this PR the surface would keep naming the real host while the traffic went to a fixture.

## Why the earlier assessment missed it

`research/bloom-product-surface/order-recovery/contract/offline-http-options.md` asked whether the endpoint override could name a *local server*. It cannot — it rejects `localhost:8080`, and that was recorded as a dead end. The question it did not ask was whether the declared host could be made one you control. It can, using the feature exactly as designed.

## What the alternative costs

A domain, DNS-01 automation and a 90-day renewal; an expired certificate fails the harness with the same opaque TLS error this PR set out to avoid, though with an obvious cause. The private key has to reach wherever the harness runs. A machine that can never obtain a certificate cannot run it.

That is real operational work, and it is still less than carrying a trust-widening seam in a custody-adjacent daemon indefinitely.

## State

Branch `petal-http-test-root` is kept at `8b5a9b56`, CI green, with the review defects fixed (gate shape, fallible constructor, and a CI job that actually compiles the gated code). Nothing here is lost if the no-change route turns out to be unworkable — but it should be tried first.

